PR #2, opened 2020-05-15
fix tarball name

buixor:
- include the version name in the tarball


AlteredCoder:
I think the doc needs to be fixe also ?

PR #16, opened 2020-05-20
Linter fixes

buixor:
Starting to fix main linting issues according to [golangci-lint](https://github.com/golangci/golangci-lint):
 - Missing err checks
 - Dead code
 - Overall code simplification

Adding a golangci-lint github action. With the default setup the `--new-from-rev=HEAD~` doesn't work (because we're in detached HEAD mode after the checkout action ?), so I fetch and force the diff with master (for now).

After asking around about the issue, it seems that our solution is the one so far : https://github.com/golangci/golangci-lint-action/issues/16

Let's improve this when the option is available again !


cmd/crowdsec-cli/api.go
```diff
@@ -189,7 +185,6 @@ cscli api credentials   # Display your API credentials
 			}
 			fmt.Printf("machine_id: %s\n", outputCTX.API.Creds.User)
 			fmt.Printf("password: %s\n", outputCTX.API.Creds.Password)
-			return
```
FaricaUnknown, line 192: For my own curiosity, I wonder why why this pattern is not linter compliant. I thought put return was a best practice.

pkg/cwapi/auth.go
```diff
@@ -113,7 +113,7 @@ func (ctx *ApiCtx) Init(cfg string, profile string) error {
 		return err
 	}
 	//start the background go-routine
-	go ctx.pushLoop()
+	go ctx.pushLoop() //nolint:errcheck
```
FaricaUnknown, line 116: Maybe we should issue guidelines or some kind of rules to be allowed to escape linter ?

jirfag:
hi, golangci/golangci-lint-action#16 is fixed now

pkg/parser/node.go
```diff
@@ -412,27 +410,24 @@ func (n *Node) compile(pctx *UnixParserCtx) error {
 	}
 	/* compile leafs if present */
 	if len(n.SuccessNodes) > 0 {
-		for idx, _ := range n.SuccessNodes {
+		for idx := range n.SuccessNodes {
```
FaricaUnknown, line 413: I am very surprised at this one. Do you have some hint on this ?

pkg/parser/node.go
```diff
@@ -412,27 +410,24 @@ func (n *Node) compile(pctx *UnixParserCtx) error {
...
-			} else {
-				//n.logger.Debugf("Leaf compilation suceeded: %v\n", n.SuccessNodes[idx])
 			}
-			//set child node to parent stage
 		}
 		valid = true
 	}
 	/* load statics if present */
-	for idx, _ := range n.Statics {
+	for idx := range n.Statics {
```
FaricaUnknown, line 430: same as above

pkg/time/rate/rate_test.go
```diff
@@ -400,7 +400,7 @@ type wait struct {
 func runWait(t *testing.T, lim *Limiter, w wait) {
 	start := time.Now()
 	err := lim.WaitN(w.ctx, w.n)
-	delay := time.Now().Sub(start)
+	delay := time.Since(start)
```
FaricaUnknown, line 403: I think this is not on our code. I guess the `Since()` is kind of new for code that should be usable on older version of golang.

pkg/cwapi/auth.go
FaricaUnknown: Maybe add an extra comment on why we bypass linter, would be enough ? You did that below, and it's ok with me.

cmd/crowdsec-cli/api.go
buixor: code simplification : return doesn't add anything

pkg/parser/node.go
buixor: Because we only care about the index in the iteration, not the value itself

pkg/parser/node.go
buixor: same answer as above : https://play.golang.org/p/ijKEvRUKKNa

PR #25, opened 2020-05-22
add json support via expr helpers

buixor:
- Add a method to allow extraction of one or more field from json `JsonExtract(evt.Parsed.Field, "my.sub_field.json[0]")`
 - Add unit test for json extraction


AlteredCoder:
I would have called the function only `JSON()` or `Json()` instead of `JsonExtract()`

buixor:
@AlteredCoder : I'm afraid calling it "only" `Json` or `JSON` might be confusing, maybe we can find a middle ground :)

pkg/exprhelpers/jsonextract.go
```diff
@@ -0,0 +1,37 @@
...
+	return strvalue
+}
+
+func JsonExtract(jsblob string, target string) string {
+	if !strings.HasPrefix(target, "[") {
+		target = strings.Replace(target, "[", ".[", -1)
+	}
+	fullpath := strings.Split(target, ".")
+
+	log.Printf("path -> %+v", fullpath)
```
FaricaUnknown, line 35: This seems like a kind of debug. Maybe replace this with `log.Debugf("path -> %+v", fullpath)`

pkg/parser/parsing_test.go
```diff
@@ -98,162 +115,223 @@ func testOneParser(t *testing.T, dir string) error {
...
-			if err != nil {
-				log.Errorf("Failed to process %s : %v", spew.Sdump(in), err)
+		testSet = append(testSet, tf)
+	}
+	return testSet
+}
+
+func matchEvent(expected types.Event, out types.Event, debug bool) ([]string, bool) {
+	var retInfo []string
+	var valid bool
```
FaricaUnknown, line 183: `var valid = false` but not mandatory because it's the zero value

pkg/parser/parsing_test.go
```diff
@@ -98,162 +115,223 @@ func testOneParser(t *testing.T, dir string) error {
...
+	//allow to check as well for stage and processed flags
+	if expected.Stage != "" {
+		if expected.Stage != out.Stage {
+			if debug {
+				retInfo = append(retInfo, fmt.Sprintf("mismatch stage %s != %s", expected.Stage, out.Stage))
+			}
+			valid = false
+			goto checkFinished
+		} else {
+			valid = true
```
FaricaUnknown, line 197: The tool reporting this linter error, has the pretention to not display false positive. But from what I see it is a false positive.

pkg/parser/parsing_test.go
```diff
@@ -98,162 +115,223 @@ func testOneParser(t *testing.T, dir string) error {
...
+	for n := 0; n < count; n++ {
+		if testFile(tests, *pctx, pnodes) != true {
+			return fmt.Errorf("test failed !")
+		}
 	}
 	return nil
 }
 
-func testFile(t *testing.T, file string, pctx UnixParserCtx, nodes []Node) bool {
+func prepTests() (*UnixParserCtx, error) {
```
FaricaUnknown, line 132: Can you some comments here, to be able to read more easily.

buixor:
> I didn't see the json file used for the test. Can you point me this file ?

They are in `pkg/parser/tests/base-json-extract` and represent a scenario where we extract first from json and then apply grok patterns on it.

PR #26, opened 2020-05-24
Merge cscli crowdsec config

AlteredCoder:
`cscli` now use `crowdsec` configuration file instead of its own.

### DIFF
--- a/cmd/crowdsec-cli/api.go
+++ b/cmd/crowdsec-cli/api.go
@@ -158,7 +158,7 @@ cscli api credentials   # Display your API credentials
 				return err
 			}
 
-			err = outputCTX.LoadAPIConfig(path.Join(config.InstallFolder, apiConfigFile))
+			err = outputCTX.LoadAPIConfig(path.Join(config.installFolder, apiConfigFile))
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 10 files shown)


cmd/crowdsec-cli/api.go
```diff
@@ -158,7 +158,7 @@ cscli api credentials   # Display your API credentials
 				return err
 			}
 
-			err = outputCTX.LoadAPIConfig(path.Join(config.InstallFolder, apiConfigFile))
+			err = outputCTX.LoadAPIConfig(path.Join(config.installFolder, apiConfigFile))
```
buixor, line 161: you don't want InstallFolder to be exported ?

cmd/crowdsec-cli/main.go
```diff
@@ -37,31 +36,30 @@ func initConfig() {
 		log.SetLevel(log.ErrorLevel)
 	}
 
+	csConfig := csconfig.NewCrowdSecConfig()
+	if err := csConfig.GetCliConfig(&config.ConfigFilePath); err != nil {
+		log.Fatalf(err.Error())
+	}
+	config.configFolder = filepath.Join(filepath.Clean(csConfig.CsCliFolder))
```
buixor, line 43: I think you can get rid of join or forgot to add second element ?

PR #37, opened 2020-05-26
Support data download

AlteredCoder:
This MR add the support of downloading external dataset in parser/scenario, mostly to use in `filter` 
 directive.
Data are download while installing parser/scenario with `cscli`

Two new function are available for the expr filters that laverage this feature :

`File(<filename>)` return an array of string (the file content is splitted by new line)

example:

```
filter: evt.Parsed.http_user_agent in File("bad_user_agents.txt")
```

`RegexpInFile(stringToCompare, <filename_of_regex>` check if any regex in the file match the string

example:
```
filter: RegexpInFile(evt.Parsed.http_user_agent, "bad_user_agents_re.txt")
```

see https://github.com/crowdsecurity/hub/blob/master/scenarios/crowdsecurity/http-bad-user-agent.yaml for a real example

this MR also fixes :
 - cscli relative/absolute symlink bug (appears in testing environements)
 - #39 : default acquis.yaml is misleading
 - Bug in test_env.sh script
 - Add tests for exprhelpers lib


pkg/exprhelpers/exprlib.go
```diff
@@ -26,9 +33,71 @@ func EndsWith(s string, suff string) bool {
...
+	dataFile = make(map[string][]string)
+	dataFileRegex = make(map[string][]*regexp.Regexp)
+	return nil
+}
+
+func FileInit(fileFolder string, filename string, fileType string) error {
+	filepath := path.Join(fileFolder, filename)
+	file, err := os.Open(filepath)
+	if err != nil {
+		log.Fatal(err)
```
buixor, line 54: I think we might want to return an error here so we can give context to the user :
'crowdsec didn't start because datafile X in YYY.yaml doesn't exist" ?

pkg/exprhelpers/exprlib.go
```diff
@@ -26,9 +33,71 @@ func EndsWith(s string, suff string) bool {
...
+	if _, ok := dataFile[filename]; !ok {
+		dataFile[filename] = []string{}
+	}
+	if fileType == "" {
+		fileType = "string"
+	}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		switch fileType {
+		case "regex":
```
buixor, line 67: if you want to do this (but it's for the user's experience so it's ok), I would suggest :

```golang
switch ... {
  case "regex", "regexp":
     ..
  case ...:

}
```

pkg/exprhelpers/exprlib.go
```diff
@@ -26,9 +33,71 @@ func EndsWith(s string, suff string) bool {
...
+			dataFileRegex[filename] = append(dataFileRegex[filename], regexp.MustCompile(scanner.Text()))
+		case "string":
+			dataFile[filename] = append(dataFile[filename], scanner.Text())
+		default:
+			log.Errorf("unknown data type '%s' for : '%s'", fileType, filename)
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		log.Fatal(err)
```
buixor, line 79: here as well, returning error to give user's context

pkg/leakybucket/manager.go
```diff
@@ -275,6 +280,15 @@ func LoadBucket(g *BucketFactory) error {
 		g.processors = append(g.processors, blackhole)
 	}
 
+	if len(g.Data) > 0 {
+		for _, data := range g.Data {
+			err = exprhelpers.FileInit(dataFolder, data.DestPath, data.Type)
```
buixor, line 285: and this is where we should print the error with context :)

pkg/types/dataset.go
```diff
@@ -0,0 +1,74 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("download response 'HTTP %d' : %s", resp.StatusCode, string(body))
+	}
+
+	file, err := os.OpenFile(destPath, os.O_RDWR|os.O_CREATE, 0644)
```
buixor, line 45: don't we want O_TRUNC ?

pkg/exprhelpers/exprlib.go
```diff
@@ -25,10 +32,69 @@ func EndsWith(s string, suff string) bool {
...
+	}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		switch fileType {
+		case "regex", "regexp":
+			dataFileRegex[filename] = append(dataFileRegex[filename], regexp.MustCompile(scanner.Text()))
+		case "string":
+			dataFile[filename] = append(dataFile[filename], scanner.Text())
+		default:
+			log.Errorf("unknown data type '%s' for : '%s'", fileType, filename)
```
buixor, line 71: I think we should return an error here

PR #42, opened 2020-05-27
Logging cleanup and fixes

buixor:
- Ensure node-specific or bucket-specific loggers get a base config similar to main logger
 - Cleanup logging in various spots to provide context or lower verbostiry
 - `user.yaml` doesn't try to start prometheus or use API

### DIFF
--- a/pkg/leakybucket/manager.go
+++ b/pkg/leakybucket/manager.go
@@ -191,7 +191,7 @@ func LoadBucket(g *BucketFactory) error {
 	var err error
 	if g.Debug {
 		var clog = logrus.New()
-		if types.ConfigureLogger(clog) != err {
+		if err := types.ConfigureLogger(clog); err != nil {
 			log.Fatalf("While creating bucket-specific logger : %s", err)
 		}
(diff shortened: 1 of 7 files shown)


pkg/leakybucket/manager.go
```diff
@@ -191,7 +191,9 @@ func LoadBucket(g *BucketFactory) error {
 	var err error
 	if g.Debug {
 		var clog = logrus.New()
-		clog.SetFormatter(&log.TextFormatter{FullTimestamp: true})
+		if types.ConfigureLogger(clog) != err {
```
erenJag, line 194: Which error do you check exactly ? 
Isn't the inverse you want to check ?

pkg/parser/node.go
```diff
@@ -336,6 +341,9 @@ func (n *Node) compile(pctx *UnixParserCtx) error {
 	that will be used only for processing this node ;) */
 	if n.Debug {
 		var clog = logrus.New()
+		if types.ConfigureLogger(clog) != err {
```
erenJag, line 344: same here

PR #67, opened 2020-06-08
preparation for hub ci

buixor:
- Move the LoadAcquisitionConfig to the pkg `acquisition`
 - Add a flag to allow `pkg/parser` to dump `types.Event`'s state after each parser
 - Add CI that tests current code against the hub tests

note: this is in preparation of the hub CI


pkg/types/utils.go
```diff
@@ -93,3 +95,16 @@ func ConfigureLogger(clog *log.Logger) error {
...
 	return nil
 }
+
+func Clone(a, b interface{}) {
+
+	buff := new(bytes.Buffer)
+	enc := gob.NewEncoder(buff)
+	dec := gob.NewDecoder(buff)
+	if err := enc.Encode(a); err != nil {
+		log.Fatalf("failed cloning %T", a)
```
AlteredCoder, line 105: i think we better prefer to return an error instead of fatal here

pkg/types/utils.go
```diff
@@ -93,3 +95,16 @@ func ConfigureLogger(clog *log.Logger) error {
...
+	}
+	if err := dec.Decode(b); err != nil {
+		log.Fatalf("failed cloning %T", b)
```
AlteredCoder, line 108: same here

PR #91, opened 2020-06-24
Sqlite : Support automatic db flushing

buixor:
Crowdsec :
 - Support database (sqlite) retention policy via the main yaml file.

Allow to specify "max_records" and/or "max_records_age".
These parameters will trigger (every minute) a check that will cleanup expired records from the SQLite database based on the policy.

 - `AutoCommit` routine in backend is not automatically started when Output is instanciated (so that cscli doesn't start one for ex)

Cscli :

 - Reduce verbosity of cwhub in debug mode

 - Reduce verbosity of output startup (and verbosity of ie `cscli ban add`)

 - Allow 'ban add' to accept reason on more than one arg `cscli ban add ip 1.2.3.4 2h this is a test`

### DIFF
--- a/pkg/sqlite/commit.go
+++ b/pkg/sqlite/commit.go
@@ -17,8 +17,6 @@ func (c *Context) DeleteExpired() error {
 		if retx.RowsAffected > 0 {
 			log.Infof("Flushed %d expired entries from Ban Application", retx.RowsAffected)
 		}
-	} else {
-		log.Infof("flush is disabled")
 	}
 	return nil
 }
@@ -136,6 +134,9 @@ func (c *Context) AutoCommit() {
 	ticker := time.NewTicker(200 * time.Millisecond)
 	cleanUpTicker := time.NewTicker(1 * time.Minute)
 	expireTicker := time.NewTicker(1 * time.Second)
+	if !c.flush {
+		log.Infof("flush is disabled")
+	}
 	for {
 		select {
 		case <-c.PusherTomb.Dying():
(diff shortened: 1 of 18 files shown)


AlteredCoder:
when i add a ban with cscli the output is too verbose:
```
sudo cscli ban add ip 1.2.3.18 5m test
WARN[0000] Starting SQLite backend, path:/var/lib/crowdsec/data/crowdsec.db
INFO[0000] backend plugin 'sqlite' loaded
WARN[0000] starting autocommit
INFO[0000] flush is disabled
INFO[0000] Wrote ban to database.

```

pkg/sqlite/commit.go
```diff
@@ -21,18 +33,110 @@ func (c *Context) Flush() error {
...
+	//let's fetch all expired records that are more than XX days olds
+	sos := []types.BanApplication{}
+
+	if c.maxDurationRetention == 0 {
+		return nil
+	}
+
+	//look for soft-deleted events that are OLDER than maxDurationRetention
+	ret := c.Db.Unscoped().Table("ban_applications").Where("deleted_at is not NULL").
+		Where(fmt.Sprintf("deleted_at > data('now','-%d minutes')", int(c.maxDurationRetention.Minutes()))).
```
AlteredCoder, line 49: it looks like the `data()` function is not supported by sqlite ?

buixor:
> when i add a ban with cscli the output is too verbose:
> 
> ```
> sudo cscli ban add ip 1.2.3.18 5m test
> WARN[0000] Starting SQLite backend, path:/var/lib/crowdsec/data/crowdsec.db
> INFO[0000] backend plugin 'sqlite' loaded
> WARN[0000] starting autocommit
> INFO[0000] flush is disabled
> INFO[0000] Wrote ban to database.
> ```

this is solved now

PR #99, opened 2020-06-29
Update sqlite plugin to database plugin that supported multiples DB backends

erenJag:
As we're using gorm, we update the sqlite plugin to database plugin that supported multiple DB backends that are supported by gorm.


pkg/database/database.go
```diff
@@ -0,0 +1,108 @@
...
+			return nil, fmt.Errorf("failed to open %s database : %s", cfg["db_name"], err)
+		}
+	}
+
+	if val, ok := cfg["debug"]; ok && val == "true" {
+		log.Infof("Enabling debug for %s", cfg["type"])
+		c.Db.LogMode(true)
+	}
+
+	c.flush, _ = strconv.ParseBool(cfg["flush"])
```
AlteredCoder, line 86: can you check the error please

PR #102, opened 2020-06-29
[wip] Update sqlite plugin to database plugin that supported multiples DB backends

erenJag:
As we're using gorm, we update the sqlite plugin to database plugin that supported multiple DB backends that are supported by gorm.


buixor:
mew cleaner MR

PR #115, opened 2020-07-02
add ability to filter 'ban list' output (`--ip` `--range` `--as` `--country` `--reason`)

buixor:
Filter `ban list` output (**note** : it's done at `cscli` level, not at SQL query level) :
 - simple filter : as/ip/country/reason
 - smart `--range` : `--range 0.0.0.0/0` will return all bans

### DIFF
--- a/cmd/crowdsec-cli/ban.go
+++ b/cmd/crowdsec-cli/ban.go
@@ -88,10 +88,7 @@ func simpleBanToSignal(targetIP string, reason string, expirationStr string, act
 }
 
 func filterBans(bans []map[string]string) ([]map[string]string, error) {
-	// fmt.Printf("source,ip,reason,bans,action,country,as,events_count,expiration\n")
-	// 	for _, rm := range ret {
-	// 		fmt.Printf("%s,%s,%s,%s,%s,%s,%s,%s,%s\n", rm["source"], rm["iptext"], rm["reason"], rm["bancount"], rm["action"], rm["cn"], rm["as"], rm["events_count"], rm["until"])
-	// 	}
+
 	var retBans []map[string]string
 
 	for _, ban := range bans {
@@ -100,7 +97,6 @@ func filterBans(bans []map[string]string) ([]map[string]string, error) {
 		var keep bool = true
 		var err error
 
-		//try to parse ip or range from ban
 		if ban["iptext"] != "" {
 			if strings.Contains(ban["iptext"], "/") {
 				log.Debugf("%s is a range", ban["iptext"])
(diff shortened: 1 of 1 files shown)


cmd/crowdsec-cli/ban.go
```diff
@@ -84,7 +87,109 @@ func simpleBanToSignal(targetIP string, reason string, expirationStr string, act
...
+			retBans = append(retBans, ban)
+		} else {
+			log.Debugf("[discard] discard %v", ban)
+		}
+	}
+	return retBans, nil
+}
+
 func BanList() error {
+	//var ipFilter, rangeFilter, reasonFilter, countryFilter, asFilter string
```
AlteredCoder: can you remove this line please

cmd/crowdsec-cli/ban.go
```diff
@@ -84,7 +87,109 @@ func simpleBanToSignal(targetIP string, reason string, expirationStr string, act
...
+			}
+			if filterBinRange.Contains(banIP) {
+				log.Debugf("[keep] range filter %s contains %s", rangeFilter, banIP.String())
+				keep = true
+			} else {
+				log.Debugf("[discard] range filter %s doesn't contain %s", rangeFilter, banIP.String())
+				keep = false
+			}
+		}
+		if reasonFilter != "" {
```
AlteredCoder, line 147: Maybe for this one, it can be helpful to support regex ? so you can search things like this :
```cscsli ban list --reason *scan*```

File: cmd/crowdsec-cli/ban.go [hunk 3]
```diff
@@ -84,7 +87,109 @@ func simpleBanToSignal(targetIP string, reason string, expirationStr string, act
...
+		var keep bool = true
+		var err error
+
+		//try to parse ip or range from ban
+		if ban["iptext"] != "" {
+			if strings.Contains(ban["iptext"], "/") {
+				log.Debugf("%s is a range", ban["iptext"])
+				banIP, banRange, err = net.ParseCIDR(ban["iptext"])
+				if err != nil {
+					log.Warningf("failed to parse '%s' as a range : %s", ban["iptext"], err)
```
AlteredCoder, line 105: Do we want to print this to the user in case of this happening ?
If yes, maybe we should give more details about where this IP come from ?
```
failed to parse '%s' ip found in the backend database as a range ... 
```

cmd/crowdsec-cli/ban.go
buixor: I used "contains" for this purpose, so that `cscsli ban list --reason scan` would do the same. Do you think it's enough ?

AlteredCoder: Ah yes, forget this com

buixor:
TODO : tests `--range` with `ban add range X`

PR #118, opened 2020-07-03
4 lines installer

buixor:
add 4 lines installer to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,23 @@ Besides detecting and stopping attacks in real time based on your logs, it allow
 
 Find the [latest release](https://github.com/crowdsecurity/crowdsec/releases/latest)
 
+Ensure you have dependencies :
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,39 @@ One of the advantages of Crowdsec when compared to other solutions is its crowde
...
+ ```
+</details>
+
+
+
+```bash
+curl -s https://api.github.com/repos/crowdsecurity/crowdsec/releases/latest | grep browser_download_url| cut -d '"' -f 4  | wget -i -
+tar xvzf crowdsec-release.tgz
+cd crowdsec-v*
+sudo ./wizard.sh -i
```
AlteredCoder, line 66: The only problem with RHEL and our wizard is that `cscli` is not in the $PATH. What should we do ?

PR #126, opened 2020-07-07
Doc : fix whitelists documentation + document `data` for parsers/scenarios + document expr helpers

buixor:
- fix whitelist documentation
 - create expr documentation


buixor:
need to be more explicit about its usage in the profiles as well

PR #127, opened 2020-07-08
fix #124

### LINKED ISSUE #124: Bug/crowdsec : in whitelist config, statics are applied even if the whitelist not succeed
**Describe the bug**
In a whitelist configuration, if we want to add a static in `meta`, the static will be applied even if the whitelist not succeed.

**To Reproduce**
```yaml
# cat /etc/crowdsec/config/parsers/s02-enrich/mywhitelists.yaml
name: crowdsecurity/whitelists
description: "Whitelist events from my ip addresses"
debug: true
whitelist:
  reason: "my ip ranges"
  ip:
    - "1.2.3.4"
statics:
  - meta: toto
    value: lol
```

**Expected behavior**
The statics should be applied only if the whitelist is a success

erenJag:
fix #124

### DIFF
--- a/pkg/parser/node.go
+++ b/pkg/parser/node.go
@@ -144,6 +144,8 @@ func (n *Node) process(p *types.Event, ctx UnixParserCtx) (bool, error) {
 	if n.Name != "" {
 		NodesHits.With(prometheus.Labels{"source": p.Line.Src, "name": n.Name}).Inc()
 	}
+	isWhitelisted := false
+	hasWhitelist := false
 	var src net.IP
 	/*overflow and log don't hold the source ip in the same field, should be changed */
 	/* perform whitelist checks for ips, cidr accordingly */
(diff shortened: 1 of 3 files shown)


pkg/parser/node.go
```diff
@@ -298,7 +304,14 @@ func (n *Node) process(p *types.Event, ctx UnixParserCtx) (bool, error) {
...
 	}
-	if len(n.Statics) > 0 {
+	if hasWhitelist && isWhitelisted && len(n.Statics) > 0 {
+		clog.Debugf("+ Processing %d statics", len(n.Statics))
+		// if all else is good in whitelist, process node's statics
+		err := ProcessStatics(n.Statics, p, clog)
+		if err != nil {
+			clog.Fatalf("Failed to process statics : %v", err)
+		}
+	} else if len(n.Statics) > 0 && !hasWhitelist {
```
buixor, line 314: to be picky, can you make it one `if` to avoid copy/pasta the code ?

erenJag: done

PR #136, opened 2020-07-15
Simulation support

buixor:


buixor:
todo : deploy a default config for simulation (that is present in the default.yaml so that user can use the feature without touching configuration files).

PR #138, opened 2020-07-16
fix post merge

AlteredCoder:


buixor:
TBD for mysql user experience :
 - explain to the user how to create the MySQL user, the database and the grants.
 - explain the trick with MySQL > 5.7 and the `SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));` magic

buixor:
TBD: 
```
(/home/bui/github/crowdsec/crowdsec/pkg/database/commit.go:52) 
[2020-07-16 16:50:10]  Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''-43200 minutes')) ORDER BY updated_at desc' at line 1 
ERRO[16-07-2020 16:51:10] error in old records cleanup : failed to get count of old records: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''-43200 minutes')) ORDER BY updated_at desc' at line 1 

```

apparently from :

```golang
func (c *Context) CleanUpRecordsByAge() error {
...
//look for soft-deleted events that are OLDER than maxDurationRetention
	ret := c.Db.Unscoped().Table("ban_applications").Where("deleted_at is not NULL").
		Where(fmt.Sprintf("deleted_at > date('now','-%d minutes')", int(c.maxDurationRetention.Minutes()))).
		Order("updated_at desc").Find(&sos)

	if ret.Error != nil {
		return errors.Wrap(ret.Error, "failed to get count of old records")
	}
```

buixor:
`cscli backup` doesn't save/restore the `simulation.yaml` file

erenJag:
The things left to do have been put in an issue : #139

PR #145, opened 2020-07-24
Add unitest in `pkg/acquisition` and `pkg/cwapi`

AlteredCoder:
This PR add unit test in `pkg/acquisition` (~60% of coverage) and `pkg/cwapi` (~66% of coverage.

It fixes also the way we are using `sling` the `cwapi` package.

It also update the README to add the coverage badge from coveralls.io and rename the two badge `Go` and `build-binary-package` to respectively `tests` and `build`


erenJag:
As a first review, seems good, there is just some linter errors should be checked and fix if needed.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,35 @@
+name: Tests
+
+on:
+  push:
+    branches: [ master , unitest]
```
buixor, line 5: I think we can remove this one before merging

README.md
```diff
@@ -8,8 +8,9 @@
 
 
 <p align="center">
-<img src="https://github.com/crowdsecurity/crowdsec/workflows/Go/badge.svg">
-<img src="https://github.com/crowdsecurity/crowdsec/workflows/build-binary-package/badge.svg">
+<img src="https://github.com/crowdsecurity/crowdsec/workflows/tests/badge.svg">
+<img src="https://github.com/crowdsecurity/crowdsec/workflows/build/badge.svg">
+<a href='https://coveralls.io/github/crowdsecurity/crowdsec?branch=unitest'><img src='https://coveralls.io/repos/github/crowdsecurity/crowdsec/badge.svg?branch=unitest' alt='Coverage Status' /></a>
```
buixor, line 13: I think we need to change branch here before merge

pkg/acquisition/file_reader.go
```diff
@@ -156,7 +156,9 @@ func InitReaderFromFileCtx(files []FileCtx) (*FileAcquisCtx, error) {
 				log.Errorf("error while globing '%s' : %v", fglob, err)
 				return nil, err
 			}
-
+			if len(files) == 0 {
+				return nil, fmt.Errorf("unable to glob path '%s'", fglob)
```
buixor, line 160: if I'm not mistaken, it means that if we have a `acquis.yaml` with at least one glob that doesn't resolve to a file, it's going to prevent crowdsec from starting  :

```bash
$ cat config/acquis.yaml 
filenames:
 - /tmp/test*.log
 - /tmp/noppy.log
labels:
  type: syslog
               $ ./crowdsec -c dev.yaml 
...
              INFO[27-07-2020 11:31:53] Opening file '/tmp/test.log' (pattern:/tmp/test*.log) 
INFO[27-07-2020 11:31:53] Opening file '/tmp/testasd.log' (pattern:/tmp/test*.log) 
FATA[27-07-2020 11:31:53] Error while loading acquisition config : Failed to start acquisition : unable to start file acquisition, bailout unable to glob path '/tmp/noppy.log' 
                                                                                                                                                                          
```

I think a warning or even an error message is fair, but this is a bit too strict.

pkg/cwapi/signals.go
```diff
@@ -57,6 +49,11 @@ func (ctx *ApiCtx) pushSignals() error {
 			return fmt.Errorf("api push signal: return bad HTTP code (%d): %s", resp.StatusCode, string(body))
 		}
 	}
+
+	if resp.StatusCode != 401 && (jsonResp.Message == "" || jsonResp.Message != "OK" || jsonResp.StatusCode != 200) {
```
buixor, line 53: why ignore 401 ?

AlteredCoder: Because `401` mean the token expired. We want to renew it instead of returning an error

pkg/acquisition/file_reader.go
AlteredCoder: fixed

README.md
AlteredCoder: Fixed

PR #151, opened 2020-07-28
tests for pkg/database

buixor:
Those are mostly functional (using SQLite) rather than pure unit-tests, but it seems more relevant.
See `ginkgo.Describe("TestWrites" ...` for mocking of SQL using SQLMock & such.


pkg/database/read.go
```diff
@@ -15,9 +15,8 @@ func (c *Context) GetBansAt(at time.Time) ([]map[string]string, error) {
 	bas := []types.BanApplication{}
 	rets := make([]map[string]string, 0)
 	/*get non-expired records*/
-	//c.Db.LogMode(true)
-	//records := c.Db.Order("updated_at desc").Where(`strftime("%s", until) >= strftime("%s", ?) AND strftime("%s", created_at) < strftime("%s", ?)`, at, at).Group("ip_text").Find(&bas) /*.Count(&count)*/
-	records := c.Db.Order("updated_at desc").Where("until >= ? AND created_at < ?", at, at).Group("ip_text").Find(&bas) /*.Count(&count)*/
+	log.Printf("at is '%s'", at)
```
erenJag, line 18: Set this as debug

PR #183, opened 2020-08-06
add a warning when a grok pattern ends with \n

buixor:
... because it might not be what the user wants and might lead to some weird pattern matching issues :+1:


erenJag:
As you can see the coverage decrease, if it's possible to add a test to coverage this change please, will be perfect :+1:

erenJag (approved):

PR #194, opened 2020-08-28
make cscli use crowdsec version for hub

AlteredCoder:


.github/workflows/ci_hub-tests.yml
```diff
@@ -31,21 +33,26 @@ jobs:
...
         done
     - name: Setup hub ci in crowdsec
-      working-directory: ./crowdsec-xxx/tests/
+      working-directory: ./crowdsec-${{ steps.crowdsec.outputs.tag }}/tests/
       run: |
         cp -R ../../hub-tests/tests .
         cp ../../hub-tests/main .
     - name: Run the HUB CI
-      working-directory: ./crowdsec-xxx/tests/
+      working-directory: ./crowdsec-${{ steps.crowdsec.outputs.tag }}/tests/
```
buixor, line 55: so just to be sure, it means that if the job runs on a MR to master, it will always assume that it's running on the latest release ? (makes sense, just want to be sure)

cmd/crowdsec-cli/install.go
```diff
@@ -63,6 +64,24 @@ you should [update cscli](./cscli_update.md).
...
+				latest, err := cwversion.Latest()
+				if err != nil {
+					log.Fatalf("unable to get last crowdsec version: %s", err)
+				}
+
+				if cwversion.Version == latest {
+					cwhub.HubBranch = "master"
+				} else {
+					log.Warnf("Crowdsec is not the latest version. Current version is '%s' and latest version is '%s'. Please update it!", cwversion.Version, latest)
+					cwhub.HubBranch = cwversion.Version
```
buixor, line 81: maybe we want to specify to which branch we're going to stick if it's not master ?

cmd/crowdsec-cli/update.go
```diff
@@ -20,8 +21,27 @@ Fetches the [.index.json](https://github.com/crowdsecurity/hub/blob/master/.inde
...
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			if !config.configured {
-				return fmt.Errorf("you must configure cli before interacting with hub.")
+				return fmt.Errorf("you must configure cli before interacting with hub")
 			}
+
+			/*
+				if no branch has been specified in flags for the hub, then use the one corresponding to crowdsec version
+			*/
+			if cwhub.HubBranch == "" {
```
buixor, line 30: same as above ^^

buixor (changes requested):
can we merge the 3 checks for last version + warning display to a single one ? besides that lgtm

cmd/crowdsec-cli/install.go
AlteredCoder: I think this is what the ligne above does no ?

.github/workflows/ci_hub-tests.yml
AlteredCoder: If a job runs on a MR to master, it will build crowdsec with the codebase of the current PR but will assume that the crowdsec version is the latest created tag

cmd/crowdsec-cli/update.go
AlteredCoder: Same answer as above

.github/workflows/ci_hub-tests.yml
buixor: cool !

cmd/crowdsec-cli/install.go
AlteredCoder: mmh my bad, it doesn't precise the branch, but says what is the current crowdsec version and what is the latest. Is it enough ?

cmd/crowdsec-cli/install.go
```diff
@@ -63,6 +63,10 @@ you should [update cscli](./cscli_update.md).
 			if !config.configured {
 				return fmt.Errorf("you must configure cli before interacting with hub")
 			}
+
+			if err := setHubBranch(); err != nil {
+				log.Errorf("error while setting hub branch: %s", err)
```
buixor, line 68: I think we want to return the error here, so that the `PersistentPreRunE` can properly fail

buixor (changes requested):
should return the error in the `PersistentPreRunE`, so it can fail properly

buixor (approved):
lgtm

PR #211, opened 2020-09-09
Lapi struct merge

buixor:
- Merge structures with `models`


registergoofy (approved):
I don't see anything obvious missing, except for the unit tests.

PR #218, opened 2020-09-11
add api server

AlteredCoder:


pkg/database/decisions.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	return nil
+}
+
+func (c *Client) DeleteDecisionsWithFilter(filter map[string][]string) (int, error) {
+	var err error
+
+	decisions := c.Ent.Debug().Decision.Delete()
+
+	for param, value := range filter {
```
buixor, line 95: can't we factorize the filter building thing ? they seem to be shared between list and delete operations

buixor (approved):
lgtm, things to be discussed/solved :
 - incoming models validation (gin router vs lower functions)
 - relevance to have the db functions (or a wrapper) accept directly the model to stick as much as possible to api

PR #224, opened 2020-09-14
[WIP][LAPI] Add test for the new scenario version feature

registergoofy:
* add test for scenario version embedding in leakybucket feature
* fix a bug where all checks weren't done when verifying tests were successful
* fix a null-deref which appeared kind of randomly in some corner cases


buixor (approved):
lgtm

PR #227, opened 2020-09-15
add validation in swagger + update code to new generated models

erenJag:
* Add validation in swagger 
* update apiserver pkg to use new generated models
* fix some typos


registergoofy (approved):
LGTM

PR #229, opened 2020-09-16
|wip] Dont 't export HubIdx

registergoofy:
Hubidx is not exported anymore


pkg/cwhub/cwhub.go
```diff
@@ -96,9 +96,40 @@ func getSHA256(filepath string) (string, error) {
 	return fmt.Sprintf("%x", h.Sum(nil)), nil
 }
 
+func GetItemMap(itemType string) map[string]Item {
```
buixor, line 99: I think if we go to the extent of making helpers, shouldn't we have a proper getter so that the other modules avoid iterating over the map to find said scenario/parser/wtver ?

buixor (approved):
let's do one more step

PR #233, opened 2020-09-17
Add mysql and improve

AlteredCoder:


pkg/apiserver/apiserver.go
```diff
@@ -61,7 +61,7 @@ func (s *APIServer) Run() {
 
 	file, err := os.Create(s.logFile)
 	if err != nil {
-		log.Fatalf(err.Error())
+		log.Fatalf("unable to create log file '%s': %s", s.logFile, err.Error())
```
buixor, line 64: not needed RN, but I guess we want to start returning errors here instead of fataling as it will be called from "another" main

pkg/database/alerts.go
```diff
@@ -16,24 +16,165 @@ import (
...
+
+func (c *Client) CreateAlert(alertItem *models.Alert) (string, error) {
 	owner, err := c.QueryMachineByID(alertItem.MachineID)
 	if err != nil {
-		if errors.Cause(err) == UserNotExists {
-			owner = nil
-		} else {
-			return 0, errors.Wrap(QueryFail, fmt.Sprintf("machine '%s': %s", alertItem.MachineID, err))
+		if errors.Cause(err) != UserNotExists {
+			return "", errors.Wrap(QueryFail, fmt.Sprintf("machine '%s': %s", alertItem.MachineID, err))
```
buixor, line 165: not familiar, but should we wrap the actual error ?

buixor (approved):
lgtm, minor remarks

PR #240, opened 2020-09-22
Alerts

registergoofy:
add alerts subcommand to cscli


erenJag:
* You need `add` and `delete` commands.

cmd/crowdsec-cli/alerts.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/crowdsecurity/crowdsec/pkg/cwhub"
+	"github.com/go-openapi/strfmt"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var Scenario, AlertID string
+
+func NewAlertsCmd() *cobra.Command {
+	/* ---- DECISIONS COMMAND */
```
erenJag, line 17: `/* ---- ALERTS COMMAND */`

cmd/crowdsec-cli/main.go
```diff
@@ -135,6 +135,8 @@ API interaction:
 	rootCmd.AddCommand(NewMetricsCmd())
 	rootCmd.AddCommand(NewDashboardCmd())
 	rootCmd.AddCommand(NewDecisionsCmd())
+	rootCmd.AddCommand(NewAlertsCmd())
+	//	rootCmd.AddCommand(NewInspectCmd())
```
erenJag, line 139: this command isn't in the revamp plan I think, can be deleted. (the command func must be kept to use it for the inspect inside other scopes)

registergoofy:
> * You need `add` and `delete` commands.
> * You used decisions filters, you need to look at the alerts filters "on the swagger".

I don't think we want to delete alert. Even `add` alert by cscli is questionable, because alerts are meant to be generated automatically. The only reason I see we want to push alerts to Local API is to push them into the database.

If we add a delete, there's no point in differenciating alerts and decisions...

cmd/crowdsec-cli/main.go
registergoofy: oohh, you are right, I just added the comment to make it build, because the code actually lacked the `NewInspectCmd()` var

cmd/crowdsec-cli/alerts.go
```diff
@@ -0,0 +1,127 @@
...
+
+func NewAlertsCmd() *cobra.Command {
+	/* ---- ALERTS COMMAND */
+	var cmdAlerts = &cobra.Command{
+		Use:   "alerts [action]",
+		Short: "Manage alerts",
+		Long: `
+Alerts Management.
+
+To list/add/delete decisions
```
buixor, line 25: *alerts

cmd/crowdsec-cli/alerts.go
```diff
@@ -0,0 +1,127 @@
...
+				log.Fatalf("failed to parse Local API URL %s : %v ", csConfig.LapiClient.Credentials.Url, err.Error())
+			}
+
+			//initialize cwhub
+			if err := cwhub.GetHubIdx(csConfig.Cscli); err != nil {
+				log.Fatalf("Failed to load hub index : %s", err)
+			}
+
+			scenarios := []string{}
+			for _, scenario := range cwhub.GetItemMap(cwhub.SCENARIOS) {
```
buixor, line 45: should use the function that was added to cwhub : GetUpstreamInstalledScenariosAsString

cmd/crowdsec-cli/alerts.go
```diff
@@ -0,0 +1,127 @@
...
+				log.Fatalf("Unable to list alerts : %v", err.Error())
+			}
+
+			err = AlertsToTable(alerts)
+			if err != nil {
+				log.Fatalf("unable to list alerts : %v", err.Error())
+			}
+		},
+	}
+	cmdAlertsList.Flags().StringVar(&Scope, "scope", "", "scope to which the decision applies (ie. IP/Range/Username/Session/...)")
```
buixor, line 115: can you point the flags directly to the right struct members of the filter ?

cmd/crowdsec-cli/decisions.go
```diff
@@ -94,11 +96,21 @@ To list/add/delete decisions
 				log.Fatalf("failed to parse Local API URL %s : %v ", csConfig.LapiClient.Credentials.Url, err.Error())
 			}
 
+			scenarios := []string{}
+			for _, scenario := range cwhub.GetItemMap(cwhub.SCENARIOS) {
```
buixor, line 100: same as before, use GetUpstreamInstalledScenariosAsString

buixor (changes requested):
wip

pkg/apiclient/alerts_service.go
```diff
@@ -63,3 +77,24 @@ func (s *AlertsService) List(ctx context.Context, opts AlertsListOpts) (*models.
...
 }
+
+//to demo query arguments
+func (s *AlertsService) Delete(ctx context.Context, opts AlertsDeleteOpts) (*models.DeleteAlertsResponse, *Response, error) {
+	var alerts models.DeleteAlertsResponse
+	params, err := qs.Values(opts)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("alerts/?%s", params.Encode())
```
erenJag, line 88: `fmt.Sprintf("alerts?%s", params.Encode())`

cmd/crowdsec-cli/decisions.go
registergoofy: > same as before, use GetUpstreamInstalledScenariosAsString

yes, sure. But we have to get this merged before. This feature is not in this branch.

cmd/crowdsec-cli/alerts.go
registergoofy: sure, but the feature is not there yet. I suggest to merge this before, and add GetUpstreamInstalledScenariosAsString in an other PR afterwards.

erenJag (approved):

PR #258, opened 2020-09-28
[Wip] lapi integration

buixor:
- Makefile(s) build api as well
 - Fix debug levels
 - Fix models.Alert build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DATA_PREFIX = $(PREFIX)"/var/run/crowdsec/"
 PID_DIR = $(PREFIX)"/var/run/"
 CROWDSEC_FOLDER = "./cmd/crowdsec"
 CSCLI_FOLDER = "./cmd/crowdsec-cli/"
+API_FOLDER= "./cmd/api/"
 CROWDSEC_BIN = "crowdsec"
 CSCLI_BIN = "cscli"
 BUILD_CMD="build"
@@ -64,6 +65,15 @@ else
 	@echo "Required golang version is $(REQUIRE_GOVERSION). The current one is $(CURRENT_GOVERSION). Exiting.."
 	@exit 1;
 endif
+
+api:
+ifeq ($(lastword $(RESPECT_VERSION)), $(CURRENT_GOVERSION))
+	@make -C $(API_FOLDER) build --no-print-directory
+else
+	@echo "Required golang version is $(REQUIRE_GOVERSION). The current one is $(CURRENT_GOVERSION). Exiting.."
+	@exit 1;
+endif
+
 
 
 crowdsec:
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ RELDIR = crowdsec-$(BUILD_VERSION)
 
 all: clean test build
 
-build: clean goversion crowdsec cscli
+build: clean goversion crowdsec cscli api
 
 static: goversion crowdsec_static cscli_static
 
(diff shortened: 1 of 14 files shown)


cmd/crowdsec/main.go
```diff
@@ -357,18 +357,9 @@ func main() {
...
-	// 	if len(parsers.nodes) == 0 {
-	// 		log.Fatalf("no parser(s) loaded, abort.")
-	// 	}
-
-	// 	if len(holders) == 0 {
-	// 		log.Fatalf("no bucket(s) loaded, abort.")
-	// 	}
-	// }
+	if cConfig.API == nil || cConfig.API.Client == nil || cConfig.API.Client.Credentials == nil {
+		log.Fatalf("Missing Client API credentials, abort.")
```
erenJag, line 361: We need to be more specific on which config is missing (local or online API credentials)

pkg/leakybucket/overflows.go
```diff
@@ -126,73 +190,61 @@ func NewAlert(leaky *Leaky, queue *Queue) types.RuntimeAlert {
...
+	if err != nil {
+		return runtimeAlert, errors.Wrap(err, "unable to collect sources from bucket")
+	}
+	runtimeAlert.Sources = sources
+	//Include source info in format string
+	sourceStr := ""
+	if len(sources) > 1 {
+		sourceStr = fmt.Sprintf("%d Sources on scope.", len(sources))
+	} else if len(sources) == 1 {
+		for k, _ := range sources {
```
registergoofy, line 214: Use `for k := range sources` here

pkg/parser/enrich.go
```diff
@@ -44,7 +44,6 @@ func Loadplugin(path string) ([]EnricherCtx, error) {
 	if err != nil {
 		log.Warningf("load (fake) plugin load : %v", err)
 		c.initiated = false
-		return nil, err
```
registergoofy, line 47: Why ? I think we should fatal if this fails ?

registergoofy (changes requested):
minor changes.

Makefile
```diff
@@ -6,6 +6,7 @@ DATA_PREFIX = $(PREFIX)"/var/run/crowdsec/"
 PID_DIR = $(PREFIX)"/var/run/"
 CROWDSEC_FOLDER = "./cmd/crowdsec"
 CSCLI_FOLDER = "./cmd/crowdsec-cli/"
+API_FOLDER= "./cmd/api/"
```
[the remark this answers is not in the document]
buixor: yes it is for testing purpose, it's going to get removed once the mains are merge

cmd/crowdsec/output.go
```diff
@@ -138,9 +137,10 @@ LOOP:
 				return fmt.Errorf("postoverflow failed : %s", err)
 			}
 
-			if *event.Overflow.Alert.Scenario == "" && event.Overflow.Mapkey != "" {
+			if event.Overflow.Alert == nil && event.Overflow.Mapkey != "" {
 				buckets.Bucket_map.Delete(event.Overflow.Mapkey)
-			} else {
+			} else if event.Overflow.Alert != nil {
+				log.Printf("%s", *event.Overflow.Alert.Message)
```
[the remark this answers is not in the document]
buixor: no, it really displays (Printf == Infof) :

```
INFO[28-09-2020 12:09:00] Ip 127.0.0.1 performed 'crowdsecurity/http-bad-user-agent' (2 events over 0s) at 2020-07-21 15:48:22 +0200 CEST 
```

pkg/parser/enrich.go
buixor: no, the plugins are "attempted" to be loaded even if the parser itself isn't enabled.
It isn't very elegant but we have to deal with this until we transform the enrichment into proper modules/plugins ^^

PR #261, opened 2020-09-29
add randomness to machine-id when registering.

registergoofy:
* add randomness to machine-id when registering. This is to work around shameless cloud provider that provide all vps with the same machine-id...


buixor (approved):
lgtm

PR #262, opened 2020-09-29
[lapi] cwhub fixes

buixor:
Now that crowdsec sends scenario's version and hash along with the signal to API, `pkg/leakybucket` needs to be able to find back a scenario in cwhub from the scenario itself. Add a helper function that will retrieve the cwhub item based on file path.

### DIFF
--- a/pkg/leakybucket/manager_load.go
+++ b/pkg/leakybucket/manager_load.go
@@ -181,14 +181,11 @@ func LoadBuckets(cscfg *csconfig.CrowdsecServiceCfg, files []string) ([]BucketFa
 			bucketFactory.BucketName = seed.Generate()
 			bucketFactory.ret = response
 			hubItem, err := cwhub.GetItemByPath(cwhub.SCENARIOS, bucketFactory.Filename)
-			if err != nil {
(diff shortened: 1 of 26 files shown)


pkg/leakybucket/manager_load.go
```diff
@@ -180,12 +180,12 @@ func LoadBuckets(cscfg *csconfig.CrowdsecServiceCfg, files []string) ([]BucketFa
 			bucketFactory.Filename = filepath.Clean(f)
 			bucketFactory.BucketName = seed.Generate()
 			bucketFactory.ret = response
-			hubItem := cwhub.GetItem(cwhub.SCENARIOS, bucketFactory.Name)
+			hubItem, err := cwhub.GetItemByPath(cwhub.SCENARIOS, bucketFactory.Filename)
```
registergoofy, line 183: add a check on err here.

registergoofy (approved):
Can you just fix 
`hubItem, err := cwhub.GetItemByPath(cwhub.SCENARIOS, bucketFactory.Filename)`
by checking err in pkg/leakybucket/manager_load.go#L183

PR #267, opened 2020-09-30
fix initialization stuff

registergoofy:
* get rid of unused ConfigFilePath check (better done with cobra's PersistentFlags)
* initialize logger level before using logger


buixor (approved):
lgtm

PR #281, opened 2020-10-01
Minor changes

AlteredCoder:
- review `cscli alerts list` output table
- stop loading cwhub in cscli when we don't need it
- add possibility to `force` when using `cscli watchers add` so it will replace the watcher if exist
- reimport api test read/write code
- fix a bug in apiclient pkg
- update machine IP in database for a watcher if the IP is not known
- don't use hardcoder password in wizard.sh


erenJag (approved):

cmd/crowdsec-cli/alerts.go
```diff
@@ -22,14 +21,16 @@ var ActiveDecision bool
 
 func AlertsToTable(alerts *models.GetAlertsResponse) error {
 	if csConfig.Cscli.Output == "raw" {
-		fmt.Printf("id,Scope/Value,reason,version,message,country,as,events_count,created_at\n")
+		fmt.Printf("id,Scope/Value,reason,version,country,as,events_count,created_at\n")
```
erenJag, line 24: Could you please just remove the version in this mode because it's already in the reason.

PR #288, opened 2020-10-04
fix wip_lapi bugs

AlteredCoder:
This PR fixes : 

- a null deref in apiclient/auth.go when the crowdsec API is not up
- The `StreamDecision` function in controller/decisions.go

### DIFF
--- a/pkg/apiclient/auth.go
+++ b/pkg/apiclient/auth.go
@@ -44,7 +44,7 @@ func (t *APIKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	// Make the HTTP request.
 	resp, err := t.transport().RoundTrip(req)
-	if resp == nil {
+	if err != nil {
 		log.Errorf("resp-api: auth with api key failed return nil response, error: %s", err)
 		return resp, err
 	}
--- a/pkg/apiclient/auth.go
+++ b/pkg/apiclient/auth.go
@@ -44,6 +44,10 @@ func (t *APIKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	// Make the HTTP request.
 	resp, err := t.transport().RoundTrip(req)
+	if resp == nil {
+		log.Errorf("resp-api: auth with api key failed return nil response, error: %s", err)
+		return resp, err
+	}
 	if log.GetLevel() >= log.TraceLevel {
 		dump, _ := httputil.DumpResponse(resp, true)
 		log.Tracef("resp-api: %s", string(dump))
--- a/pkg/apiserver/controllers/decisions.go
+++ b/pkg/apiserver/controllers/decisions.go
@@ -92,55 +92,87 @@ func (c *Controller) DeleteDecisions(gctx *gin.Context) {
 
 func (c *Controller) StreamDecision(gctx *gin.Context) {
 	var data []*ent.Decision
-
 	ret := make(map[string][]*models.Decision, 0)
 	ret["new"] = []*models.Decision{}
 	ret["deleted"] = []*models.Decision{}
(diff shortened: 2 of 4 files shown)


pkg/apiclient/auth.go
```diff
@@ -44,6 +44,10 @@ func (t *APIKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	}
 	// Make the HTTP request.
 	resp, err := t.transport().RoundTrip(req)
+	if resp == nil {
```
registergoofy, line 47: idiomatic way of doing this is testing err being no-nil. Are you doing something different, here ?

pkg/apiserver/controllers/decisions.go
```diff
@@ -92,55 +92,87 @@ func (c *Controller) DeleteDecisions(gctx *gin.Context) {
...
+	// if the blocker just start, return all decisions
+	if val, ok := gctx.Request.URL.Query()["startup"]; ok {
+		if val[0] == "true" {
+			data, err := c.DBClient.QueryAllDecisions()
+			if err != nil {
+				log.Errorf("failed querying decisions: %v", err)
+				gctx.JSON(http.StatusInternalServerError, gin.H{"message": err.Error()})
+				return
+			}
+			ret["new"], err = FormatDecisions(data)
```
registergoofy, line 126: Please check the err value. Or use _

pkg/apiserver/controllers/decisions.go
```diff
@@ -92,55 +92,87 @@ func (c *Controller) DeleteDecisions(gctx *gin.Context) {
...
-		gctx.JSON(http.StatusInternalServerError, gin.H{"message": err.Error()})
+		if _, ok := err.(*ent.NotFoundError); ok {
+			gctx.JSON(http.StatusForbidden, gin.H{"message": err.Error()})
+		} else {
+			gctx.JSON(http.StatusInternalServerError, gin.H{"message": "not allowed"})
+		}
+		return
 	}
 
-	lastPullTime, err := time.Parse(time.RFC3339, lastPull)
```
registergoofy, line 122: If err is not worth checking, please use _

registergoofy (commented):
Minor modification:
* error checking. Maybe it's even worth having a look at `errors.Wrap`
* renaming blocker in variable by bouncer

pkg/apiclient/auth.go
AlteredCoder: i didn't wrote this code. But my code was crashing because `resp` was equal to `nil`.

pkg/apiserver/controllers/decisions.go
AlteredCoder: the part that you are commenting has been removed

pkg/apiserver/controllers/decisions.go
```diff
@@ -92,56 +92,102 @@ func (c *Controller) DeleteDecisions(gctx *gin.Context) {
...
-	if err != nil {
-		gctx.JSON(http.StatusInternalServerError, gin.H{"message": err.Error()})
+	if blockerInfo == nil {
+		gctx.JSON(http.StatusInternalServerError, gin.H{"message": "not allowed"})
+		return
+	}
+
+	// if the blocker just start, return all decisions
+	if val, ok := gctx.Request.URL.Query()["startup"]; ok {
```
buixor, line 119: you might want to check that the key is present in the map to avoid deref ?

buixor (approved):
lgtm

pkg/apiserver/controllers/decisions.go
AlteredCoder: this is what the code does no ?

PR #295, opened 2020-10-06
Srcr postgresql

srcr:
Added PostgreSQL support in the database plugin.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/jinzhu/gorm"
 	_ "github.com/jinzhu/gorm/dialects/mysql"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
 	_ "github.com/jinzhu/gorm/dialects/sqlite"
 	_ "github.com/mattn/go-sqlite3"
 	"gopkg.in/tomb.v2"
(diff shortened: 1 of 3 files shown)


AlteredCoder:
Hello @srcr , thanks for contributing to the project !
We are going to review and test your PR and then will be able to merge it :)

Thank you!

pkg/database/database.go
```diff
@@ -50,6 +51,22 @@ func checkConfig(cfg map[string]string) error {
 		if val, ok := cfg["db_name"]; !ok || val == "" {
 			return fmt.Errorf("please specify a 'db_name' to MySQL db in the configuration")
 		}
+	case "postgres":
```
AlteredCoder, line 54: since postgres configuration is the same that mysql one, i think you can use the mysql `case` instead of rewriting all the checks like:

```go
case "mysql", "postgres":
     ....
```

and we can then change all the error messages with :

```go
fmt.Errorf("please specify a 'db_name' to %s db in the configuration", dbType)
...
fmt.Errorf("please specify a 'db_host' to %s db in the configuration", dbType)
...

```

srcr: I'm just a simple spaghetti coder :) I can make it work but the code is not that pretty. Your solution is a lot cleaner.
I'm currently running with my code and it works fine except for cscli ban list which throws :
```
FATA[0000] failed to list bans : unable to get records from Database : pq: column "ban_applications.id" must appear in the GROUP BY clause or be used in an aggregate function
```

SQL log:
```
Oct  6 19:17:07  postgres[46544]: [3-1] ERROR:  column "ban_applications.id" must appear in the GROUP BY clause or be used in an aggregate function at character 8
Oct  6 19:17:07  postgres[46544]: [3-2] STATEMENT:  SELECT * FROM "ban_applications"  WHERE "ban_applications"."deleted_at" IS NULL AND ((until >= $1)) GROUP BY ip_text ORDER BY updated_at desc
```

registergoofy:
Thank you for your contribution !

We are on the way on rewriting a bunch of code. Database management has been completely reworked, and we even changed the backend (no more goorm). We are about to release a release candidate.

So we're sorry but as this postresql feature doesn't work, we are not at ease to merge it to master as is right now. Please be patient and see the new postgresql feature in the next following week.

PR #299, opened 2020-10-08
wip/lapi : respect remediation flag in scenario

buixor:
Do not generate default decisions in the alert that is pushed to APIL if the scenario doesn't have remediation set to "true" in the labels.


registergoofy (approved):
lgtm

PR #300, opened 2020-10-08
refresh API Swagger model

AlteredCoder:
Question:

 - Do we still want to have the route for deleting watchers?
 - Do we still want to the `/ping` route ?

### DIFF
--- a/pkg/models/localapi_swagger.yaml
+++ b/pkg/models/localapi_swagger.yaml
@@ -41,23 +41,6 @@ paths:
         '400':
           description: Invalid filters
           schema: {}
-    delete:
-      description: Returns a list of new/expired decisions. Intended for blockers that need to "stream" decisions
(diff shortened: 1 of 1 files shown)


erenJag:
For me,
* we should delete the method that allow a watcher to delete himself because there is no logical user story in our context to permit that.
* `/ping` can be deleted too (unused).

Don't forget to regenerate the models with this new swagger file.

buixor:
> * we should delete the method that allow a watcher to delete himself because there is no logical user story in our context to permit that.

I think for multi-machines setup (one APIL, several Crowdsec) it makes sense. And it can make sense as well at the APIC level and could be called when the user uninstalls crowdsec.

erenJag:
oh yes my bad, so agree for watchers deletion.

buixor:
deprecated done by other MR

PR #302, opened 2020-10-11
Wip/lapi catch panic

buixor:
add a defer'ed function to catch panic in each and all go-routines. According to what I've read we can't catch it globally. Could add a wrapper, but as we use both tombs and direct goroutines, it would be painful :)


registergoofy (commented):
lgtm

PR #309, opened 2020-10-19
[wip/lapi] profiles

buixor:


pkg/models/helpers.go
```diff
@@ -0,0 +1,12 @@
+package models
+
```
buixor, line 2: can we move this to exprhelpers ?

PR #311, opened 2020-10-19
Update README.md

donatj:
> image

As it is, it is not immediately apparent that the Redhat specific instructions are *not* the curl line below the `for RedHat based distributions` heading.

This cleans up the installation instructions a little to look like

> image


buixor (approved):
lgtm, thanks for your contribution !

PR #313, opened 2020-10-20
pkg/cwhub: fix dropped error

alrs:
An error variable was being passed in to `parser_visit()`, and then ignored. This is a fix.


buixor:
Hello,

sorry for the huge delay :sweat_smile: this is kinda outdated and was breaking stuff, closing.

PR #315, opened 2020-10-20
[wip_lapi] make apil communicate with apic

AlteredCoder:
- add code in apil to communicate with APIC
- fix bug in apiserver
- make apiserver catch user agent to update machines/bouncers version and name
- make useragent configurable by client in `apiclient`
- display version and type for bouncers/machines in cscli


cmd/crowdsec/api.go
```diff
@@ -17,26 +17,22 @@ func initAPIServer() (*apiserver.APIServer, error) {
...
+func serveAPIServer(apiServer *apiserver.APIServer) error {
 	apiTomb.Go(func() error {
-		defer types.CatchPanic("serveAPIServer")
-		log.Info("local API server starting")
-		<-apiTomb.Dying() // lock until go routine is dying
-		if err := apiServer.Shutdown(); err != nil {
-			log.Fatalf(err.Error())
-		}
+		defer types.CatchPanic("crowdsec/serveAPIServer")
+		go func() {
```
buixor, line 23: do we need the go-routine just here within the tomb ? isn't it redundant ?

cmd/crowdsec/serve.go
```diff
@@ -218,21 +218,19 @@ func Serve() error {
...
 		if err != nil {
-			return fmt.Errorf("unable to init api server: %s", err)
+			return errors.Wrap(err, "api server init")
 		}
 		if !cConfig.Crowdsec.LintOnly {
-			err := runAPIServer(apiServer)
-			if err != nil {
-				return fmt.Errorf("unable to run api server: %s", err)
+			if err := serveAPIServer(apiServer); err != nil {
+				return err
```
buixor, line 225: maybe a wrapf here :)

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,231 @@
...
+		return ret, errors.Wrapf(err, "parse local API URL '%s': %v ", config.Credentials.URL, err.Error())
+	}
+
+	password := strfmt.Password(config.Credentials.Password)
+	t := &apiclient.JWTTransport{
+		MachineID: &config.Credentials.Login,
+		Password:  &password,
+	}
+
+	pullInterval, err := time.ParseDuration(PullInterval)
```
buixor, line 57: You don't want to make it part of OnlineApiClientCfg ? (same goes for pushInterval and MetricsInterval)

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,231 @@
...
+		select {
+		case <-a.pushTomb.Dying(): // if one apic routine is dying, do we kill the others?
+			a.pullTomb.Kill(nil)
+			a.metricsTomb.Kill(nil)
+			log.Infof("push tomb is dying, sending cache (%d elements) before exiting", len(cache))
+			err := a.Send(cache)
+			return err
+		case <-ticker.C:
+			// flush
+			a.mu.Lock()
```
buixor, line 102: for the lock dance I think we want to minimize the lock time :
```golang
Lock()
copy := existing
Unlock()
Send()
```

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,231 @@
...
+				return errors.Wrap(err, "pull top")
+			}
+			if a.startup {
+				a.startup = false
+			}
+			for _, decision := range data.New {
+				alertCreated, err := a.dbClient.Ent.Alert.
+					Create().
+					SetScenario(*decision.Scenario).
+					SetSourceIp(*decision.Value).
```
buixor, line 140: I thin we want to set origin as consensus or something like that? It will allow user to easily see the decisions that are coming from the consensus

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,231 @@
...
+					SetScope(*decision.Scope).
+					SetOwner(alertCreated).Save(a.dbClient.CTX)
+				if err != nil {
+					log.Fatalf("failed creating decision from top: %v", err)
+				}
+			}
+			log.Printf("pull top: added %d entries", len(data.New))
+
+			var filter map[string][]string
+			for _, decision := range data.Deleted {
```
buixor, line 174: I think we want to process `Deleted` before `Added` in case we would have a deleted and added decision on a similar scope within the same pull ?

buixor (changes requested):
cool wip, I think I missed the monitoring of tombs for push/pull/metrics/ so that we can shutdown when one of this routines dies

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,256 @@
...
+				alertCreated, err := a.dbClient.Ent.Alert.
+					Create().
+					SetScenario(*decision.Scenario).
+					SetSourceIp(*decision.Value).
+					//SetSourceAsNumber(alert["as_num"]).
+					//SetSourceAsName(alert["as_org"]).
+					//SetSourceCountry(alert["country"]).
+					Save(a.dbClient.CTX)
+				if err != nil {
+					log.Fatalf("unable to create alors from topX: %s", err)
```
buixor, line 172: maybe we want to return error here instead ?

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,256 @@
...
+				}
+
+				duration, err := time.ParseDuration(*decision.Duration)
+				if err != nil {
+					log.Fatalf("unable to parse decision duration '%s': %s", *decision.Duration, err)
```
buixor, line 177: maybe we want to return error here instead ?

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,256 @@
...
+				}
+				startIP, endIP, err := controllers.GetIpsFromIpRange(*decision.Value)
+				if err != nil {
+					log.Fatalf("failed querying alerts: Range %v is not valid", *decision.Value)
```
buixor, line 181: maybe we want to return error here instead ?

pkg/apiserver/apic.go
```diff
@@ -0,0 +1,256 @@
...
+					SetScenario(*decision.Scenario).
+					SetType(*decision.Type).
+					SetStartIP(startIP).
+					SetEndIP(endIP).
+					SetValue(*decision.Value).
+					SetScope(*decision.Scope).
+					SetOrigin("crowdsec-api").
+					SetOwner(alertCreated).Save(a.dbClient.CTX)
+				if err != nil {
+					log.Fatalf("failed creating decision from top: %v", err)
```
buixor, line 195: maybe we want to return error here instead ?

pkg/apiserver/middlewares/v1/api_key.go
```diff
@@ -89,6 +91,22 @@ func (a *APIKey) MiddlewareFunc() gin.HandlerFunc {
 			}
 		}
 
+		useragent := strings.Split(c.Request.UserAgent(), "/")
+
+		if len(useragent) != 2 {
+			log.Errorf("bad user agent from ('%s') : %s", useragent, c.ClientIP())
+			c.JSON(http.StatusForbidden, gin.H{"message": "bad user agent"})
```
buixor, line 98: maybe it's a bit aggressive to return a 403 if we don't like the user-agent format ? can't we just silently skip it ?

buixor (changes requested):
minor changes

PR #322, opened 2020-10-22
Wip/lapi cscli dedup

buixor:
- To make `cscli decisions list` output more friendly & concise, merge decisions on `{source,type,scope}` to avoid repeating items


AlteredCoder (commented):
The code seems ok :+1:
But it looks like you don't keep the decisions with the longest expiration ?

PR #323, opened 2020-10-22
added detail regarding platforms/architectures

8go:
- I made the mistake of following your instructions without thinking too much
- the installation failed. I quickly realized it was because your installation only works on amd64 platform
- to avoid that others make the same mistake I added a few lines to clarify and point other platform users into the right direction


erenJag (approved):
Thanks for you contribution. Indeed we forgot specifying that sorry!
We'll create an issue regarding other platforms to be available in the next releases.

PR #327, opened 2020-10-23
cscli : properly honor the --force flag of upgrade

buixor:
- honor the `--force` flag of `cscli upgrade` to avoid user having to force install over


registergoofy (approved):
lgtm

PR #328, opened 2020-10-23
check if delete alert is from localhost

AlteredCoder:


buixor:
I think we want to amend the `cscli alerts delete` command to indicate it

buixor (approved):
lgtm

Issue #333, opened 2020-10-27
Improvement/Makefile Build process should check for a minimum of Go 1.13


zathras777:
When building using make build, the version of Go should be checked.

My initial attempt at building was on an older system with Go 1.10 installed which led to a failed build. As a minimum of Go 1.13 appears to be required it would have been far better to have this highlighted by a simple check in the makefile.

Also, I totally missed the 1.13 requirement in the docs prior to building, so maybe it should be clearer?

buixor:
Thanks for the report, 110% agreed :)

buixor:
Hello,

Out of curiosity, what `Makefile` command did you use to build? Was going to fix it and I thought that a lot of the targets already check for golang version (ie. https://github.com/crowdsecurity/crowdsec/blob/master/Makefile#L62)

Thanks!

zathras777:
I followed the instructions and did

$ make build
rm -f crowdsec
CURRENT_GOVERSION="1.15.2"
RESPECT_VERSION=" 1.13"

While this is good and makes sense, when the version was lower it should
probably have stopped? With the older version it simply carried on and
promptly failed.

buixor:
fixed on master, thanks !

PR #338, opened 2020-10-29
minor improvements : makefile + `cscli api register`

buixor:
- Properly check in makefile the golang version (https://github.com/crowdsecurity/crowdsec/issues/333)
 - Make `cscli api register` inform the user of "where" to store those credentials if launched "by hand"


erenJag (approved):
lgtm

PR #349, opened 2020-11-01
Adjust the pattern for modsecurity entries

zathras777:
Entries related to modsecurity weren't being parsed until the following changes were made.


buixor:
Hello,

Thanks for the bug-report and suggestion ! Would you mind sharing some sample modsec logs so that we can integrate it to tests?

If you don't mind, we'll re-open the MR on the https://github.com/crowdsecurity/hub, so that we can update the [modsec parser](https://github.com/crowdsecurity/hub/blob/master/parsers/s01-parse/crowdsecurity/modsecurity.yaml) directly, allowing people to have the fixed one without having to get a new version of crowdsec (we are trying to move away from the patterns distributed with crowdsec itself, as it makes it harder to update those).

PR #362, opened 2020-11-04
Wip/lapi : `cscli machines` : honor the `--force` flag

buixor:
If the target file already exists, only overwrite it if the `--force` flag is specified

### DIFF
--- a/cmd/crowdsec-cli/machines.go
+++ b/cmd/crowdsec-cli/machines.go
@@ -169,6 +169,7 @@ cscli machines add MyTestMachine --password MyPassword
 			}
 		},
 		Run: func(cmd *cobra.Command, args []string) {
+			var dumpFile string
 			var err error
 
 			// create machineID if doesn't specified by user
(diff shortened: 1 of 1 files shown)


cmd/crowdsec-cli/machines.go
```diff
@@ -211,7 +227,6 @@ cscli machines add MyTestMachine --password MyPassword
 			}
 			log.Infof("Machine '%s' created successfully", machineID)
 
-			var dumpFile string
 			if outputFile != "" {
```
erenJag, line 230: there is duplicated code here.
```golang
                        if outputFile != "" {
				dumpFile = outputFile
			} else if csConfig.API.Client.CredentialsFilePath != "" {
				dumpFile = csConfig.API.Client.CredentialsFilePath
			} else {
				dumpFile = ""
			}
```

buixor: yes nice spot

erenJag (approved):
:+1:

PR #367, opened 2020-11-05
errors should be handle with the errors model

erenJag:
errors are handled by the apiclient following the errorResponse model generate by the swagger to have same errors format on CAPI and LAPI.


buixor (approved):
lgtm

PR #376, opened 2020-11-06
New metabase

AlteredCoder:


pkg/metabase/metabase.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+var (
+	metabaseDefaultUser     = "crowdsec@crowdsec.net"
+	metabaseDefaultPassword = "!!Cr0wdS3c_M3t4b4s3??"
+	containerName           = "/crowdsec-metabase"
+	metabaseImage           = "metabase/metabase"
+	containerSharedFolder   = "/metabase-data"
+
+	metabaseSQLiteDBURL = "https://crowdsec-statics-assets.s3-eu-west-1.amazonaws.com/metabase_sqlite.zip"
```
buixor, line 47: I think we can get rid of those ?

buixor (approved):
minor remarks but good

PR #383, opened 2020-11-07
support json and raw for cscli alerts inspect

AlteredCoder:


buixor (commented):
dup with `pimp cscli config ` mr

PR #386, opened 2020-11-08
lapi error management

buixor:
- favor `errors.Wrapf`
 - be sure to display original error if we're to replace it with a generic one for API client


registergoofy (approved):
lgtm

AlteredCoder (approved):

PR #396, opened 2020-11-09
fix cscli alerts/decisions list

AlteredCoder:
- add alert id in cscli decisions list
- add a limit param for cscli alerts list
- print alerts order by date (desc) in cscli alerts list

### DIFF
--- a/cmd/crowdsec-cli/alerts.go
+++ b/cmd/crowdsec-cli/alerts.go
@@ -21,6 +21,7 @@ import (
 )
 
 var printMachine bool
+var limit int
 
 func DecisionsFromAlert(alert *models.Alert) string {
 	ret := ""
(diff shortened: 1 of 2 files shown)


cmd/crowdsec-cli/alerts.go
```diff
@@ -315,6 +323,7 @@ cscli alerts list --type ban`,
 	cmdAlertsList.Flags().StringVar(alertListFilter.ScopeEquals, "scope", "", "restrict to alerts of this scope (ie. ip,range)")
 	cmdAlertsList.Flags().StringVarP(alertListFilter.ValueEquals, "value", "v", "", "the value to match for in the specified scope")
 	cmdAlertsList.Flags().BoolVarP(&printMachine, "machine", "m", false, "print machines that sended alerts")
+	cmdAlertsList.Flags().IntVarP(&limit, "limit", "l", 50, "limit size of alerts list table")
```
buixor, line 326: does it means that by side effect putting `0` will be infinite limit ?

buixor (approved):
lgtm, question tho

PR #414, opened 2020-11-11
Fix wizard unattended mode

buixor:


registergoofy (commented):
seems like a bunch of indentation fix + installation of linux default collection

Is the OS detection still relevant then ?

registergoofy (approved):
lgtm

PR #415, opened 2020-11-11
fix wizard: not print cscli error on dashboard remove

erenJag:


wizard.sh
```diff
@@ -378,7 +378,7 @@ delete_bins() {
 # uninstall crowdsec and cscli
 uninstall_crowdsec() {
     systemctl stop crowdsec.service
-    ${CSCLI_BIN} dashboard remove -f -y || echo "metabase not running"
+    ${CSCLI_BIN} dashboard remove -f -y 2> /dev/null || echo "metabase not running"
```
buixor, line 381: I think we might instead want dashboard remove not to `fatalf` if the dashoard isn't present

buixor (changes requested):
suggestin

AlteredCoder (approved):

PR #434, opened 2020-11-13
fix database gc

AlteredCoder:
- Fix database gc with MaxAge
- Add default reason in cscli
- Fix `cscli config show` to show db flush config


erenJag (approved):
lgtm

PR #438, opened 2020-11-16
add docker support

erenJag:


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN make release
+
+WORKDIR /go/src/crowdsec
+RUN /bin/bash wizard.sh --docker-mode
+RUN sed -ri 's/^(\s*)(daemonize\s*:\s*true\s*$)/\1daemonize: false/' /etc/crowdsec/config.yaml
+RUN sed -ri 's/^(\s*)(log_media\s*:\s*file\s*$)/\1log_media: stdout/' /etc/crowdsec/config.yaml
+RUN cscli hub update
+RUN cscli collections install crowdsecurity/linux
+
+RUN rm -rf /go/src/crowdsec
```
registergoofy, line 21: When the Dockerfile is run, each command creates a new layer. It's possible to extract each layer, it's not a good idea to try to hide files used to build the the software, because they are still accessible.

Good way to do this, is to use multi-stage Dockerfile. It's not a lot of work, I think it's worth it.

denisgolius:
@erenJag 

1. Is there any wiki/example how to use it with nginx in docker? 
2. Also I'm interesting for how to use it if nginx write logs in stdout?

erenJag:
Hi @denisgolius,

> 1. Is there any wiki/example how to use it with nginx in docker?

This Dockerfile and it's documentation are still in development and will be improved in the next days. It also only going to work from the next release which is in the target branch of this PR for the moment.

> 2. Also I'm interesting for how to use it if nginx write logs in stdout?

Our advise is to wait for the next release to test this feature. Before that, we can already tell you that you need to do some magic with docker-compose to have crowdsec running with others containers sharing the same log files etc.

Maybe a user guide will be published about that :wink:

buixor (changes requested):
minor remarks but overall good wip :+1:

buixor (approved):
lgtm

PR #440, opened 2020-11-16
have some random seed for jwt generation

registergoofy:
enforce random seed for jwt generation by default. seed is changed at each restart.

### DIFF
--- a/pkg/apiserver/middlewares/v1/jwt.go
+++ b/pkg/apiserver/middlewares/v1/jwt.go
@@ -147,7 +147,7 @@ func NewJWT(dbClient *database.Client) (*JWT, error) {
 		secret []byte
 	)
 
-	secret_string := os.Getenv("SECRET")
+	secret_string := os.Getenv("CS_LAPI_SECRET")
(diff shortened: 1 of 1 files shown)


pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -142,17 +143,32 @@ func Unauthorized(c *gin.Context, code int, message string) {
...
 func NewJWT(dbClient *database.Client) (*JWT, error) {
 	// Get secret from environment variable "SECRET"
-	secret := os.Getenv("SECRET")
-	if secret == "" {
-		secret = "crowdsecret"
+	var (
+		secret []byte
+	)
+
+	secret_string := os.Getenv("SECRET")
```
erenJag, line 150: Can we change this VAR name to something like `CS_LAPI_SECRET` or `CROWDSEC_LAPI_SECRET`, to avoid having same var names if crowdsec is running with other services using this var name.

erenJag (approved):

PR #455, opened 2020-11-20
[wip_lapi] Fix flush alerts

AlteredCoder:
- Use pagination to flush items based on "max_items"
- flush alerts from their creation date and not `started_at` for `max_age`


buixor (approved):
lgtm, cool :+1:

PR #456, opened 2020-11-22
journald support & acquisition revamp

buixor:
- revamp acquisition to prepare for modules : use interface for data sources
 - add support for journald : `acquis.yaml` supports `journalctl_filter` list of arguments passed to `journalctl` 
 - allow `journalctl` to be used as well in forensic mode `crowdsec -jfilter "-u ssh --since yesterday" -type syslog`
 - remove source logging even in debug mode to make it more viable 

```yaml
journalctl_filter:
 - "-u"
 - "mysql"
labels:
  type: syslog
``` 

TBD :
 - [x] File Datasource tests
 - [x] journald tests :grimacing: 
 - how to deal with the case where the log file isn't present at startup but might appear later. currently, the user needs to SIGHUP crowdsec, is it ok?


pkg/acquisition/journalctl_reader.go
```diff
@@ -0,0 +1,174 @@
...
+	Stdout  io.ReadCloser
+	Stderr  io.ReadCloser
+	Decoder *json.Decoder
+	SrcName string
+}
+
+var JOURNALD_CMD = "journalctl"
+var JOURNALD_DEFAULT_TAIL_ARGS = []string{"--follow"}
+var JOURNALD_DEFAULT_CAT_ARGS = []string{}
+
```
registergoofy, line 44: IMHO these shouldn't be variables, but constants, to ensure they are immutable.

pkg/acquisition/journalctl_reader.go
```diff
@@ -0,0 +1,174 @@
...
+			evt := types.Event{Line: l, Process: true, Type: types.LOG, ExpectMode: leaky.LIVE}
+			out <- evt
+		}
+		clog.Debugf("finished reading from journalctl")
+		if err := scanner.Err(); err != nil {
+			clog.Debugf("got an error while reading %s : %s", j.SrcName, err)
+			readErr <- err
+			return
+		}
+		readErr <- nil
```
registergoofy, line 130: IMHO, you should close the chan instead of sending nil.

pkg/acquisition/journalctl_reader.go
```diff
@@ -0,0 +1,174 @@
...
+	}()
+
+	for {
+		select {
+		case <-t.Dying():
+			clog.Debugf("journalctl datasource %s stopping", j.SrcName)
+			return nil
+		case err := <-readErr:
```
registergoofy, line 138:
```
case err, ok := <- readErr:
  if !ok {
    // the readErr chan was closed
    return nil
  }
  clog....
  tKill(err)
```

pkg/acquisition/acquisition.go
```diff
@@ -0,0 +1,154 @@
...
+	Configure(DataSourceCfg) error
+	/*the readers must watch the tomb (especially in tail mode) to know when to shutdown.
+	tomb is as well used to trigger general shutdown when a datasource errors */
+	StartReading(chan types.Event, *tomb.Tomb) error
+	Mode() string //return CAT_MODE or TAIL_MODE
+	//Not sure it makes sense to make those funcs part of the interface.
+	//While 'cat' and 'tail' are the only two modes we see now, other modes might appear
+	//StartTail(chan types.Event, *tomb.Tomb) error
+	//StartCat(chan types.Event, *tomb.Tomb) error
+}
```
registergoofy, line 82: Nice, I love this !

registergoofy (approved):
minor comments

pkg/acquisition/journalctl_reader.go
buixor: yes they should stay as variables: unit tests use those to mock journalctl execution

pkg/acquisition/journalctl_reader.go
buixor: as discussed, skipped for now

PR #457, opened 2020-11-23
have the Mapkey stored in the overflow

registergoofy:
very minimal change to get the bucket mapkey in the overflow.


buixor (approved):
lgtm

PR #459, opened 2020-11-23
Deepcopy

registergoofy:
change deepcopy dep lib

### DIFF
--- a/pkg/parser/runtime.go
+++ b/pkg/parser/runtime.go
@@ -293,12 +293,11 @@ func Parse(ctx UnixParserCtx, xp types.Event, nodes []Node) (types.Event, error)
 			if ret {
 				isStageOK = true
 				if ParseDump {
-					evtcopy := types.Event{}
 					// if err := types.Clone(&event, &evtcopy); err != nil {
 					// 	log.Fatalf("while cloning Event in parser : %s", err)
 					// }
-					evtcopy = deepcopy.Copy(event)
-					StageParseCache[stage][node.Name] = evtcopy
+					evtcopy := deepcopy.Copy(event)
+					StageParseCache[stage][node.Name] = evtcopy.(types.Event)
 				}
 			}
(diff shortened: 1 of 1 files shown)


pkg/parser/runtime.go
```diff
@@ -292,11 +293,11 @@ func Parse(ctx UnixParserCtx, xp types.Event, nodes []Node) (types.Event, error)
...
 				isStageOK = true
 				if ParseDump {
-					evtcopy := types.Event{}
-					if err := types.Clone(&event, &evtcopy); err != nil {
-						log.Fatalf("while cloning Event in parser : %s", err)
-					}
-					StageParseCache[stage][node.Name] = evtcopy
+					// if err := types.Clone(&event, &evtcopy); err != nil {
+					// 	log.Fatalf("while cloning Event in parser : %s", err)
+					// }
```
buixor, line 298: comments

buixor (approved):
dead code, and what does the new lib brings ?

registergoofy:
> dead code, and what does the new lib brings ?

The code is not dead. It's use by and for heavy debug purpose, in hub-tests. This dependency has no impact in crowdsec as ParseDump is always false.

buixor:
> > dead code, and what does the new lib brings ?
> 
> The code is not dead, but It's used for heavy debug purpose only in hub-tests. This dependency has no impact in crowdsec as ParseDump is always false.

No I mean : what does this new lib brings ?
And for dead code I meant the comments sorry

registergoofy:
> > > dead code, and what does the new lib brings ?
> > 
> > 
> > The code is not dead, but It's used for heavy debug purpose only in hub-tests. This dependency has no impact in crowdsec as ParseDump is always false.
> 
> No I mean : what does this new lib brings ?
> And for dead code I meant the comments sorry

The lib we used until now is not able to clone the entire types.Event struct. Deepcopy can. As this as few to no impact on prod, there's no pain in changing this.

PR #461, opened 2020-11-23
fix prometheus startup

AlteredCoder:


cmd/crowdsec/metrics.go
```diff
@@ -78,6 +78,18 @@ func dumpMetrics() {
 }
 
 func registerPrometheus(config *csconfig.PrometheusCfg) {
+	if !config.Enabled {
+		return
+	}
+	if config.ListenAddr == "" {
+		log.Errorf("prometheus is enabled, but the listen address is empty, using '127.0.0.1'")
```
buixor, line 85: make it info I guess

cmd/crowdsec/metrics.go
```diff
@@ -78,6 +78,18 @@ func dumpMetrics() {
...
+		config.ListenAddr = "127.0.0.1"
+	}
+	if config.ListenPort == 0 {
+		log.Errorf("prometheus is enabled, but the listen port is empty, using '9090'")
```
buixor, line 89: make it info I guess + it should be `6060` as it's the port used by `cscli` by default

buixor (changes requested):
I think the default port isn't matching the one used by cscli

buixor (approved):
lgtm

PR #467, opened 2020-11-24
New functional tests

buixor:
- install crowdsec from release (`--bininstall`)
 - start service with systemctl
 - inject logs into another instance in time-machine
 - check (with `cscli decisions list` + filters) expected results
 - delete existing bans
 - add a new file to acquis.yaml
 - restart service
 - inject same logs but in tail mode
 - check results again

### DIFF
--- a/.github/workflows/ci_functests-install.yml
+++ b/.github/workflows/ci_functests-install.yml
@@ -2,9 +2,9 @@ name: Hub-CI
 
 on:
   push:
-    branches: [ wip_lapi_xxx_no ]
+    branches: [ functional_tests ]
   pull_request:
-    branches: [ wip_lapi_xxx_no ]
+    branches: [ functional_tests ]
 
 jobs:
   build:
@@ -25,8 +25,6 @@ jobs:
         cd crowdsec-xxx
         sudo bash -x ./wizard.sh --bininstall
     - name: Post-installation check
-      #capi shouldn't be registered
-      # sudo cscli capi status -> this would error
       run: |
         sudo cscli hub update
     - name: Install collection
--- a/.github/workflows/ci_functests-install.yml
+++ b/.github/workflows/ci_functests-install.yml
@@ -2,9 +2,13 @@ name: Hub-CI
 
 on:
   push:
-    branches: [ functional_tests ]
+    branches:
+      - wip_lapi
+      - master
   pull_request:
-    branches: [ functional_tests ]
+    branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_functests-install.yml
```diff
@@ -2,9 +2,13 @@ name: Hub-CI
...
   push:
-    branches: [ wip_lapi_xxx_no ]
+    branches:
+      - wip_lapi
+      - master
   pull_request:
-    branches: [ wip_lapi_xxx_no ]
+    branches:
+      - wip_lapi
+      - master
```
registergoofy, line 11: This'll have to be changed in the final release.

wizard.sh
```diff
@@ -433,13 +433,6 @@ main() {
         fi
         log_info "installing crowdsec"
         install_crowdsec
-        # lapi register
-        MACHINE_ID=""
-        if [[ ${DOCKER_MODE} == "false" ]]; then
-            ${CSCLI_BIN_INSTALLED} machines add --force "$(cat /etc/machine-id)" --password "$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)" -f "${CROWDSEC_CONFIG_PATH}/${CLIENT_SECRETS}"
-        fi
-            
-        log_info "Crowdsec LAPI registered"
```
registergoofy, line 442: Why get rid of this ?

registergoofy (commented):
first comment is cosmetics
second comment is more a question than a comment

wizard.sh
buixor: The official one is that: `--bininstall` doesn't have to register to LAPI. The less official one is that `$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)` caused the docker image of CI to hang

.github/workflows/ci_functests-install.yml
buixor: I guess we can just remove `wip_lapi` ? or you see something else?

wizard.sh
registergoofy: ok then

.github/workflows/ci_functests-install.yml
registergoofy: fair enough

registergoofy (approved):
lgtm

PR #493, opened 2020-12-01
Improve create alerts input

erenJag:
* Here we check in each alert decisions if `start_ip` & `end_ip` are valid. Before merging that we need to evaluate if it's less efficient or not.

Example result :
```
eren@linux ~/crowdsec (improve_createAlerts_input)$ http POST localhost:8080/v1/alerts 'Authorization: Bearer eyJhbGciOiJI.....' < pkg/apiserver/tests/alert_sample.json 
HTTP/1.1 400 Bad Request
Content-Length: 66
Content-Type: application/json; charset=utf-8
Date: Tue, 01 Dec 2020 11:32:08 GMT

{
    "message": "'start_ip' must be lower than 'end_ip' in a decision"
}
```

### DIFF
--- a/pkg/apiserver/controllers/v1/alerts.go
+++ b/pkg/apiserver/controllers/v1/alerts.go
@@ -121,6 +121,12 @@ func (c *Controller) CreateAlert(gctx *gin.Context) {
 	for _, alert := range input {
 		if len(alert.Decisions) > 0 {
 			log.Debugf("alert %s already has decisions, don't apply profiles", *alert.Message)
+			for _, decision := range alert.Decisions {
+				if decision.EndIP != 0 && decision.StartIP != 0 && decision.EndIP < decision.StartIP {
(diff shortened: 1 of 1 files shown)


pkg/apiserver/controllers/v1/alerts.go
```diff
@@ -121,6 +121,12 @@ func (c *Controller) CreateAlert(gctx *gin.Context) {
 	for _, alert := range input {
 		if len(alert.Decisions) > 0 {
 			log.Debugf("alert %s already has decisions, don't apply profiles", *alert.Message)
+			for _, decision := range alert.Decisions {
+				if decision.EndIP != 0 && decision.StartIP != 0 && decision.EndIP < decision.StartIP {
+					gctx.JSON(http.StatusBadRequest, gin.H{"message": "'start_ip' must be lower than 'end_ip' in decisions"})
```
buixor, line 126: lower **or equal** (if the target is a single IP)

buixor (changes requested):

buixor (approved):
lgtm

PR #504, opened 2020-12-03
Add ci docker push

erenJag:
https://hub.docker.com/r/crowdsecurity/crowdsec


buixor (approved):
can we find a way to only tag latest if it's a release and not a prerelease ?

buixor (approved):
nice

PR #505, opened 2020-12-03
allow to specify username when register to lapi

AlteredCoder:
allow to do: `cscli lapi register --username <username>`


buixor (approved):
lgtm

LtSich:
when you registar with lapi it's for "machines", you should keep that name no ?

cscli lapi register --machines <machine>
Like with cscli machines add <machine>

PR #512, opened 2020-12-06
Enable detection of `httpd`

### LINKED ISSUE #195: Bug/wizard doesn't properly detect apache2 on CentOS 7
**Describe the bug**

Wizard doesn't detect apache2 on centos 7. Probably because the service is called httpd and the logs are in `/var/log/httpd/*.log`

**To Reproduce**
1. Run wizard on centos with apache2

**Expected behavior**
detect apache2 logs

 - OS: centos 7
 - Version: v0.3.0

sbs2001:
Fixes #195 

- [x] Fix `httpd` service discovery on RH based distros. 

- [x] Make installation process for `httpd` work. 

Signed-off-by: Shivam Sandbhor <shivam.sandbhor@gmail.com>

### DIFF
--- a/wizard.sh
+++ b/wizard.sh
@@ -3,7 +3,6 @@
 set -o pipefail
 #set -x
 
-
 RED='\033[0;31m'
 BLUE='\033[0;34m'
 GREEN='\033[0;32m'
@@ -48,6 +47,7 @@ ACTION=""
 DEBUG_MODE="false"
 
 SUPPORTED_SERVICES='apache2
+httpd
 nginx
 sshd
 mysql
@@ -122,6 +122,7 @@ detect_services () {
 
 declare -A log_input_tags
 log_input_tags[apache2]='type: apache2'
+log_input_tags[httpd]='type: httpd'
 log_input_tags[nginx]='type: nginx'
 log_input_tags[sshd]='type: syslog'
 log_input_tags[rsyslog]='type: syslog'
@@ -132,6 +133,7 @@ log_input_tags[linux]="type: syslog"
 
 declare -A log_locations
 log_locations[apache2]='/var/log/apache2/*.log,/var/log/*httpd*.log'
+log_locations[httpd]='/var/log/httpd/*_log'
 log_locations[nginx]='/var/log/nginx/*.log'
 log_locations[sshd]='/var/log/auth.log,/var/log/sshd.log,/var/log/secure'
 log_locations[rsyslog]='/var/log/syslog'
--- a/wizard.sh
+++ b/wizard.sh
@@ -47,6 +47,7 @@ ACTION=""
 DEBUG_MODE="false"
 
 SUPPORTED_SERVICES='apache2
+httpd
 nginx
 sshd
 mysql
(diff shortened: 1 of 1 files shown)


sbs2001:
snapshot1
snapshot2

sbs2001:
Note that, the users won't see "httpd" as shown in images now, rather they will see "apache2" instead, like they normally should.

buixor:
Hello,

Thanks for the PR !
One question tho : does the `systemctl list-unit-files` returns httpd or apache2 or rhel/centos ?
(The idea behind the question is to see if we can easily keep the wizard more generic by avoiding httpd/apache2 specific cases)

sbs2001:
@buixor  It returns httpd

buixor:
I might be tempted to add a new entry httpd in the services, along with apache2. However, I guess it will mess up with the collection detection ?

sbs2001:
@buixor  yes, I initially did it that way (treated them different) but then it messes the collections.

buixor:
ok, then I guess it's worth as it is a transition. Next move should be to move acquisition configuration as being part of collections (so that wizard doesn't have to grow overcomplicated in time).

sbs2001:
@buixor  correct me if I am wrong: 

By "move acquisition configuration as being part of collections"  you mean, to move the acq config which contains config like where to look logs to collections like the ones at https://github.com/crowdsecurity/hub/blob/master/collections/crowdsecurity/apache2.yaml ?

If that's the plan, than how could a user opt to ignore certain detected log files ?

sbs2001:
In the current case, the way creating `httpd` service messes the collections is via https://github.com/crowdsecurity/crowdsec/blob/cbbe210736736fb15ea03b8f7dbe24477485b9e7/wizard.sh#L202  problem being there's no collection named `httpd` .

buixor:
Sorry for the delay, I'm going to test it and merge if it's ok :+1:

wizard.sh
```diff
@@ -90,7 +89,12 @@ detect_services () {
     for SVC in ${SUPPORTED_SERVICES} ; do
         log_info "Checking if service '${SVC}' is running (ps+systemd)"
         for SRC in "${SYSTEMD_SERVICES}" "${PSAX}" ; do
-            echo ${SRC} | grep ${SVC} >/dev/null
+            if [$SRC != "apache2" ] ; then
```
buixor, line 92: I would suggest adding `httpd` to the `SUPPORTED_SERVICES` and substitute it with `apache2` just before we add it to the `DETECTED_SERVICES` array :


in the `if [ $? -eq 0 ];` block before adding `SVC` to the detected service array :

```bash
#on centos, apache2 is named httpd                                                                                                                                                                                            
if [[ ${SVC} == "httpd" ]] ; then
   SVC="apache2";
fi
```


I think it will make the code easier to read.

buixor (changes requested):
suggestion to make code easier to read

wizard.sh
sbs2001: That's more concise 

Done :)

buixor (approved):
great thanks :)

PR #536, opened 2020-12-14
control pushed signals

buixor:
ensure decisions from local or tainted scenarios aren't push, neither are manual decisions


erenJag (approved):
lgtm

PR #537, opened 2020-12-14
export node logger

registergoofy:
Access to this is needed from hub-tests. I think it's legit and it's not worth writing a getter for this. Any thought ?


buixor (approved):
lgtm

PR #543, opened 2020-12-15
up doc

buixor:
- specific doc on crowdsec upgrade
 - improve a bit doc on migration from 0.3 to 1.0
 - link the doc in release drafter


erenJag (approved):
lgtm

PR #557, opened 2021-01-04
jwt token generation improvement

registergoofy:
*add some warning comment for those who want to choose their secret
*strictly follow the golang doc for using crypto/rand
*fatal if not enough entropy

### DIFF
--- a/pkg/apiserver/middlewares/v1/jwt.go
+++ b/pkg/apiserver/middlewares/v1/jwt.go
@@ -150,18 +150,20 @@ func NewJWT(dbClient *database.Client) (*JWT, error) {
 	//Please be aware that brute force HS256 is possible.
 	//PLEASE choose a STRONG secret
 	secret_string := os.Getenv("CS_LAPI_SECRET")
-
 	if secret_string == "" {
 		secret = make([]byte, 64)
 		if n, err := rand.Read(secret); err != nil {
-			log.Fatalf("Unable to generate a new random seed for JWT generation")
+			log.Fatalf("unable to generate a new random seed for JWT generation")
 		} else {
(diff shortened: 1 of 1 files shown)


pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -147,14 +147,16 @@ func NewJWT(dbClient *database.Client) (*JWT, error) {
 		secret []byte
 	)
 
+	//Please be aware that brute force HS256 is possible.
+	//PLEASE choose a STRONG secret
 	secret_string := os.Getenv("CS_LAPI_SECRET")
```
buixor, line 152: maybe check here the length of the secret ?

pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -147,14 +147,16 @@ func NewJWT(dbClient *database.Client) (*JWT, error) {
...
 
 	if secret_string == "" {
-		secret = make([]byte, 8)
-		if n, err := rand.Reader.Read(secret); err != nil {
+		secret = make([]byte, 64)
+		if n, err := rand.Read(secret); err != nil {
```
buixor, line 156: is 64 arbitrarily chosen or is there any kind of best practice we can refer to ?

buixor (changes requested):
maybe check / warn on secret length

pkg/apiserver/middlewares/v1/jwt.go
registergoofy: It's the length used at auth0.

buixor (approved):
lgtm

PR #558, opened 2021-01-05
Fix races

registergoofy:
* proper use of tomb for LeakRoutine
* Fix a bunch of races
* have some locking feature between DumpState and Queue management
* Make sure we empty the overflows when we externally kill LeakRoutine


cmd/crowdsec/crowdsec.go
```diff
@@ -134,30 +159,10 @@ func waitOnTomb() {
...
-						log.Printf("LeakRoutines commit over.")
-						break
-					}
-					successiveStillRounds++
-				}
-				rounds++
-				time.Sleep(5 * time.Second)
-			}
-			return
+			bucketsTomb.Kill(nil)
```
buixor, line 163: We need to ensure that we are not losing overflows being sent by buckets when doing this. Are we sure this can't lead to losing overflows from bucket that have overflowed but are waiting to be able to send their overflow into the appropriate chan ?

buixor (commented):
Tests :
 - no loss of overflows on forensics mode
 - check reload behaviour

cmd/crowdsec/crowdsec.go
registergoofy: * I tried creating 5k+ overflows from special crafted scenarios (trigger + 1 capacity bucket ). We are not loosing any decisions.
* reload seems to happen fine

cmd/crowdsec/crowdsec.go
```diff
@@ -40,46 +41,75 @@ func initCrowdsec() (*parser.Parsers, error) {
...
 
 	//start go-routines for parsing, buckets pour and ouputs.
-	for i := 0; i < cConfig.Crowdsec.ParserRoutinesCount; i++ {
-		parsersTomb.Go(func() error {
-			defer types.CatchPanic("crowdsec/runParse")
-			err := runParse(inputLineChan, inputEventChan, *parsers.Ctx, parsers.Nodes)
-			if err != nil {
-				log.Fatalf("starting parse error : %s", err)
-				return err
+	parserWg := &sync.WaitGroup{}
```
buixor, line 49: comment the reason of the waitgroup as it might not be obvious for future readers

cmd/crowdsec/crowdsec.go
```diff
@@ -90,12 +120,12 @@ func runCrowdsec(parsers *parser.Parsers) error {
...
 }
 
-func serveCrowdsec(parsers *parser.Parsers) {
+func serveCrowdsec(parsers *parser.Parsers, cConfig *csconfig.GlobalConfig) {
 	crowdsecTomb.Go(func() error {
 		defer types.CatchPanic("crowdsec/serveCrowdsec")
 		go func() {
 			defer types.CatchPanic("crowdsec/runCrowdsec")
-			runCrowdsec(parsers)
+			runCrowdsec(cConfig, parsers)
```
buixor, line 128: can we check errors here ?

cmd/crowdsec/crowdsec.go
```diff
@@ -119,9 +149,6 @@ func waitOnTomb() {
 			/*if it's acquisition dying it means that we were in "cat" mode.
 			while shutting down, we need to give time for all buckets to process in flight data*/
 			log.Warningf("Acquisition is finished, shutting down")
-			bucketCount := leaky.LeakyRoutineCount
-			rounds := 0
-			successiveStillRounds := 0
 			/*
 				While it might make sense to want to shut-down parser/buckets/etc. as soon as acquisition is finished,
```
buixor, line 153: maybe kill the comm as it's not accurate anymore. Maybe add a quick explaination about the shutdown mecanism for future readers.

cmd/crowdsec/serve.go
```diff
@@ -18,7 +19,7 @@ import (
 )
 
 //debugHandler is kept as a dev convenience : it shuts down and serialize internal state
-func debugHandler(sig os.Signal) error {
+func debugHandler(sig os.Signal, cConfig *csconfig.GlobalConfig) error {
```
buixor, line 22: linter

cmd/crowdsec/serve.go
```diff
@@ -90,22 +91,20 @@ func reloadHandler(sig os.Signal) error {
...
 		if err != nil {
 			return fmt.Errorf("unable to init crowdsec: %s", err)
 		}
 		//restore bucket state
 		if tmpFile != "" {
-			log.Warningf("Restoring buckets state from %s", tmpFile)
-			if err := leaky.LoadBucketsState(tmpFile, buckets, holders); err != nil {
-				log.Fatalf("unable to restore buckets : %s", err)
-			}
+			log.Warningf("we are now using %s as a state file", tmpFile)
```
buixor, line 100: improve message

buixor (changes requested):
minor changes

buixor:
deprecated vs rebased, we can kill this one ?

PR #564, opened 2021-01-09
Fix ip range operations

buixor:
- db model : decision now has **int64** start_ip, start_suffix, end_ip, end_suffix
    - note that [DBs/ent](https://github.com/lib/pq/issues/72) do not support unsigned int 64 bits. We hack our way around by using int as unsigned int (see uint2int)

- db queries : db queries now properly handle the ipv6 ranges. A boolean `contains`  has been added in most of them (not controllable via API yet), that allows to tell if you want bans *contained* by your ip/range, or bans *containing* your ip/range

-  start_ip and end_ip have been discarded from API
-  killed duplicate ip manipulation code
- regenerated entc code with latest version

TBD :
 - [x] tests for the ipToints functions
 - [x] expose contains flag in query
 - [ ] find a way to avoid duplication of the query building code


scripts/functest.sh
```diff
@@ -0,0 +1,122 @@
...
+docurl ${APIK} "/v1/decisions?range=1.2.3.0/24" | ${JQ} '. == null' || fail
+${CSCLI} decisions list -r 1.2.3.0/24 --contained -o json |${JQ} '.[].decisions[0].value == "1.2.3.4"'  || fail
+docurl ${APIK} "/v1/decisions?range=1.2.3.0/24&contains=false" | ${JQ} '.[0].value == "1.2.3.4"' || fail
+#
+# range ipv4 tests
+#
+#add range decision
+${CSCLI} decisions add -r 4.4.4.0/24 || fail
+${CSCLI} decisions list -o json | ${JQ} '.[0].decisions[0].value == "4.4.4.0/24", .[1].decisions[0].value == "1.2.3.4"' || fail
+docurl ${APIK} "/v1/decisions" | ${JQ} '.[0].value == "4.4.4.0/24", .[1].value == "1.2.3.4"' || fail
```
registergoofy, line 67: is order always maintained for '.[0].value == "4.4.4.0/24", .[1].value == "1.2.3.4"' ?

scripts/functest.sh
```diff
@@ -0,0 +1,122 @@
...
+${CSCLI} decisions list -i aaaa:2222:3333:4445:5555:6666:7777:8888 -o json | jq -e '. == null' || fail
+${CSCLI} decisions list -i aaa1:2222:3333:4444:5555:6666:7777:8887 -o json | jq -e '. == null' || fail
+#check subrange within/out of range
+${CSCLI} decisions list -r aaaa:2222:3333:4444:5555::/80 -o json | jq -e '.[].decisions[0].value == "aaaa:2222:3333:4444::/64"'  || fail
+${CSCLI} decisions list -r aaaa:2222:3333:4441:5555::/80 -o json | jq -e '. == null'  || fail
+${CSCLI} decisions list -r aaa1:2222:3333:4444:5555::/80 -o json | jq -e '. == null'  || fail
+#check outer range
+${CSCLI} decisions list -r aaaa:2222:3333:4444:5555:6666:7777:8888/48 -o json | jq -e '. == null' || fail
+${CSCLI} decisions list -r aaaa:2222:3333:4444:5555:6666:7777:8888/48 -o json --contained | jq -e '.[].decisions[0].value == "aaaa:2222:3333:4444::/64"' || fail
+
```
registergoofy, line 118: may be this one aaa:2222:3333:4445:5555:6666:7777:8888/48 -o json --contained | jq -e '. == null' || fail

scripts/functest.sh
AlteredCoder: It seems that it is maintained

scripts/functest.sh
AlteredCoder: done

PR #568, opened 2021-01-13
Sanitize id from either source

srcr:
On FreeBSD the uuid has dashes. So moved line 74 outside the if loop to have it sanitize either id source of dashes.


AlteredCoder (approved):
Hello @srcr , thanks for the contribution!

This PR looks go to me!

PR #569, opened 2021-01-13
MAKE is now a variable

srcr:
For building on FreeBSD we need gmake instead of make. As first step making MAKE a variable is an improvement.


buixor (approved):
lgtm

PR #570, opened 2021-01-13
PostgreSQL connection with 'sslmode=disable'

srcr:
Most of the time you have PostgreSQL on the same device so using SSL on 127.0.0.1 connection is redundant, still hardcoding this to disabled pretty bad :|


buixor:
For this one, I would suggest making it an option (but it's ok to make it the default).
Let us come back to you on this topic with an alternative proposal!

buixor:
Hello, we merged https://github.com/crowdsecurity/crowdsec/pull/772 that is solving this issue as well :+1:

Thanks again,

PR #591, opened 2021-01-28
add two options: configure and noop

registergoofy:
add two configuration options:
* noop: do nothing
* configure: detect services and install collections


buixor (approved):
lgtm

PR #601, opened 2021-02-04
allow environment variable in configuration file

AlteredCoder:
Note: you need to be `root` when exporting the environment variable or write it in `/etc/environement` since crowdsec run as root.


registergoofy (approved):
lgtm

buixor (changes requested):
lgtm, but we should document it :)

buixor (approved):
lgtm

PR #608, opened 2021-02-05
don't load lapi creds when running only api

### LINKED ISSUE #603: Bug/crowdsec in LAPI-only mode tries to read /etc/crowdsec/local_api_credentials.yaml
**Describe the bug**

When starting crowdsec in LAPI-only mode, it tries to read `local_api_credentials.yaml` which isn't needed.

**To Reproduce**
 - rm `/etc/crowdsec/local_api_credentials.yaml`
 - run `crowdsec  -no-cs`

**Expected behavior**
Not this :

```
# crowdsec  -no-cs
FATA[0000] while loading configuration : failed to load sub configurations: failed to read api client credential configuration file '/etc/crowdsec/local_api_credentials.yaml': open /etc/crowdsec/local_api_credentials.yaml: no such file or directory 

```

 - Version: 1.0.6

AlteredCoder:
fix #603

### DIFF
--- a/cmd/crowdsec-cli/main.go
+++ b/cmd/crowdsec-cli/main.go
@@ -44,9 +44,13 @@ func initConfig() {
 	csConfig = csconfig.NewConfig()
 
 	log.Debugf("Using %s as configuration file", ConfigFilePath)
-	if err := csConfig.LoadConfigurationFile(ConfigFilePath); err != nil {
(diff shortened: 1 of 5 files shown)


buixor (approved):
while I'm not too found of adding two extra params for this, it seems the simplest way :)
lgtm

Issue #609, opened 2021-02-05
Improvement/crowdsec Support X-Forwarded-For/X-Real-IP headers in API


blotus:
**Is your feature request related to a problem? Please describe.**

When running the crowdsec API behind a proxy (for example, in a kubernetes cluster with nginx ingress), the IP displayed by `cscli machines list` is the proxy IP which is meaningless in this context, and not the IP of the actual server running the crowdsec agent (in this example, `10.244.4.17` is the IP of one ingress controller):
```shell
/tmp $ cscli machines list
----------------------------------------------------------------------------------------------------------------------------------------------
 NAME                                              IP ADDRESS   LAST UPDATE           STATUS  VERSION                                         
----------------------------------------------------------------------------------------------------------------------------------------------
 4ff20b29e53a474ebbb9c8b89f8c79766nyqAuxI7k4Q5UJO  127.0.0.1    2021-02-04T10:14:30Z  ✔️       v1.0.4-25562e95759d20c5200ba98d0aac33cc66575ef2 
 4ff20b29e53a474ebbb9c8b89f8c7976id3rtTyQj556aPKj  10.244.4.17  2021-02-05T15:17:10Z  🚫                                                      
----------------------------------------------------------------------------------------------------------------------------------------------
```

**Describe the solution you'd like**

Being able to tell crowdsec to use the X-Forwarded-For and/or the X-Real-IP headers to get the clients IP.

buixor:
thanks, merged :)

PR #610, opened 2021-02-06
Add use_forwarded_for_headers configuration option for LAPI

blotus:
This implements #609.
A new configuration option, `api.server.use_forwarded_for_headers`, is added to allow LAPI to use either the `X-Forwarded-For` or `X-Real-IP` headers to get the true client IP.
This is not enabled by default due to the security concerns if LAPI is not behind a trusted proxy/LB.

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -61,6 +61,12 @@ func NewServer(config *csconfig.LocalApiServerCfg) (*APIServer, error) {
 	}
 	log.Debugf("starting router, logging to %s", logFile)
 	router := gin.New()
+	/* See https://github.com/gin-gonic/gin/pull/2474:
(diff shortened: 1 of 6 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -61,10 +61,7 @@ func NewServer(config *csconfig.LocalApiServerCfg) (*APIServer, error) {
 	}
 	log.Debugf("starting router, logging to %s", logFile)
 	router := gin.New()
-	/*related to https://github.com/gin-gonic/gin/pull/2474
```
buixor, line 64: please keep the ref to gin's issue

buixor (changes requested):
lgtm besides keeping ref to this gin issue so we don't forget why it is default behaviour

buixor (approved):
lgtm

PR #671, opened 2021-03-10
update prometheus default exporter port

### LINKED ISSUE #632: Bug/prometheus default port
**Describe the bug**
It seems that the default port for prometheus scraping is 9090. We should stick with nice defaults. So we should change prometheus default port from 6060 to 9090

he2ss:
Fix #632 
Actually, we should choose default **exporter** port and not the **prometheus server** port. There is a maintained defaults ports list : https://github.com/prometheus/prometheus/wiki/Default-port-allocations. I just added the selected port in this list (CTRL+F 'crowdsec').


buixor (approved):
lgtm

buixor:
postponed

PR #674, opened 2021-03-10
update the config.yaml file

### LINKED ISSUE #45: cscli/crowdsec update available detection
`cscli` / `crowdsec` should be able to detect when an update is available for them

### LINKED ISSUE #658: Bug/cscli branch management
**Describe the bug**

Our version scheme is X.Y.Z. But distribution scheme, can be slightly different, for exemple debian add a number to enable numbering the packaging of the same version.

With cscli hub management this leads to a Warning:
WARN[01-03-2021 06:54:11 PM] Crowdsec is not the latest version. Current version is '1.0.8-1' and latest version is 'v1.0.7'. Please update it!

We should be able to parse and manage it correctly.

AlteredCoder:
- fix cscli: remove the `hub_branch: master` from the configuration file, else cscli will target `master` even for older version (also fix #45)
- fix wizard: now remove also the LAPI log file when uninstalling crowdsec
- fix crowdsec: fix error type and messages returned in pkg/database/machines.go
- fix #658

### DIFF
--- a/.github/workflows/ci_functests-install.yml
+++ b/.github/workflows/ci_functests-install.yml
@@ -32,7 +32,7 @@ jobs:
       run: BUILD_VERSION=${{ steps.keydb.outputs.release }} make release
     - name: Install release
       run: |
-        cd crowdsec-xxx
+        cd crowdsec-${{ steps.keydb.outputs.release }}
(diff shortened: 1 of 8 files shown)


he2ss (approved):
lgtm

PR #676, opened 2021-03-11
pkg/apiclient: pick up dropped errors

alrs:
This picks up dropped error variables in `pkg/apiclient`.

### DIFF
--- a/pkg/apiclient/auth_test.go
+++ b/pkg/apiclient/auth_test.go
@@ -8,6 +8,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestApiAuth(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/apiclient/auth_test.go
```diff
@@ -76,6 +78,7 @@ func TestApiAuth(t *testing.T) {
 	}
 
 	_, resp, err = newcli.Decisions.List(context.Background(), alert)
+	require.NoError(t, err)
```
AlteredCoder, line 81: the CI fail because this test must return an error. As you can see below, we check that the error message contains `APIKey is empty`.

AlteredCoder (changes requested):

AlteredCoder:
Hello @alrs,

Thanks for your contribution. I think that if you fix my comment, the CI will pass :)

alrs:
Should be good, @AlteredCoder, lmk if you see anything else.

AlteredCoder (approved):

PR #684, opened 2021-03-12
fix #677

### LINKED ISSUE #677: Bug/crowdsec: Debug SQL queries
**Describe the bug**
We need to set the crowdsec log level to `trace` if we want to debug the `database` part.

**To Reproduce**
1. Set `debug` in the `db_config` bloc
2. `cat /var/log/crowdsec.log`
3. No SQL queries are printed

**Expected behavior**
SQL queries should be printed in the logs

AlteredCoder:
- crowdsec: `debug` mode for the database part now print SQL queries

Note: this works only when logging on `stdout` :/


buixor (approved):
lgtm

PR #689, opened 2021-03-14
Use --no-cache with apk to skip manual apk update

PeterDaveHello:
Save one step, also prevent additional local cache files need cleaning.


he2ss (approved):
Thanks for this @PeterDaveHello :+1:

PR #692, opened 2021-03-15
Ent update : 0.7.0

buixor:
- update entgo to 0.7.0
 - regenerate schema
 - update documentation (compatibility + schemas)


he2ss (approved):
lgtm

buixor:
> Sorry, forgot to tell you as you suggested it to change [this field type](https://github.com/crowdsecurity/crowdsec/blob/master/pkg/database/ent/schema/event.go#L24).

as seen together, it's not needed. the MaxLen is taken into account (at least on MySQL where it matters [1], but not on SQLite where it seems to have no impact).

[1] 

```
mysql> describe events;
+--------------+---------------+------+-----+---------+----------------+
| Field        | Type          | Null | Key | Default | Extra          |
+--------------+---------------+------+-----+---------+----------------+
| id           | bigint        | NO   | PRI | NULL    | auto_increment |
| created_at   | timestamp     | YES  |     | NULL    |                |
| updated_at   | timestamp     | YES  |     | NULL    |                |
| time         | timestamp     | YES  |     | NULL    |                |
| serialized   | varchar(4095) | NO   |     | NULL    |                |
| alert_events | bigint        | YES  | MUL | NULL    |                |
+--------------+---------------+------+-----+---------+----------------+
6 rows in set (0.00 sec)

```

he2ss (approved):

PR #696, opened 2021-03-18
Unified functional tests

buixor:
- migrate/split/extend functional tests to shell scripts
 - allow re-use of scripts by other packages system (ie `.deb`)
 - kill some dead code in makefile + scripts
 - remove dependency on `/dev/urandom` in the wizard, as a more reliable alternative is provided by `cscli` (cf. [discourse issue](https://discourse.crowdsec.net/t/wizard-stalled-line-594-dev-urandom-tr-dc-a-za-z0-9/122))


.github/workflows/ci_functests-install.yml
```diff
@@ -1,13 +1,11 @@
-name: Functionnal tests install
+name: Functionnal tests
```
registergoofy, line 1: only one "n"

.github/workflows/ci_functests-install.yml
```diff
@@ -33,64 +31,28 @@ jobs:
     - name: Install release
       run: |
         cd crowdsec-${{ steps.keydb.outputs.release }}
-        sudo bash -x ./wizard.sh --bininstall
-        sudo cscli machines add -a
-    - name: Post-installation check
+        sudo bash -x ./wizard.sh --unattended
```
registergoofy, line 34: keep debug ?

registergoofy (approved):
lgtm

PR #700, opened 2021-03-19
pkg/apiserver: fix dropped error

alrs:
This fixes a dropped error and prunes an unused `context.Context` from `pkg/apiserver`.

Even if the `Context` was being used, the Go documentation recommends against that. From `go doc context`:

> Do not store Contexts inside a struct type; instead, pass a Context
explicitly to each function that needs it. The Context should be the first
parameter, typically named ctx:


buixor (approved):
lgtm, thanks for your contribution

PR #708, opened 2021-03-22
fix debian-like installation documentation

registergoofy:
fix debian-like distribution installation documentation by getting rid of add-apt-repository


buixor (approved):
lgtm

PR #712, opened 2021-03-22
[stale] add OpenBSD support for generating the machine id if /etc/machine-id …

rnagy:
…is unavailable

use the hw.uuid sysctl on openbsd to get the hardware UUID in case /etc/machine-id
is not available because /proc is linux-only

### DIFF
--- a/cmd/crowdsec-cli/machines.go
+++ b/cmd/crowdsec-cli/machines.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 	"runtime"
-	"syscall"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/crowdsecurity/crowdsec/pkg/csconfig"
@@ -23,6 +22,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
+	"golang.org/x/sys/unix"
 )
 
 var machineID string
@@ -71,7 +71,7 @@ func generateID() (string, error) {
 		var bID []byte
 		switch runtime.GOOS {
 		case "openbsd":
-			id, err = syscall.Sysctl("hw.uuid")
+			id, err = unix.Sysctl("hw.uuid")
 		default:
 			bID, err = ioutil.ReadFile(uuid)
 			id = string(bID)
--- a/cmd/crowdsec-cli/machines.go
+++ b/cmd/crowdsec-cli/machines.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"strings"
 	"time"
+	"runtime"
+	"syscall"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/crowdsecurity/crowdsec/pkg/csconfig"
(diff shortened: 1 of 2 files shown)


AlteredCoder:
Hello @rnagy , thanks for your contribution!

Can you also push the go.mod please ? The CI failed because of this i think :/

rnagy:
The go.mod I have has no diffs to the one in HEAD.

AlteredCoder:
Do you have a link to the documentation of the `syscall` module that you use ? 
In the official documentation (https://golang.org/pkg/syscall/) , the `Sysctl` function doesn't exist.
What is your go version also please ?

rnagy:
`go version go1.15.2 openbsd/amd64
`

Anyway it seems that syscall is deprecated, so let me get back to you with another solution.

AlteredCoder:
Ok @rnagy , thanks !

cmd/crowdsec-cli/machines.go
```diff
@@ -21,6 +22,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
+	"golang.org/x/sys/unix"
```
cuddytrippin, line 25: T

rnagy: ?

buixor: @rnagy : I reported @cuddytrippin for spam on other issues as well :/

AlteredCoder:
@rnagy it seems that the build still doesn't work :/, let me know if you need any help on this

sbz:
The library used for getting the host uuid [machineid](https://github.com/denisbrodbeck/machineid) does not support OpenBSD yet, I'm suggesting instead of patching crowdsec to provide the OpenBSD support in [machineid](https://github.com/denisbrodbeck/machineid) 

Currently [machineid](https://github.com/denisbrodbeck/machineid#what-you-get) for the 'BSD' use `/etc/hostid` which is valid on FreeBSD and fallback using `kenv` which is not available on OpenBSD. I propose you add for OpenBSD the following:

Read from `/etc/machine-id` if it exist has proposed by [issues#9](https://github.com/denisbrodbeck/machineid/issues/9) and fallback using `sysctl hw.uuid` like the usage of [machine.ID()](https://github.com/crowdsecurity/crowdsec/blob/master/cmd/crowdsec-cli/machines.go#L63) call should be able to handle the OpenBSD case 🤝  

The fallback strategy is implemented [here](https://github.com/denisbrodbeck/machineid/blob/master/id_bsd.go#L18) for the BSD in [machineid](https://github.com/denisbrodbeck/machineid)

rnagy:
> The library used for getting the host uuid [machineid](https://github.com/denisbrodbeck/machineid) does not support OpenBSD yet, I'm suggesting instead of patching crowdsec to provide the OpenBSD support in [machineid](https://github.com/denisbrodbeck/machineid)
> 
> Currently [machineid](https://github.com/denisbrodbeck/machineid#what-you-get) for the 'BSD' use `/etc/hostid` which is valid on FreeBSD and fallback using `kenv` which is not available on OpenBSD. I propose you add for OpenBSD the following:
> 
> Read from `/etc/machine-id` if it exists as proposed by [issues#9](https://github.com/denisbrodbeck/machineid/issues/9) and fallback using `sysctl hw.uuid` like that the usage of [machine.ID()](https://github.com/crowdsecurity/crowdsec/blob/master/cmd/crowdsec-cli/machines.go#L63) call should be able to handle the OpenBSD case 
> 
> The fallback strategy is implemented [here](https://github.com/denisbrodbeck/machineid/blob/master/id_bsd.go#L18) for the BSD in [machineid](https://github.com/denisbrodbeck/machineid)
> 
> For the context, I started to do the same patch with https://gist.github.com/sbz/3c23c52d6817246b3682922944a9c3ba but I realized after it's not needed.

That project seems dead, but it is worth a try... i've submitted a pull request. This ecosystem is really not good.

cuddytrippin (approved):

AlteredCoder:
Hello @rnagy , we will wait to see if they accept your PR. If not, we will integrate those changes in `crowdsec`

AlteredCoder:
Hello @rnagy , your PR in the machineid project is still pending .. If it still the case at the end of the week, we will add your code to crowdsec directly next week.

rnagy:
> Hello @rnagy , your PR in the machineid project is still pending .. If it still the case at the end of the week, we will add your code to crowdsec directly next week.

What if I fork that project and then you can point the build system to use that repo?

PR #715, opened 2021-03-22
fix pattern registration

AlteredCoder:


pkg/parser/node.go
```diff
@@ -407,6 +408,10 @@ func (n *Node) compile(pctx *UnixParserCtx, ectx []EnricherCtx) error {
 	for _, pattern := range n.SubGroks {
 		n.Logger.Tracef("Adding subpattern '%s' : '%s'", pattern.Key, pattern.Value)
 		if err := pctx.Grok.Add(pattern.Key.(string), pattern.Value.(string)); err != nil {
+			if err == grokky.ErrAlreadyExist {
+				n.Logger.Debugf("grok '%s' already registred", pattern.Key)
```
buixor, line 412: should we warn the user that the pattern is being overwritten ? It might be a good help to avoid obscure stuff where the user creates a pattern that is being overwritten by something else ?

buixor (changes requested):
suggestion, but overall lgtm

buixor (approved):
lgtm

PR #718, opened 2021-03-25
FreeBSD changes

sbz:
This change contains the following:

* Update Makefile build definitions for other platforms, we can now build a fresh clone ootb using `gmake -f Makefile`
* Ensure we do not emit `systemctl` messages to the user and simple [do not repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)


sbz:
Rebased on last `master`

PR #721, opened 2021-03-26
FreeBSD changes (#718)

AlteredCoder:
* Update build Makefile to be able to build on other platforms

Split Makefiles to isolate every specifics per platform if specific file
is not found, the include will default to linux.

* Isolate linux specific messages emitted by crowdsec command line

On other platforms, we shouldn't write messages with `systemctl`
occurrences. This commit isolate the specific messages and ensure the
messages are relevant for the given platform according result of `GOOS`.


he2ss (approved):
lgtm

Issue #739, opened 2021-04-07
Bug/Don't block alerts with empty timestamps


buixor:
cf. https://github.com/crowdsecurity/crowdsec/issues/736

**Describe the bug**
Alerts with empty timestamp trigger errors when sent to LAPI

**To Reproduce**
Have a parser/scenario that triggers overflow without capturing any timestamp

**Expected behavior**
Default time to `now()`

Issue #736, opened 2021-04-06
Bug/ unable to create alerts: event timestamp '' : parsing time \"\" as \"2006-01-02T15:04:05Z07:00\"


FidoUser:
Please, start your issue name (after `Bug`) with the component name impacted by this feature request and a small description of the Bug. Example: `Bug/cscli: issue with ....` and remove this line :)

**Describe the bug**
My custom decisions periodically can not work as expected
In log
```
time="06-04-2021 17:44:18" level=info msg="(3d61908636144fe39e836ccb745a58adsve5wWrxFJQ4udV6/crowdsec) crowdsecurity/asterisk-bf by ip 10.9.78.97 : 4h ban on Ip 10.9.78.97"
time="06-04-2021 17:44:18" level=error msg="while pushing to api : failed sending alert to LAPI: API error: unable to create alerts: event timestamp '' : parsing time \"\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"2006\": unable to parse time"
```
These events add to logfile until restart crowdsec
 

**To Reproduce**
Application log event like this
`[2021-04-06 17:17:49] NOTICE[911] chan_sip.c: Registration from '<sip:12313@10.9.99.9>' failed for '10.9.78.97:51076' - Wrong password`

Parser:

```
onsuccess: next_stage
filter: "evt.Parsed.program == 'asterisk'"
name: crowdsecurity/asterisk-logs
description: "Parse Asterisk full logs"
nodes:
  - grok:
      pattern: Registration from '<%{WORD:protocol}:%{WORD:login}@%{IP:ast_IP}>' failed for '%{IP:source_ip}:[0-9]+' - (Wrong password|Username/auth name mismatch)
      apply_on: message
      statics:
        - meta: target_user
          expression: "evt.Parsed.login"
        - meta: source_ip
          expression: "evt.Parsed.source_ip"

statics:
    - meta: log_type
      value: asterisk_failed-auth
```

**Expected behavior**
timestamp can use current  time and decisions will add IP

**Screenshots**

**Technical Information (please complete the following information):**
 - OS: Ubuntu 20.04.2 LTS (GNU/Linux 5.4.0-65-generic x86_64)
 - Version v1.0.9-4-debian

FidoUser:
Issue reproduce with this **scenario**

```
type: leaky

name: crowdsecurity/asterisk-bf
description: "Detect asterisk bruteforce"
[...]

PR #756, opened 2021-04-16
add TimeNow in the exprlib helpers

registergoofy:
* add an helper for getting the current time. Fixes https://github.com/crowdsecurity/crowdsec/issues/736.
* Improve ParseDate by not letting unknown date getting through. time.Now() is used when `evt.TimeStr` is not parseable. should fix https://github.com/crowdsecurity/crowdsec/issues/739

### DIFF
--- a/pkg/parser/enrich.go
+++ b/pkg/parser/enrich.go
@@ -50,25 +50,26 @@ func Loadplugin(path string) ([]EnricherCtx, error) {
 }
 
 func GenDateParse(date string) (string, time.Time) {
-	var retstr string
-	var layouts = [...]string{
-		time.RFC3339,
-		"02/Jan/2006:15:04:05 -0700",
-		"Mon Jan 2 15:04:05 2006",
-		"02-Jan-2006 15:04:05 europe/paris",
-		"01/02/2006 15:04:05",
-		"2006-01-02 15:04:05.999999999 -0700 MST",
-		//Jan  5 06:25:11
-		"Jan  2 15:04:05",
-		"Mon Jan 02 15:04:05.000000 2006",
-		"2006-01-02T15:04:05Z07:00",
-		"2006/01/02",
-		"2006/01/02 15:04",
-		"2006-01-02",
-		"2006-01-02 15:04",
-		"2006/01/02 15:04:05",
-		"2006-01-02 15:04:05",
-	}
+	var (
+		layouts = [...]string{
+			time.RFC3339,
+			"02/Jan/2006:15:04:05 -0700",
+			"Mon Jan 2 15:04:05 2006",
+			"02-Jan-2006 15:04:05 europe/paris",
+			"01/02/2006 15:04:05",
+			"2006-01-02 15:04:05.999999999 -0700 MST",
+			//Jan  5 06:25:11
+			"Jan  2 15:04:05",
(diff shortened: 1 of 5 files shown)


he2ss (approved):
lgtm

pkg/parser/enrich.go
```diff
@@ -85,17 +86,27 @@ func GenDateParse(date string) (string, time.Time) {
...
+		return "", time.Time{}
+	}
+	return string(retstr), now
 }
 
 func ParseDate(in string, p *types.Event, x interface{}) (map[string]string, error) {
 
 	var ret map[string]string = make(map[string]string)
-
+	log.Printf("Prout")
```
he2ss, line 102: remove useless debug

pkg/parser/enrich.go
```diff
@@ -85,17 +86,27 @@ func GenDateParse(date string) (string, time.Time) {
...
 	tstr, tbin := GenDateParse(in)
 	if !tbin.IsZero() {
+		log.Printf("Here")
```
he2ss, line 105: same here

pkg/parser/enrich.go
```diff
@@ -85,17 +86,27 @@ func GenDateParse(date string) (string, time.Time) {
...
 		ret["MarshaledTime"] = string(tstr)
 		return ret, nil
 	}
+	log.Printf("not Here")
```
he2ss, line 109: same

pkg/parser/enrich.go
```diff
@@ -85,17 +86,27 @@ func GenDateParse(date string) (string, time.Time) {
 			return string(retstr), t
 		}
 	}
-	return retstr, time.Time{}
+
+	now := time.Now()
```
he2ss, line 90: You should check if the param date is empty or not. Because the retstr and the time returned are both overwritten.

pkg/parser/runtime.go
```diff
@@ -133,8 +133,11 @@ func (n *Node) ProcessStatics(statics []types.ExtraField, event *types.Event) er
 		}
 
 		if value == "" {
-			clog.Debugf("Empty value for %s, skip.", printStaticTarget(static))
-			continue
+			//allow ParseDate to have empty input
+			if static.Method != "ParseDate" {
```
he2ss, line 137: Don't think if this is necessary because you don't ignore others statics with potential empty value (it was the case before). Just check with the team if there no weird behavior with that.

he2ss (changes requested):
change comments before please

pkg/parser/runtime.go
registergoofy: My tests showed that this exception is mandatory. Node process just skip empty field...

pkg/parser/enrich.go
registergoofy: I see nothing overwritten here. If the param date is parsed, then it has already returned directly from the for loop. So if we are in this code, then date parsing failed.

btw I removed the retstr var because it wasn't scoped correctly.

he2ss: my bad, read it too fast.

pkg/parser/runtime.go
he2ss: the second time I misunderstand the logic sorry. Forget it.

he2ss (approved):

PR #760, opened 2021-04-19
Update CI

blotus:
- Force machine-id when running tests
 - Use the source branch instead of master when dispatching to packaging repo


buixor (approved):
lgtm

PR #762, opened 2021-04-23
cscli: sort meta by key in alerts inspect output

blotus:
Sort the metas by key name when running `cscli alerts inspect -d XX`.


registergoofy (approved):
lgtm

PR #763, opened 2021-04-23
add System to cwversion to have platform in UA

buixor:
- cwversion.System is defined by the makefile and should be like 'linux' or 'docker' etc.


registergoofy (approved):
lgtm

PR #768, opened 2021-04-27
fix test on prerelease

registergoofy:
make the functional tests wait for the packages to be fully built and uploaded


buixor (approved):
lgtm

PR #772, opened 2021-05-03
Add postgresql sslmode option

svesve:
Add postgresql sslmode option, fix spaces in docs

### DIFF
--- a/docs/v1.X/docs/references/crowdsec-config.md
+++ b/docs/v1.X/docs/references/crowdsec-config.md
@@ -37,7 +37,7 @@ db_config:
   db_path: /var/lib/crowdsec/data/crowdsec.db
   #user:
   #password:
-  #db_name: 
+  #db_name:
   #host:
   #port:
   flush:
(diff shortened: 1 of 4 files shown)


buixor:
Hello,

Thanks for the merge request, let's have a look at this :+1:

svesve:
Hello! Have u look this merge? PG sslmode option is very useful, because by default postgresql works wthout ssl

svesve:
@buixor yes i've try ssl + pgsql and this is a big trouble for quick start to use crowdsec + pg batabase
1. You need to generate selfsigned ssl certificate, bcs 99% postgressql instances use ip instead of FQDN
2. In most case, users have already worked postgresql instances without ssl
So, I think sslmode option is important part of this project.

buixor:
Thanks for the suggestion, and I agree it's important for the project. However if I'm not mistaken your patch doesn't set a default value, so it will break existing installs with postgresql no ?

svesve:
It can't brake existing install, you must define sslmode in config file, like a host, port and etc..

buixor:
> It can't brake existing install, you must define sslmode in config file, like a host, port and etc..

Yes, I was actually afraid that someone upgrading without defining `sslmode` will end up with invalid connection string and break install. But I tried locally and it seems good, so we're good to merge.

Thanks !

buixor:
Sorry for the delay, and thanks !

Issue #781, opened 2021-05-06
Bug/error with `cscli alerts delete`


buixor:
In some odd case, I got this error when trying to use `cscli alerts delete --all` :

```
TRAC[05-05-2021 05:34:29 PM] req-jwt: DELETE /v1/alerts?limit=0 HTTP/1.1
Host: localhost:8080
Authorization: Bearer ...
User-Agent: crowdsec/v1.0.13-3-...
TRAC[05-05-2021 05:36:19 PM] resp-jwt: HTTP/1.1 500 Internal Server Error
Content-Length: 78
Content-Type: application/json; charset=utf-8
Date: Wed, 05 May 2021 15:36:19 GMT
{"message":"event with alert ID '1195084': unable to delete"}{"nbDeleted":"0"} (err:<nil>) 
DEBU[05-05-2021 05:36:19 PM] resp-jwt: 500                                
FATA[05-05-2021 05:36:19 PM] Unable to delete alerts : http code 500, invalid body: invalid character '{' after top-level value
```

**Technical Information (please complete the following information):**
 - Version v1.0.13

buixor:
somehow, this bug seems to happen when https://github.com/crowdsecurity/crowdsec/issues/778 happens

PR #794, opened 2021-05-12
delete orphan nodes (fix #778)

### LINKED ISSUE #778: Bug/lapi: some `events` has the `alert_events` to NULL
**Describe the bug**
For some users, we have the `alert_events` and the `alert_decisions` set to NULL in DB.
This lead to crowdsec not flushing them and the database size growing.

**To Reproduce**
Still didn't manage to reproduce

 - crowdsec v1.0.13

I think we should at least remove all `alert_events` and the `alert_decisions` that are set to NULL when flushing the database.

buixor:
- `FlushAlerts` routine now deletes orphan events nodes
 - Should limit as well the bug as seen in #781

### DIFF
--- a/pkg/database/alerts.go
+++ b/pkg/database/alerts.go
@@ -575,11 +575,27 @@ func (c *Client) DeleteAlertWithFilter(filter map[string][]string) (int, error) 
 	return len(alertsToDelete), nil
 }
 
+func (c *Client) FlushOrphans() {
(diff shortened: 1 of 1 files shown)


AlteredCoder:
Nice !!! 
Can you please add this flush for decisions also ?

buixor:
@AlteredCoder : done on the decisions as well

AlteredCoder (approved):

PR #803, opened 2021-05-25
add docker option to specify local timezone

flo-mic:
This pull request allows to specify the timezone to use for the container, even if your operating system does not have the file `/etc/timezone` available. If you have for example a Synology NAS as docker host you do not have this file and can not send a proper timezone to the container.

The package `tzdate` allows to pass a timezone as docker argument to the container. This makes the console logs more readable as they print the timestamp as local timestamp. A list of the available time zones is available here and also linked in the related `docker/Readme.md` file: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

An example docker file which shows the usage with Europe/Berlin timezone looks like this:

```
version: "3.7"
services:
  crowdsec:
    image: crowdsecurity/crowdsec
    environment:
      - TZ=Europe/Berlin
      - COLLECTIONS=crowdsecurity/nginx
      - REGISTER_TO_ONLINE_API=true
    restart: unless-stopped
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN /bin/bash wizard.sh --docker-mode
 RUN cscli hub update && cscli collections install crowdsecurity/linux
 
 FROM alpine:latest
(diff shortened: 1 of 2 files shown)


blotus (approved):

blotus:
Thanks for your contribution !

PR #829, opened 2021-06-03
Build docker image for both amd64 and arm64

blotus:
As we already provide packages for amd64 and arm64, we should also provide a native arm64 docker image.


he2ss (approved):
lgtm

PR #833, opened 2021-06-23
Fix unset environment variable

webkp:
According to your instructions for running CrowdSec with Docker, the following environment variable must be set to process a single file: ```-e FILE_PATH="<file_path>"``` 
(see https://github.com/crowdsecurity/crowdsec/blob/master/docker/README.md#environment-variables)

But in the script docker_start.sh the environment variable FILE is expected, which is not described anywhere:
https://github.com/crowdsecurity/crowdsec/blob/b29730520f2119bd74f365beb603a5bd6d1b01ed/docker/docker_start.sh#L48-L49

I suggest to use the already set ```$FILE_PATH```

### DIFF
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -46,7 +46,7 @@ if [ "$CONFIG_FILE" != "" ]; then
     ARGS="-c $CONFIG_FILE"
 fi
 if [ "$FILE_PATH" != "" ]; then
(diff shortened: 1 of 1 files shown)


buixor:
Hello,

With the acquisition revamp, this has been changed a bit, and this PR is not needed anymore.
Thanks for the suggestion nontheless !

PR #844, opened 2021-07-01
fix #840 : check for whiptail in detect mode

### LINKED ISSUE #840: Bug/wizard.sh
wizard fail with some weird error because of missing whiptail
**Describe the bug**

**To Reproduce**
Try to use wizard.sh without whiptail.

Whiptail presence should be checked when wizard is launched with SILENT=false (whiptail is not needed when SILENT is true)

buixor:


sabban (approved):
lgtm

Issue #837, opened 2021-06-28
Bug/Podman compatibility for metabase


blotus:
From gitter:

`If I add a compatibility symlink /run/docker.sock, I get to this point:`
```
~# cscli dashboard setup 
INFO[28-06-2021 01:48:12 PM] /var/lib/crowdsec/data/metabase.db exists, skip. 
INFO[28-06-2021 01:48:12 PM] Pulling docker image metabase/metabase:v0.37.0.2 
...........
INFO[28-06-2021 01:48:14 PM] creating container '/crowdsec-metabase'      
FATA[28-06-2021 01:48:14 PM] container create: failed to create container : Error response from daemon: container create: error running container create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
```

Looks like podman does not accept a `/` in the container name, and it is currently hardcoded, making the dashboard impossible to run with podman.

buixor:
fixed in upcoming release, keeping opened until published :)

PR #845, opened 2021-07-01
fix #842 #837

### LINKED ISSUE #842: Bug/metabase
**Describe the bug**

when doing 
``` 
sudo cscli dashboard setup --listen 0.0.0.0
? For metabase docker to be able to access SQLite file we need to add a new group called 'crowdsec' to the system, is it ok for you ? Yes
INFO[01-07-2021 10:12:53 AM] Pulling docker image metabase/metabase:v0.37.0.2 
FATA[01-07-2021 10:12:53 AM] container create: failed to pull docker image : Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? 
```
**To Reproduce**
1. trying to install dashboard without docker

**Expected behavior**
The whole metabase stuff is bound to fail, but a crowdsec group is added. The system shouldn't be modified, if the stuff is bound to fail.

What I suggest is to do things otherwise, begin by checking that docker is available, and then modify the system to fit in.

buixor:
This PR fixes #842 and #837

### DIFF
--- a/cmd/crowdsec-cli/dashboard.go
+++ b/cmd/crowdsec-cli/dashboard.go
@@ -53,13 +53,12 @@ cscli dashboard stop
 cscli dashboard remove
 `,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			if err := metabase.TestAvailability(); err != nil {
+				log.Fatalf("%s", err)
+			}
 
 			if err := csConfig.LoadAPIServer(); err != nil || csConfig.DisableAPI {
 				log.Fatal("Local API is disabled, please run this command on the local API machine")
-			}
-
-			if err := metabase.TestAvailability(); err != nil {
-				log.Fatalf("%s", err)
 			}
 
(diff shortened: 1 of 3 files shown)


cmd/crowdsec-cli/dashboard.go
```diff
@@ -72,7 +67,9 @@ cscli dashboard remove
 				log.Errorf("This command requires direct database access (must be run on the local API machine)")
 				log.Fatalf(err.Error())
 			}
-
+			if err := metabase.TestAvailability(); err != nil {
```
blotus, line 70: Move this check in first.
Absence of docker / arch != amd64 are 'bigger' issues (or at least harder to fix) than having LAPI disabled.

cmd/crowdsec-cli/dashboard.go
```diff
@@ -73,6 +71,12 @@ cscli dashboard remove
 				log.Fatalf(err.Error())
 			}
 
+			if !metabase.IsContainerExist(metabaseContainerID) {
```
buixor, line 74: please a tiny comz here, we're going to forget about it very quickly :p

AlteredCoder (approved):

PR #855, opened 2021-07-07
Remove non POSIX sed usage

sbz:
`sed -r` is not compatible with POSIX. Replace it by `sed -E` which is
supported by either GNU sed or BSD sed


buixor (approved):
lgtm

PR #862, opened 2021-07-16
allow to override GOARCH and GOOS when building with the Makefile

blotus:
We default to linux/amd64 if GOARCH and GOOS are not provided in the environment.


buixor (approved):
lgtm

PR #871, opened 2021-07-28
Improve CAPI pull management

### LINKED ISSUE #870: Bug/client database logging doesn't respect log media
When setting database->log_level, the logging doesn't respect the "main" log media, and logs to stderr no matter what.

buixor:
- improve capi mgmt of community-blocklist to prepare for new consensus
 - fix #870 : ent logging now respects logger media directive
 - ensure we don't pull from capi more often than every 2 hours (with the incoming increase in distributed IP amounts, we don't want it to happen every reload)

### DIFF
--- a/pkg/apiserver/apic.go
+++ b/pkg/apiserver/apic.go
@@ -301,7 +301,7 @@ func (a *apic) PullTop() error {
 		capiPullTopX.Decisions = append(capiPullTopX.Decisions, decision)
 	}
 
-	_, err = a.dbClient.CreateAlertBulk(CapiMachineID, []*models.Alert{&capiPullTopX})
+	_, err = a.dbClient.CreateSingleAlertWithBulk(CapiMachineID, &capiPullTopX)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/apiserver/apic.go
```diff
@@ -283,37 +301,30 @@ func (a *apic) PullTop() error {
...
-			return errors.Wrap(err, "decision creation from crowdsec-api:")
-		}
+	log.Printf("capi/community-blocklist : added %d entries", len(data.New))
+
+	intID, err := strconv.Atoi(alertID)
+	if err != nil {
+		return errors.Wrap(err, "while converting alert ID")
+	}
+	/*Deleting older decisions from capi*/
+	decisions_count, err := a.dbClient.Ent.Decision.Delete().
```
he2ss, line 319: Maybe add this in the database pkg as a function to centralize all db operations.

AlteredCoder (approved):

sabban (commented):
I left one comment on code structure, not that important though.

Issue #869, opened 2021-07-26
Improvement/LAPI : Add healthcheck endpoint 


he2ss:
Add healthcheck endpoint `/health` to the LAPI that return 200 if the service is ok. 
It's standardized way to validate the status of a service.

buixor:
https://github.com/alexliesenfeld/health

nanikjava:
@buixor Interested to work on this. 

So pretty much what we need is `lapiserver/health` endpoint right ?

borisjacquot:
@nanikjava  Yes that's the idea. It has to return `200` only if the LAPI is alive. A use case could be with HAProxy, with something lilke that :
```
   option httpchk GET /health
   http-check expect status 200
```
(https://www.haproxy.com/blog/using-haproxy-as-an-api-gateway-part-3-health-checks/)

That's better than a TCP check ^^

nanikjava:
@borisjacquot thanks

@buixor will take a look https://github.com/alexliesenfeld/health and get WIP PR up.

PR #881, opened 2021-08-05
fix: add /health endpoint

nanikjava:
PR Fix : #869 

Add `/health` endpoint for health check. 

The output will be

```
{"status":"up"}
```

To test with `curl`

`curl localhost:8081/health`

### DIFF
--- a/pkg/apiserver/controllers/controller.go
+++ b/pkg/apiserver/controllers/controller.go
@@ -2,7 +2,9 @@ package controllers
 
 import (
 	"context"
+	"github.com/alexliesenfeld/health"
 	"net/http"
+	"time"
 
 	v1 "github.com/crowdsecurity/crowdsec/pkg/apiserver/controllers/v1"
 	"github.com/crowdsecurity/crowdsec/pkg/csconfig"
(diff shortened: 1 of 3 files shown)


nanikjava:
@buixor WIP PR is up. Let me know what you think about this, still rough want to show the idea first.

pkg/apiserver/controllers/controller.go
```diff
@@ -37,12 +39,22 @@ func (c *Controller) Init() error {
...
 }
 
+// endpoint for health checking
+func serveHealth() http.HandlerFunc {
+	checker := health.NewChecker(
+		health.WithCacheDuration(1*time.Second),
+		health.WithDisabledDetails(),
+	)
+	return health.NewHandler(checker)
+}q
```
buixor, line 49: I think there is unwanted `q` here

buixor:
Hello ! I think it looks like a simple but sufficient approach :+1: 
I'm off next week, but we can think about if further checks are needed or if it might be enough for a first iteration !

pkg/apiserver/controllers/controller.go
nanikjava: Removed

nanikjava:
> Hello ! I think it looks like a simple but sufficient approach 
> I'm off next week, but we can think about if further checks are needed or if it might be enough for a first iteration !

@buixor thanks...PR updated with bit of clean up

buixor (approved):
lgtm

buixor:
thanks, looks good.
let's wait for a second review and merge, should make its way in the next release :+1:

sabban (approved):
lgtm

Just one thought. This adds a dependency, I assume it's worthy.

PR #882, opened 2021-08-06
fix plugin process leakage and config discovery.

sbs2001:


pkg/csplugin/broker.go
```diff
@@ -146,6 +146,10 @@ func (pb *PluginBroker) loadConfig(path string) error {
 		return err
 	}
 	for _, configFilePath := range files {
+		if !strings.HasSuffix(configFilePath, "yaml") && !strings.HasSuffix(configFilePath, "yml") {
```
buixor, line 149: I think we want `.` as well

pkg/csplugin/broker.go
```diff
@@ -156,6 +160,9 @@ func (pb *PluginBroker) loadConfig(path string) error {
 				continue
 			}
 			setRequiredFields(&pluginConfig)
+			if _, ok := pb.pluginConfigByName[pluginConfig.Name]; ok {
+				log.Warnf("plugin config %s is repeated", pluginConfig.Name)
```
buixor, line 164: I think we want a more explicit message

buixor (changes requested):

buixor (approved):

PR #884, opened 2021-08-11
set hubBranch to master if not provided in the configuration and if c…

blotus:
…rowdsec has been built without version information


he2ss (approved):
lgtm

PR #894, opened 2021-08-17
fix #890 : log info if profile is in debug, independently of the result

### LINKED ISSUE #890: Bug/ the debugging for the profiles expressions only runs if the result is false
**Describe the bug**
The debugging for the profiles expressions only runs if the result is false.

**To Reproduce**
Have a profile that will match and enable debug in it. no debug output will be produced.

**Expected behavior**
debug should be present.

buixor:
- fix #890 : log info if profile is in debug, independently of the result


sabban (approved):
gtm

PR #895, opened 2021-08-17
Download datafile

sabban:
download datafile when cscli hub upgrade if any datafile is missing.

### DIFF
--- a/cmd/crowdsec-cli/utils.go
+++ b/cmd/crowdsec-cli/utils.go
@@ -223,7 +223,9 @@ func UpgradeConfig(itemType string, name string, force bool) {
 		found = true
 		if v.UpToDate {
 			log.Infof("%s : up-to-date", v.Name)
+
 			if !force {
+				cwhub.DownloadDataIfNeeded(csConfig.Hub, v)
 				continue
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/cwhub/download.go
```diff
@@ -223,3 +225,44 @@ func DownloadItem(hub *csconfig.Hub, target Item, overwrite bool) (Item, error)
...
+	dec := yaml.NewDecoder(itemFile)
+
+	for {
+		data := &types.DataSet{}
+		err = dec.Decode(data)
+		if err != nil {
+			if err == io.EOF {
+				break
+			} else {
+				return errors.Wrap(err, fmt.Sprintf("while reading file %s", itemFilePath))
```
buixor, line 250: wrapf instead of wrap + sprintf

cmd/crowdsec-cli/utils.go
```diff
@@ -223,7 +223,11 @@ func UpgradeConfig(itemType string, name string, force bool) {
 		found = true
 		if v.UpToDate {
 			log.Infof("%s : up-to-date", v.Name)
+
 			if !force {
```
buixor, line 227: if force is enabled, it won't download the data files again ? not sure it's what we want

sabban: it will be download elsewhere anyway. It's to avoid download it twice.

buixor (approved):
lgtm

PR #900, opened 2021-08-19
Minor changes to specific logs

ThinkChaos:
- 7794599 replaces "crowdsec api" with more precise CAPI/LAPI where possible. I might've gotten some wrong, please check
 - 1834e84 removes an extra slash present in log output
 - ea7a88c removes a spammy log


buixor (approved):
lgtm

buixor:
thanks a lot @ThinkChaos :+1:

sabban (approved):
lgtm

blotus:
Hello @ThinkChaos,

It looks like the change for removing the extra `/` is broken.
When running locally, i get the following error (note the missing `/` after patterns):
```
ERRO[20-08-2021 10:20:30] failed to load pattern aws : open /home/seb/taff/crowdsec/git/crowdsec-org/crowdsec/crowdsec-v1.1.1/tests/config/patternsaws: no such file or directory 
FATA[20-08-2021 10:20:30] crowdsec init: Failed to load parsers: failed to load parser patterns : open /home/seb/taff/crowdsec/git/crowdsec-org/crowdsec/crowdsec-v1.1.1/tests/config/patternsaws: no such file or directory 
FATA[20-08-2021 10:20:30] crowdsec init: Failed to load parsers: failed to load parser patterns : open /home/seb/taff/crowdsec/git/crowdsec-org/crowdsec/crowdsec-v1.1.1/tests/config/patternsaws: no such file or directory 
```

ThinkChaos:
Thanks @blotus , I have to admit I didn't test this (I forgot `path.Join` cleans up trailing slashes). Should be fixed now.

I also refactored the `Init` function to use proper args instead of a `map[string]interface{}`. I didn't see anywhere in the code where that was used, but as it's a public function maybe there's something I'm missing. If that's the case I'll revert the commit.

buixor:
@ThinkChaos : hello can you revert the last change on `Init` ? It's breaking the unit tests and is (I guess) used as well by the CI of the hub

ThinkChaos:
Done: forced push to the previous commit.

buixor (approved):

buixor:
thanks, merging :)

pkg/apiserver/controllers/v1/alerts.go
```diff
@@ -135,9 +135,9 @@ func (c *Controller) CreateAlert(gctx *gin.Context) {
 	}
 	select {
 	case c.CAPIChan <- input:
-		log.Debugf("alert send to CAPI channel")
+		log.Debug("alert send to CAPI channel")
 	default:
-		log.Warningf("Cannot send alert to Central API channel")
+		log.Debug("Cannot send alert to Central API channel")
```
AlteredCoder, line 140: Hello!

Sorry for the late review, can you let this line in `Warningf` please?

AlteredCoder (changes requested):

pkg/apiserver/controllers/v1/alerts.go
ThinkChaos: That was the main reason for me forking Crowdsec: every single decision on my server triggers this log.
IMO the warning at start/reload saying the CAPI is disabled should be enough. Why do you want this to be a warning specifically?

AlteredCoder: If you disable CAPI, indeed the local API shouldn't try to send the alert to the CAPI channel. I've opened an [issue](https://github.com/crowdsecurity/crowdsec/issues/907) and this will be fix in the next release.

When this will be fixed, if you see this log again while having CAPI enabled, this mean that there is a problem somewhere.

ThinkChaos: Can you point me in the correct direction so I do the change?

ThinkChaos: I force pushed a change that makes the log a warning again, but only tries to send the event if the channel isn't `nil`. Please tell me if you think the check should be done differently.

AlteredCoder: Yes i think it should be good, thanks!

AlteredCoder (approved):

PR #901, opened 2021-08-20
Allow plugins to set log level

sbs2001:
Signed-off-by: Shivam Sandbhor <shivam@crowdsec.net>

### DIFF
--- a/plugins/notifications/slack/slack_plugin.go
+++ b/plugins/notifications/slack/slack_plugin.go
@@ -13,11 +13,12 @@ import (
 )
 
 type PluginConfig struct {
-	Name    string `yaml:"name"`
-	Webhook string `yaml:"webhook"`
+	Name     string  `yaml:"name"`
+	Webhook  string  `yaml:"webhook"`
+	LogLevel *string `yaml:"log_level"`
 }
 type Notify struct {
-	WebhooksByConfigName map[string]string
+	ConfigByName map[string]PluginConfig
 }
 
 var logger hclog.Logger = hclog.New(&hclog.LoggerOptions{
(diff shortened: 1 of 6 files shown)


plugins/notifications/slack/slack_plugin.go
```diff
@@ -28,8 +29,19 @@ var logger hclog.Logger = hclog.New(&hclog.LoggerOptions{
 })
 
 func (n *Notify) Notify(ctx context.Context, notification *Notification) (*Empty, error) {
+	if _, ok := n.ConfigByName[notification.Name]; !ok {
+		return nil, fmt.Errorf("invalid plugin config name %s", notification.Name)
+	}
+	cfg := n.ConfigByName[notification.Name]
+	if cfg.LogLevel != nil {
```
buixor, line 36: I think we want to have the same check in both plugins, so maybe check here for non-empty string ?

plugins/notifications/splunk/main.go
```diff
@@ -39,12 +40,17 @@ type Payload struct {
 }
 
 func (s *Splunk) Notify(ctx context.Context, notification *Notification) (*Empty, error) {
-	logger.Info(fmt.Sprintf("received notify signal for %s config", notification.Name))
 	if _, ok := s.PluginConfigByName[notification.Name]; !ok {
 		return &Empty{}, fmt.Errorf("splunk invalid config name %s", notification.Name)
 	}
-
 	cfg := s.PluginConfigByName[notification.Name]
+	if cfg.LogLevel != nil {
```
buixor, line 47: same :)

buixor (changes requested):

buixor (approved):

PR #917, opened 2021-08-30
add suport for --since in journalctl DSN

### LINKED ISSUE #916: Bug/Regression: --since support for journalctl acquisition
When acquiring logs from journalctl in one shot mode, we used to support the `--since` flag pre v1.1, but it is not working since the acquisition rewrite.

blotus:
Fix #916


buixor (approved):
lgtm

PR #942, opened 2021-09-07
fix plugins directories

sabban:
fix plugins directories


wizard.sh
```diff
@@ -23,7 +23,7 @@ CROWDSEC_PATH="/etc/crowdsec"
 CROWDSEC_CONFIG_PATH="${CROWDSEC_PATH}"
 CROWDSEC_LOG_FILE="/var/log/crowdsec.log"
 LAPI_LOG_FILE="/var/log/crowdsec_api.log"
-CROWDSEC_PLUGIN_DIR="/var/lib/crowdsec/plugins/"
+CROWDSEC_PLUGIN_DIR="${CROWDSEC_USR_DIR}/plugins"
```
blotus, line 26: We will need to update the wizard patches to also change `${CROWDSEC_USR_DIR}`

rpm/SPECS/crowdsec.spec
```diff
@@ -39,6 +39,7 @@ BuildRequires:  systemd
 
 %build
 BUILD_VERSION=%{local_version} make
+sed -i "s#/usr/local/lib/crowdsec/plugins/#/usr/lib/crowdsec/plugins/#g" config/config.yaml
```
blotus, line 42: Why sed and not a patch ?

wizard.sh
sabban: this should be ok, because we don't need it when packaging

rpm/SPECS/crowdsec.spec
sabban: because it can be either `/usr/lib` or `/usr/lib64`. But still, this needs using macros...

blotus (approved):

PR #948, opened 2021-09-08
don't install all items from hub when upgrade --force

### LINKED ISSUE #851: Bug/cscli hub upgrade --force : install all items ?
After running `cscli hub upgrade --force` it installed all the collections and items to my machine. I would expect it only to upgrade my existing items.

AlteredCoder:
fix #851


sabban (approved):
lgtm

PR #983, opened 2021-09-21
update golang.org/x/sys dep

blotus:
This fixes a very weird crash in logrus when running `cscli` on x86 macos (see https://github.com/sirupsen/logrus/issues/1275).


sabban (approved):
lgtm

PR #986, opened 2021-09-23
fix usage of regex.Match in cloudwatch module

blotus:
The parameters were inverted, which made the `stream_regexp` config parameter unusable.


sabban (approved):
lgtm

PR #996, opened 2021-10-04
add functionnal test for hubtest

AlteredCoder:
Add a fonctionnal test script that run `cscli hubtest run -all` for packaging


buixor (changes requested):
I think we need to `cd hub/` :)

buixor (approved):

Issue #1066, opened 2021-12-01
Bug/failing to add a bouncer exits with retcode 0


mmetc:
```
$ sudo cscli bouncers add foobar
ERRO[01-12-2021 10:05:19 AM] unable to create bouncer: bouncer foobar already exists 
$ echo $?
0

```

PR #1067, opened 2021-12-01
fatal if cannot add bouncer: fix for #1066

mmetc:


mmetc:
It fails because the current tests check for "error" instead of "fatal"

buixor:
Please fix the tests :smile:

mmetc:
should be ok now

buixor (approved):
lgtm

PR #1079, opened 2021-12-11
Remove additional `apk update` in Dockerfile

PeterDaveHello:
There is no need to use `apk update` with `apk add --no-cache` when using `apk` on Alpine Linux, as using `--no-cache` will fetch the index every time and leave no local cache, so the index will always be the latest without temporary files remains in the image.

BTW, this was done in #689 before, but seemed to be reverted in #829 by unknown reason, I guess it's just accident?


blotus (approved):

blotus:
Thank you @PeterDaveHello, I have no idea why I changed it :)

Issue #1094, opened 2021-12-13
Bug/cscli / invalid memory address or nil pointer dereference


thib3113:
**Describe the bug**

I just try to install my first parser (usually I only use a bouncer, and the capi list) . 

A litle point about my installation : 
 - local_api is in a docker container (with the official image)
 - the parser is on another machine, with local_api set to the local_api host (crowdsec installed, server disabled)
 - the bouncer is on a third machine (no cscli)

First, I had some problems to install it, crowdsec fail to found `crowdsecurity/traefik-logs`, and I need to restart crowdsec (remove/install for the machine, reboot for the docker container) .

After that, I succeed to install it, and when I try to do some tests : 
`cscli explain --file /opt/traefik/logs/access.log --type traefik --trace -c config.yaml`
( cwd is in /etc/crowdsec ) .

and I get : 

```
DEBU[13-12-2021 08:46:37 PM] Using config.yaml as configuration file
time="2021-12-13T20:46:37Z" level=error msg="crowdsec - goroutine crowdsec/main crashed : runtime error: invalid memory address or nil pointer dereference"
time="2021-12-13T20:46:37Z" level=error msg="please report this error to https://github.com/crowdsecurity/crowdsec/"
time="2021-12-13T20:46:37Z" level=error msg="stacktrace/report is written to /tmp/crowdsec-crash.586038343.txt : please join it to your issue"
time="2021-12-13T20:46:37Z" level=fatal msg="crowdsec stopped"
time="2021-12-13T20:46:37Z" level=fatal msg="crowdsec stopped"

FATA[13-12-2021 08:46:37 PM] fail to run crowdsec for test: exit status 1
```

the `/tmp/crowdsec-crash.586038343.txt` : 
```
error : runtime error: invalid memory address or nil pointer dereference
version: v1.2.1-debian-pragmatic-dd03d073558e380c283afe66942f537c3da647ff
Codename: alphaga
BuildDate: 2021-11-18_10:12:27
GoVersion: 1.16.7
Platform: linux
goroutine 1 [running]:
runtime/debug.Stack(0xc000010c50, 0xc00013c6e0, 0x9d)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/crowdsecurity/crowdsec/pkg/types.WriteStackTrace(0x12e7a20, 0x1da5990, 0x14b3ce0, 0x46)
[...]

PR #1096, opened 2021-12-14
Fix null point error if server cfg is empty while doing cscli explain

sbs2001:
Fixes https://github.com/crowdsecurity/crowdsec/issues/1094


buixor (approved):
lgtm

PR #1144, opened 2022-01-04
fix race condition on repetitive trigger buckets creation

buixor:
As seen recently on specific scenarios, a stream of log that would repetitively trigger creation of trigger buckets for the same partition key might lead to "orphaned" buckets because of the way `sync.Map` behaves. Those orphaned buckets would stay around forever, leaking memory.

This fixes the issue while attempting the cleanup a bit the code (too many nested loops).


sabban (approved):
lgtm

PR #1146, opened 2022-01-05
fix #1131 : complain when validating unknown machine

### LINKED ISSUE #1131: Bug/cscli: No warning when validating unknown machine
**Describe the bug**
The cscli command produces no error when validating a machine that is not registered, i don't think this is supposed to be the correct behavior. 

**To Reproduce**
1. Execute `cscli machines validate random_string_here`
2. See 'INFO[timestamp] machine 'random_string_here' validated successfuly'

**Expected behavior**
I think there should be an error raised when validating an un-registered machine

**Screenshots**
image

**Technical Information**
 - OS: Ubuntu Server 20.04
 - Version v1.2.1-debian-pragmatic-dd03d073558e380c283afe66942f537c3da647ff

buixor:


sabban (approved):
gtm

PR #1159, opened 2022-01-12
Windows Support

blotus:
TODO:
 - [x] Sign binaries + installers
 - [x] Move configuration to %PROGRAMDATA%


config/config_win.yaml
```diff
@@ -0,0 +1,48 @@
+common:
+  daemonize: true
+  pid_dir: /var/run/
+  log_media: file
+  log_level: info
+  log_dir:  C:\Program Files\Crowdsec\log\
+  working_dir: .
+config_paths:
+  config_dir:  C:\Program Files\Crowdsec\config\
```
LogicalPhallacy, line 9: You may want to consider defaulting configs, dynamic data, logs etc... to a subfolder of C:\ProgramData instead of writing to Program Files. AppData would normally be recommended, but since it looks like you're using a windows service per system model rather than a per user model, ProgramData would likely make more sense.

blotus: Thanks for the suggestion !

pkg/acquisition/modules/file/checkaccess.go
```diff
@@ -0,0 +1,12 @@
...
+// +build linux freebsd netbsd openbsd solaris !windows
+
+package fileacquisition
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func checkAccess(file string) error {
+	return unix.Access(file, unix.R_OK)
```
buixor, line 11: I think this conflicts with https://github.com/crowdsecurity/crowdsec/pull/1179

pkg/acquisition/modules/wineventlog/wineventlog.go
```diff
@@ -0,0 +1,59 @@
+//go:build !windows
+
+package wineventlogacquisition
```
buixor, line 3: don't we want to error if someone tries to invoke it on non-windows platform ?

buixor (changes requested):
minor remarks, need to find time to do a full review :)

pkg/acquisition/modules/wineventlog/wineventlog.go
blotus: This is already done here: https://github.com/crowdsecurity/crowdsec/blob/e18a6286dd0234ce728f26c7707ba31b44d3fd5a/pkg/acquisition/modules/wineventlog/wineventlog.go#L50

pkg/acquisition/modules/wineventlog/wineventlog_windows.go
```diff
@@ -0,0 +1,356 @@
...
+
+type WinEventLogSource struct {
+	config    WinEventLogConfiguration
+	logger    *log.Entry
+	evtConfig *winlog.SubscribeConfig
+	query     string
+	name      string
+}
+
+type WindowsEvent struct {
```
buixor: unused

pkg/acquisition/modules/wineventlog/wineventlog_windows.go
```diff
@@ -0,0 +1,356 @@
...
+				for _, event := range renderedEvents {
+					linesRead.With(prometheus.Labels{"source": w.name}).Inc()
+					l := types.Line{}
+					l.Raw = event
+					l.Module = w.GetName()
+					l.Labels = w.config.Labels
+					l.Time = time.Now()
+					l.Src = w.name
+					l.Process = true
+					if !w.config.UseTimeMachine {
```
buixor, line 197: just a slight suggestion : we could have a helper at the `config` obj level to deal with the LIVE / TIMEMACHINE FLAG in the object ?

cmd/crowdsec/main.go
```diff
@@ -10,7 +10,6 @@ import (
 	_ "net/http/pprof"
 	"time"
 
-	"github.com/confluentinc/bincover"
```
sabban, line 13: Can you check that this is ok ? Maybe this code moved ?

cmd/crowdsec/main.go
```diff
@@ -278,41 +273,5 @@ func main() {
...
-
-	if exitCode, err := Serve(cConfig); err != nil {
-		if err != nil {
-			log.Errorf(err.Error())
-			if !bincoverTesting {
-				os.Exit(exitCode)
-			}
-			bincover.ExitCode = exitCode
-		}
-	}
```
sabban, line 317: same here

pkg/time/rate/rate_test.go
```diff
@@ -182,6 +182,9 @@ func TestLongRunningQPS(t *testing.T) {
 		t.Skip("low resolution time.Sleep invalidates test (golang.org/issue/14183)")
 		return
 	}
+	if runtime.GOOS == "windows" {
+		t.Skip("test is unreliable on windows")
+	}
```
sabban, line 187: This file is stolen from golang sources with some minor modification. Is this not windows compatible ? It may even be worth checking the newer version of golang if it needs updating.

scripts/test_env.ps1
```diff
@@ -0,0 +1,90 @@
...
+$parser_s01="$parser_dir\s01-parse"
+$parser_s02="$parser_dir\s02-enrich"
+$scenarios_dir="$config_dir\scenarios"
+$postoverflows_dir="$config_dir\postoverflows"
+$hub_dir="$config_dir\hub"
+$plugins=@("http", "slack", "splunk")
+$plugins_dir="plugins"
+$notif_dir="notifications"
+
+
```
sabban, line 78: I don't see a check that crowdsec binary exists. Maybe it's worth checking that ?

buixor (approved):
LGTM

sabban (approved):
minor comments

Issue #1165, opened 2022-01-14
Bug/Weird timezone handling in the database


blotus:
We do not handle very coherently a LAPI using a a timezone other than UTC.

For example, when using Mysql:
 - We use `parseTime=True` in the mysql driver, which will parse the `Time` objects and convert them to UTC
 - But we rely on the `Default` configuration of ent to automatically fill the `created_at` fields, which for some reason will not get converted to UTC (I'm not 100% sure, but this seems the most logical explanation).
 - This lead to `created_at` having a different TZ than the other fields.

We should replace the default value of `Time.Now` with something like `Time.Now().UTC`.

We also need to investigate if this issue occurs with sqlite or postgres.

buixor:
We most likely want to :
 - set ent's `created_at` & so to a custom func that does `time.Now().UTC()`
 - and convert existing `time.Now()` in the code to `time.Now().UTC()` too

Doing only the first half we lead to some issues if ent's driver doesn't parse dates (as mysql does)

buixor:
cf. https://github.com/crowdsecurity/crowdsec/pull/1167

AlteredCoder:
fixed in #1167

PR #1167, opened 2022-01-16
switch to utc time everywhere

buixor:


pkg/time/rate/rate.go
```diff
@@ -124,9 +124,9 @@ func NewLimiter(r Limit, b int) *Limiter {
 	}
 }
```
buixor, line 126: oh my bad for the replace, I'm not sure we want to change the time things in this package ...

buixor (commented):
requested changes

buixor:
fix #1165

AlteredCoder (approved):

Issue #1180, opened 2022-01-18
Bug/notification-email enabling causes crowdsec/main to crash 


jpalko:
**Describe the bug**
I am trying to enable notification-email and once I have modified /etc/crowdsec/profiles.yaml and /etc/crowdsec/notifications/email.yaml and try to restart the service systemd tells that it does not start.

**To Reproduce**
1. Configure /etc/crowdsecnotifications/email.yaml for google mail delivery
2. Uncomment in profiles.yaml:
```
notifications:
 - email_default
```
3. systemd restart crowdsec
4. See error:
```
Job for crowdsec.service failed because the control process exited with error code.
See "systemctl status crowdsec.service" and "journalctl -xeu crowdsec.service" for details.
```

**Expected behavior**
Service starts and email notifications get enabled.

**Technical Information (please complete the following information):**
 - OS: Fedora
 - Version 35
 - SELinux targeted policy enabled

**Additional context**
/var/log/crowdsec.log:
```
time="18-01-2022 18:55:37" level=info msg="Crowdsec v1.2.3-linux-a6e405422c732e9b6c46ae1004a2b80297df8336"
time="18-01-2022 18:55:37" level=info msg="Loading prometheus collectors"
time="18-01-2022 18:55:37" level=info msg="Loading CAPI pusher"
time="18-01-2022 18:55:37" level=info msg="initiating plugin broker"
time="18-01-2022 18:55:37" level=error msg="crowdsec - goroutine crowdsec/main crashed : runtime error: invalid memory address or nil pointer dereference"
time="18-01-2022 18:55:37" level=error msg="please report this error to https://github.com/crowdsecurity/crowdsec/"
time="18-01-2022 18:55:37" level=error msg="stacktrace/report is written to /tmp/crowdsec-crash.349495449.txt : please join it to your issue"
time="18-01-2022 18:55:37" level=fatal msg="crowdsec stopped"
```

/tmp/crowdsec-crash.349495449.txt:
```
error : runtime error: invalid memory address or nil pointer dereference
version: v1.2.3-a6e405422c732e9b6c46ae1004a2b80297df8336
Codename: alphaga
BuildDate: 2022-01-06_10:23:53
GoVersion: 1.17.5
Platform: linux
goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.17.5/x64/src/runtime/debug/stack.go:24 +0x65
[...]

PR #1181, opened 2022-01-19
Fix panic broker

sbs2001:
Partially fixes https://github.com/crowdsecurity/crowdsec/issues/1180 . It won't panic and crash main process, however plugins still won't work.


pkg/csplugin/broker.go
```diff
@@ -406,8 +406,14 @@ func getProcessAtr(username string, groupname string) (*syscall.SysProcAttr, err
 	}
 	g, err := user.LookupGroup(groupname)
 	if err != nil {
-		return nil, err
+		if groupname == "nogroup" {
```
buixor, line 409: this part is too hackish to be fixed in code. let's deal with this specific part in packaging please :)

buixor (changes requested):
ok for the crash fix, but don't hardcode fallback user/group in code, but at package level please ;)

buixor (approved):
lgtm

PR #1198, opened 2022-01-24
fix fc35 package

sabban:
I believe this patchset should fix fc35 package, as enabling crowdsec will create the needed preset file.


buixor (approved):
lgtm

PR #1216, opened 2022-01-30
Prestage files and copy on init to fix bind mount issues

TheSpad:
Currently if you try and init a new Crowdsec container using bind mounts (or tmpfs) rather than named volumes, it fails. This is due to [this](https://docs.docker.com/storage/volumes/#populate-a-volume-using-a-container) difference in behaviour between named volumes and bind mounts. In essence a named volume will copy the contents of the container directory into the volume on first run but a bind mount will instead mount the host directory to *replace* the container directory and its contents.

This change copies the built files for /etc/crowdsec and /var/lib/crowdsec into a /staging folder on build, and then on init it checks for the presence of files in the target location and copies them from /staging if they're not already present. That way anyone using a bind mount will still be able to init the container without having to take manual steps.

I've picked a lack of `config.yaml` and `local_api_credentials.yaml` as proxies for "is the container setup yet" but you're welcome to select other files, or even to individually check and copy every file in /etc/crowdsec (though I feel that would be overkill).

Tested as much as I can locally as everything seems to work as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,18 +13,20 @@ RUN SYSTEM="docker" make release
 RUN cd crowdsec-v* && ./wizard.sh --docker-mode && cd -
 RUN cscli hub update && cscli collections install crowdsecurity/linux && cscli parsers install crowdsecurity/whitelists
 FROM alpine:latest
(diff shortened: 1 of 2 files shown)


nitescuc (approved):
Very good contribution, we appreciate it. Many thanks :-)

PR #1220, opened 2022-01-31
string comparison fix

mmetc:
This is definitely a bug:

`if [ false=true ]`

(without spaces around '=') always tests true


buixor:
nice catch, did you see if/how it changes behavior ?

mmetc:
> nice catch, did you see if/how it changes behavior ?

It always calls install_collection, no matter what

buixor (approved):

PR #1229, opened 2022-02-01
Set custom hostname for local agent credentials

woopstar:
Add a new environment variable to set custom hostname for the local API when regenerating the credentials


buixor (approved):

buixor:
awesome thanks

PR #1233, opened 2022-02-02
WIP: Create debian docker package with journalctl

woopstar:
This PR will add a build step to crowdsec to release a Debian-based docker container that contains systemd, which will include journalctl.

This is used if you want to stream journalctl log files inside the docker container to do collections on.

For this to work, we will need to use `bash` instead of `sh` in the docker start script but should have minimal impact on the alpine container.

It will also add a new ENV variable to disable scenarios. Usecase for this is:

You enable the Nginx collection. It has the scenarios in it. When the container starts, it will re-enable all scenarios in the collection even though you have disabled some of them.

### DIFF
--- a/.github/workflows/release_publish_docker-image.yml
+++ b/.github/workflows/release_publish_docker-image.yml
@@ -9,7 +9,7 @@ jobs:
     name: Push Docker image to Docker Hub
     runs-on: ubuntu-latest
     steps:
-      - 
+      -
         name: Check out the repo
         uses: actions/checkout@v2
       -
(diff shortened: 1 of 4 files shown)


docker/docker_start.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
he2ss, line 1: Now the alpine based Dockerfile will need bash to run this script.

he2ss (changes requested):

docker/docker_start.sh
woopstar: See `Dockerfile` above. `bash` is already added

he2ss (approved):

.github/workflows/release_publish_docker-image.yml
```diff
@@ -56,4 +56,17 @@ jobs:
...
\ No newline at end of file
+            org.opencontainers.image.revision=${{ github.sha }}
+      -
+        name: Build and push debian image with journalctl
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile.debian
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.prep.outputs.tags }}-debian
```
woopstar, line 67: @he2ss is the tagging correct here for pushing the debian image?

he2ss: yep, seems a good tag name. :+1: .

PR #1254, opened 2022-02-10
Upgrade download datafiles if doesn't exist

AlteredCoder:
If we change a datafile in an existing parser/scenario, cscli will not download the new datafile when running `cscli <parsers|scenarios> upgrade` . This PR fix this issue.


buixor (approved):
lgtm

Issue #1245, opened 2022-02-08
Bug/crowdsec raspberrypi install & wizzard


nitescuc:
**Describe the bug**
1. Running the wizzard from command line complains about envsubst binary is needed ...
2. Not attempting to start crowdsec, port 8080 is already used or lapi was disabled message even though the ports have been changed into config files

**To Reproduce**
Port 8080 is in use on the system
1. sudo apt update && sudo apt install crowdsec
2. got message "Not attempting to start ..." (correct)
3. Change config.yaml + local_api_credentials.yaml to move to 8042
4. sudo apt remove crowdsec
5. Check that configuration files are not removed (ok)
6. sudo apt install crowdsec
7. got message "Not attempting to start ..." (wrong as configuration has been modified to listen on 8042)
8. Attempt to run /usr/share/crowdsec/wizard.sh -c
9. got "FATA[08/02/2022:10:27:46] crowdsec_wizard: envsubst binary is needed to use do a full install with the wizard, exiting ..."

**Expected behavior**
Installing crowdsec when the configuration has been modified to listen on a free port shouldn't warn about port 8080 being already used and installation should continue
Should be able to launch the wizard without error

**Screenshots**

**Technical Information (please complete the following information):**
 - OS: Raspbian Buster
 - Version 1.3.0

**Additional context**
Crowdsec does start normally with sudo service crowdsec start

sabban:
Hi,

I tried but didn't manage to reproduce step7.

Concerning step 9, it's not bug it was meant to be a feature to choose not add gettext dependency.

sabban:
This should be fixed by https://github.com/crowdsecurity/crowdsec/pull/1258

PR #1258, opened 2022-02-11
fix upgrade when somthing is already listening on 8080

sabban:
* this fix the bug https://github.com/crowdsecurity/crowdsec/issues/1245 by getting the real used port from config file.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -89,7 +89,7 @@ if [ "$1" = configure ]; then
 
     systemctl --quiet is-enabled crowdsec || systemctl unmask crowdsec && systemctl enable crowdsec
 
-    PORT=$(cat /etc/crowdsec/config.yaml|grep listen_uri |cut -d ":" -f3)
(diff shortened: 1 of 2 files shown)


debian/postinst
```diff
@@ -89,10 +89,11 @@ if [ "$1" = configure ]; then
 
     systemctl --quiet is-enabled crowdsec || systemctl unmask crowdsec && systemctl enable crowdsec
 
-    if [ -z "$(ss -nlt 'sport = 8080' | grep -v ^State)" ] || [ "$LAPI" = false ]; then
+    PORT=$(cat /etc/crowdsec/config.yaml|grep listen_uri |cut -d ":" -f3)
```
buixor, line 92: why not using `cscli config show` to achieve this ? less error prone I'd guess (+ we will want the same for RPMs)

buixor (changes requested):
question

debian/postinst
sabban: yes, you're right, let's do this.

buixor (approved):

PR #1260, opened 2022-02-11
add IpToRange helpers and allows to have an expression with scope Range

AlteredCoder:


buixor (changes requested):
nice, can we add some pkg/leakybucket/ tests for this feature plz ?

buixor (approved):
lgtm !

PR #1277, opened 2022-02-17
Send all installed scenario to LAPI

### LINKED ISSUE #1276: Bug/Crowdsec : tainted scenarios do not appear in the console's instance card
If you have a tainted scenario, it will not appear in the console's "scenario" list.

AlteredCoder:
fix #1276


blotus:
We should send all scenarios to LAPI (regardless of their status), but custom scenarios should only be sent to CAPI if custom is enabled in the console config (same thing for tainted).

buixor (approved):

Issue #1169, opened 2022-01-17
Bug/crowdsec: syslog datasource doesn't reload/restart properly


FinweVI:
**Describe the bug**
I'm using the syslog datasource.
```
---
source: syslog
listen_addr: 0.0.0.0
listen_port: 514
labels:
 type: syslog
```

So far it's working fine, though when I `reload` CrowdSec, the datasource seems to be stopped and never restarted (no activity on the server running crowdsec).
If I `restart` the process, it's fixing the issue, though it usually takes systemd to `SIGKILL` crowdsec for a proper restart.
```
janv. 17 10:44:43 crowdsec.kayrros systemd[1]: Stopping Crowdsec agent...
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: crowdsec.service: State 'stop-sigterm' timed out. Killing.
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: crowdsec.service: Killing process 60625 (crowdsec) with signal SIGKILL.
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: crowdsec.service: Main process exited, code=killed, status=9/KILL
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: crowdsec.service: Failed with result 'timeout'.
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: Stopped Crowdsec agent.
janv. 17 10:46:13 crowdsec.kayrros systemd[1]: Starting Crowdsec agent...
janv. 17 10:46:15 crowdsec.kayrros systemd[1]: Started Crowdsec agent.
```

**To Reproduce**
Mostly vanilla setup handling high amount of syslog messages over the network.
Then reload/restart through `systemctl`

**Expected behavior**
A reload shouldn't break the syslog datasource.
A restart shouldn't need a SIGKILL to work.

**Technical Information (please complete the following information):**
 - OS: Ubuntu Bionic
 - Version: 1.2.3

**Additional context**
This is a single crowdsec instance ingesting all of the syslog messages from a few hundreds servers.
The high volume of message could be an issue.

PR #1288, opened 2022-02-24
Exit syslog acquis only after server is dead

sbs2001:
Fixes https://github.com/crowdsecurity/crowdsec/issues/1169

To reproduce the above issue, syslog server needs decent traffic. This can be simulated by running the following python script :
```python
import subprocess
while True:
	subprocess.Popen(['logger', '"asdaad"', '-d', '-n', '127.0.0.1', '-P', '514'])
```

The issue is caused because:
The server tomb wait call at https://github.com/crowdsecurity/crowdsec/blob/9a42190e13b26a39182125caebb301b132ec39ae/pkg/acquisition/modules/syslog/syslog.go#L188 never finishes . This is because it is blocked at https://github.com/crowdsecurity/crowdsec/blob/9a42190e13b26a39182125caebb301b132ec39ae/pkg/acquisition/modules/syslog/internal/syslogserver.go#L78 . This channel is being listened to at https://github.com/crowdsecurity/crowdsec/blob/9a42190e13b26a39182125caebb301b132ec39ae/pkg/acquisition/modules/syslog/syslog.go#L195 but it never get "selected" because the select is stuck at the wait. 

The stuff for s.serverTomb.Dying() is removed because it's harmful to exit before the server is really dead (port isn't freed).

### DIFF
--- a/pkg/acquisition/modules/syslog/internal/syslogserver.go
+++ b/pkg/acquisition/modules/syslog/internal/syslogserver.go
@@ -61,7 +61,7 @@ func (s *SyslogServer) StartServer() *tomb.Tomb {
 		for {
 			select {
 			case <-t.Dying():
-				s.Logger.Info("syslog server tomb is dying")
+				s.Logger.Info("Syslog server tomb is dying")
 				err := s.KillServer()
 				return err
 			default:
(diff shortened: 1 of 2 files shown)


buixor:
makes complete sense, thanks :+1: 

I'll let @blotus review, LGTM

blotus (approved):

buixor (approved):
lgtm

PR #1293, opened 2022-02-28
Warn when log file in explain command is large.

sbs2001:


cmd/crowdsec-cli/explain.go
```diff
@@ -108,3 +112,16 @@ cscli explain --dsn "file://myfile.log" --type nginx
 
 	return cmdExplain
 }
+
+func getLineCountForFile(filepath string) int {
```
buixor, line 116: might need a defer'ed close ? might bail out after X lines anyway (to avoid losing time here)

buixor (changes requested):
suggestion/question

cmd/crowdsec-cli/explain.go
sbs2001: thx, fixed now

buixor (approved):

PR #1297, opened 2022-02-28
Email notification plugin add sender_name option

tuxtof:
Ability to define sender_name for the email notification plugin
to fix issue email coming from `From`

### DIFF
--- a/plugins/notifications/email/main.go
+++ b/plugins/notifications/email/main.go
@@ -40,6 +40,7 @@ type PluginConfig struct {
 	SMTPUsername   string   `yaml:"smtp_username"`
 	SMTPPassword   string   `yaml:"smtp_password"`
 	SenderEmail    string   `yaml:"sender_email"`
+	SenderName     string   `yaml:"sender_name"`
 	ReceiverEmails []string `yaml:"receiver_emails"`
 	EmailSubject   string   `yaml:"email_subject"`
 	EncryptionType string   `yaml:"encryption_type"`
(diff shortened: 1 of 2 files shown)


tuxtof:
linked PR for documentation https://github.com/crowdsecurity/crowdsec-docs/pull/184

buixor (approved):
LGTM, thanks !

plugins/notifications/email/main.go
```diff
@@ -88,7 +89,7 @@ func (n *EmailPlugin) Notify(ctx context.Context, notification *protobufs.Notifi
 	logger.Debug("smtp connection done")
 
 	email := mail.NewMSG()
-	email.SetFrom(fmt.Sprintf("From <%s>", cfg.SenderEmail)).
+	email.SetFrom(fmt.Sprintf("%s <%s>", cfg.SenderName, cfg.SenderEmail)).
```
sbs2001, line 92: Just to not break existing installations, setting a default `SenderName` would be nice. Either here or even better in the `Configure` method right after unmarshal. 

```go
if cfg.SenderName == "" {
   cfg.SenderName = "CrowdSec" 
}
```

sbs2001 (changes requested):

plugins/notifications/email/main.go
tuxtof: Yes that's a good remark, but as i had not seen any config input validation or default settings, I had continued like this
it can be clearly improved I agree

tuxtof: just implement default settings and validation

tuxtof:
Based on @sbs2001 comment i implement default config and config validation

sbs2001 (approved):
LGTM. Thanks @tuxtof

tuxtof:
guys @sbs2001 @buixor do you think you can merge it for the next release ??

PR #1300, opened 2022-03-02
Copy changes to NGINXACCESS pattern from parser

Dominic-Wagner:
fix crowdsecurity/hub#394
The grok line in the nginx-logs parser got changed, but these changes got not reflected to this pattern.
The traefik parser uses this pattern, which leads to missed http-events because of the missing http verb/method.
The traefik, apache2 and nginx parsers, and http-scenarios all use the field 'verb' instead of the field 'method'.

The hubtest will fail with this PR, as afterwards I need to create an PR with the correct parsers.assert for traefik-logs.yaml.

### DIFF
--- a/config/patterns/nginx
+++ b/config/patterns/nginx
@@ -13,7 +13,7 @@ DAY2 \d{2}
 #NGINXERRTIME %{YEAR:year}/%{MONTHNUM2:month}/%{DAY2:day} %{HOUR:hour}:%{MINUTE:minute}:%{SECOND:second}
 NGINXERRTIME %{YEAR}/%{MONTHNUM2}/%{DAY2} %{HOUR}:%{MINUTE}:%{SECOND}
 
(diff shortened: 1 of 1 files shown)


Dominic-Wagner:
linked PR for documentation crowdsecurity/crowdsec-docs#187

buixor:
thanks, reviewing it :+1:

sabban:
Hi,

Sorry for the waiting time, but this PR has implication in other location as the hub, and may need further investigations or changes there.

Regards,

he2ss:
Hi @Dominic-Wagner,

To avoid waiting for the next crowdsec release to have this change and to prevent some break changes in some users custom scenarios based on this grok.
We had another approche (https://github.com/crowdsecurity/hub/pull/414). It will also fix the issue, if it's ok for you, we can merge it and it will be immediatly available in the hub.

You can close this one + the PR in the documentation.

Dominic-Wagner:
Great, closed because of crowdsecurity/hub#414

PR #1311, opened 2022-03-04
Fix api for all dbs (#1310)

buixor:
* DB agnostic lapi sanitize
* Update entgo + code generation

Signed-off-by: Shivam Sandbhor <shivam.sandbhor@gmail.com>


sabban:
This has to be tested before merging.

PR #1312, opened 2022-03-04
Docker prestage - correct database path

chad-jones:
Correct GeoIP database path - currently the database is being overwritten each time the container starts due to this


buixor (approved):

he2ss:
@chad-jones Thanks for the contribution.

Issue #1314, opened 2022-03-04
Bug/ fail to upgrade from 1.3.0 to 1.3.2


thib3113:
**Describe the bug**
Fail to upgrade from 1.3.0 to 1.3.2

here is the error : 
```
Paramétrage de crowdsec (1.3.2) ...
Updating hub
INFO[04-03-2022 04:52:21 PM] Wrote new 309255 bytes index to /etc/crowdsec/hub/.index.json
INFO[04-03-2022 04:52:21 PM] dependency of crowdsecurity/base-http-scenarios : outdated parsers crowdsecurity/http-logs
INFO[04-03-2022 04:52:21 PM] dependency of crowdsecurity/linux : sub collection crowdsecurity/sshd is broken : outdated parsers crowdsecurity/sshd-logs
INFO[04-03-2022 04:52:21 PM] dependency of crowdsecurity/traefik : outdated parsers crowdsecurity/traefik-logs
INFO[04-03-2022 04:52:21 PM] dependency of crowdsecurity/sshd : outdated parsers crowdsecurity/sshd-logs
INFO[04-03-2022 04:52:21 PM] dependency of crowdsecurity/nginx : sub collection crowdsecurity/base-http-scenarios is broken : outdated parsers crowdsecurity/http-logs
FATA[04-03-2022 04:52:22 PM] cannot fetch ListenURI from *csconfig.LocalApiServerCfg (1:19)
 | Config.API.Server.ListenURI
 | ..................^
dpkg: erreur de traitement du paquet crowdsec (--configure) 
```

**To Reproduce**
1. Disable local Server
2. upgrade

**Technical Information (please complete the following information):**
 - OS: ubuntu 20.04.4

**Additional context**
-> nothing in logs

To pass the upgrade, I've enable local server, and then re-disabling it

sabban:
I could reproduce this, next version will be release fixed

PR #1320, opened 2022-03-07
Fix 1314

sabban:
fix bug #1314 in postinst : 
In the case of Crowdsec not serving a LAPI, `cscli config show --key Config.API.Server.ListenURI` returns an error. This PR prevents this to be run when `Config.API.Server` returns `<nil>` (ie when local API Server is disabled).

Tests show that it works on rpm as well. This has to be tested again before next release.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -89,9 +89,13 @@ if [ "$1" = configure ]; then
 
     systemctl --quiet is-enabled crowdsec || systemctl unmask crowdsec && systemctl enable crowdsec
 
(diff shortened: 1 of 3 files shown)


sabban:
* In the case of Crowdsec not serving a LAPI, `config show --key Config.API.Server.ListenURI` returns an error. This PR prevents this to be run when Config.API.Server is <nil>.

buixor:
Thanks for the PR, don't we want the same logic for RPMs ?

buixor (approved):

PR #1323, opened 2022-03-08
make linux-collection and whitelists-parser can be disabled in docker container

vmolkov:
It's useful sometimes to be able to make linux collection and whitelists parser to not start in docker container, for example for testing purposes or when using custom log format that can't be handled by "default" parsers from linux collection.


buixor:
Thanks for the PR @vmolkov !

Do you mind documenting it in the `docker/README.md` too ?

vmolkov:
Found out that there are already exist DISABLE_* variables that make exactly what i need. Closing PR

PR #1329, opened 2022-03-09
some noop code removal, typos and lint suggestions

mmetc:


pkg/time/rate/rate.go
```diff
@@ -179,7 +179,6 @@ func (r *Reservation) DelayFrom(now time.Time) time.Duration {
 // Cancel is shorthand for CancelAt(time.Now()).
 func (r *Reservation) Cancel() {
 	r.CancelAt(time.Now())
```
sabban, line 181: please don't touch this file.
It's a fork from standard golang library that we should sync from upstream from time to time. We forked it because we wanted some more specific feature used in our bucket system. We should even propose our changes upstream.

sabban (changes requested):

pkg/time/rate/rate.go
mmetc: done, I've reverted that

sabban (approved):
lgtm

PR #1341, opened 2022-03-10
Register bouncers on container init

TheSpad:
This PR allows users to add bouncers on container init, rather than having to exec into the container and run `cscli`, for automated deployment scenarios. `jq` is added to the Dockerfile to support parsing cscli output.

Supports both environment variables (in the format `BOUNCER_KEY_<NAME>=<API-KEY>`) and docker secrets (in the format `BOUNCER_KEY_<name>` with the contents `<API-KEY>`). Adding multiple bouncers and mixing environment and secrets are supported (though environment will take precedence in the event of conflicting names).

The init script checks that both a name and a key value have been provided, then it checks to see if there is already an existing bouncer with that name registered (in which case it skips it), and then registers the bouncer with the NAME and KEY provided.

This allows you to do something like:
```yml
services:
  crowdsec:
    image: docker.io/crowdsecurity/crowdsec:latest
    container_name: crowdsec
    environment:
      - BOUNCER_KEY_traefik=mysecretkey12345

  bouncer-traefik:
    image: docker.io/fbonalair/traefik-crowdsec-bouncer:latest
    container_name: crowdsec-bouncer-traefik
    environment:
      - CROWDSEC_BOUNCER_API_KEY=mysecretkey12345
    depends_on:
      - crowdsec
```
On first init of the crowdsec container the logs should output something along the lines of:
```
Api key for 'traefik':

   mysecretkey12345

Please keep this key since you will not be able to retrieve it!
```
And `cscli bouncers list` should show:
```
--------------------------------------------------------------
 NAME  IP ADDRESS  VALID  LAST API PULL         TYPE  VERSION 
--------------------------------------------------------------
 traefik             ✔️   2022-03-09T20:55:12Z                
--------------------------------------------------------------
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN SYSTEM="docker" make release
 RUN cd crowdsec-v* && ./wizard.sh --docker-mode && cd -
 RUN cscli hub update && cscli collections install crowdsecurity/linux && cscli parsers install crowdsecurity/whitelists
 FROM alpine:latest
-RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community tzdata yq bash && \
+RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community tzdata yq jq bash && \
   mkdir -p /staging/etc/crowdsec && \
   mkdir -p /staging/var/lib/crowdsec
 COPY --from=build /etc/crowdsec /staging/etc/crowdsec
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -123,7 +123,9 @@ for BOUNCER in $(compgen -A variable | grep -i BOUNCER_KEY); do
     NAME=$(printf '%s' "$BOUNCER" | awk -F "_" '{printf $3}')
     if [[ -n $KEY ]] && [[ -n $NAME ]]; then
         if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o json | jq -r .[].name | grep -q "${NAME}"; then
-            if ! cscli -c "$CS_CONFIG_FILE" bouncers add "${NAME}" -k "${KEY}" > /dev/null; then
+            if cscli -c "$CS_CONFIG_FILE" bouncers add "${NAME}" -k "${KEY}" > /dev/null; then
+                echo "Registered bouncer for ${NAME}"
+            else
                 echo "Failed to register bouncer for ${NAME}"
             fi
         fi
@@ -137,7 +139,9 @@ for BOUNCER in /run/secrets/@(bouncer_key|BOUNCER_KEY)* ; do
     NAME=$(echo "${BOUNCER}" | awk -F "/" '{printf $NF}' | awk -F "_" '{printf $3}')
     if [[ -n $KEY ]] && [[ -n $NAME ]]; then    
         if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o json | jq -r .[].name | grep -q "${NAME}"; then
-            if ! cscli -c "$CS_CONFIG_FILE" bouncers add "${NAME}" -k "${KEY}" > /dev/null; then
+            if cscli -c "$CS_CONFIG_FILE" bouncers add "${NAME}" -k "${KEY}" > /dev/null; then
+                echo "Registered bouncer for ${NAME}"
+            else
                 echo "Failed to register bouncer for ${NAME}"
             fi
         fi
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -120,7 +120,7 @@ fi
 ## Register bouncers via env
 for BOUNCER in $(compgen -A variable | grep -i BOUNCER_KEY); do
     KEY=$(printf '%s' "${!BOUNCER}")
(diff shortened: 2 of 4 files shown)


zerkms:
May be `> /dev/null` the output so that secrets didn't appear in logs?

TheSpad:
> May be `> /dev/null` the output so that secrets didn't appear in logs?

I'm of two minds on this but it is probably the right way to go, I just need to add some error handling so that it notifies you properly in the logs if it can't register the bouncer.

Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN SYSTEM="docker" make release
 RUN cd crowdsec-v* && ./wizard.sh --docker-mode && cd -
 RUN cscli hub update && cscli collections install crowdsecurity/linux && cscli parsers install crowdsecurity/whitelists
 FROM alpine:latest
-RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community tzdata yq bash && \
+RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community tzdata yq jq bash && \
```
woopstar, line 16: Also required in the Dockerfile.debian maybe?

TheSpad: Good catch, the debian dockerfile didn't exist when I did my last PR so I missed it.

woopstar:
README.md file also needs to be updated with the new functionality.

he2ss:
Hi @TheSpad,

Isn't better to have `BOUNCER_NAME` and `BOUNCER_KEY`. So you will avoid additional code just to parse the bouncer name.

TheSpad:
> Hi @TheSpad,
> 
> Isn't better to have `BOUNCER_NAME` and `BOUNCER_KEY`. So you will avoid additional code just to parse the bouncer name.

Then you can't easily add more than one bouncer, unless you do numeric keys or arrays, but in either case it's probably just as much code to parse it out.

docker/docker_start.sh
```diff
@@ -117,6 +117,38 @@ if [ "$DISABLE_POSTOVERFLOWS" != "" ]; then
     cscli -c "$CS_CONFIG_FILE" postoverflows remove $DISABLE_POSTOVERFLOWS
 fi
 
+## Register bouncers via env
+for BOUNCER in $(compgen -A variable | grep -i BOUNCER_KEY); do
+    KEY=$(printf '%s' "${!BOUNCER}")
+    NAME=$(printf '%s' "$BOUNCER" | cut -d_  -f2-)
+    if [[ -n $KEY ]] && [[ -n $NAME ]]; then
+        if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o json | jq -r .[].name | grep -q "${NAME}"; then
```
he2ss, line 125: You can refactor this code (125-130) with a function and call it from those 2 contexts.

he2ss:
@TheSpad yes, you're right. I think we can accept this.
I just added one comment and I think it's fine :+1:

TheSpad:
> @TheSpad yes, you're right. I think we can accept this. I just added one comment + if you can add documentation about this env var in the readme (`docker/README.md`), so the readme in the dockerhub also will be updated please.

Done.

TheSpad:
Can I give this a nudge - is there anything else you need me to do to be able to merge this PR?

he2ss (approved):
Thanks @TheSpad for the contribution :+1:

karkazzul:
Hello,
I have tested your code like this:
```
services:
  crowdsec:
    image: docker.io/crowdsecurity/crowdsec:latest
    container_name: crowdsec
    environment:
      - BOUNCER_KEY_traefik=mysecretkey12345

  bouncer-traefik:
    image: docker.io/fbonalair/traefik-crowdsec-bouncer:latest
    container_name: crowdsec-bouncer-traefik
    environment:
      - CROWDSEC_BOUNCER_API_KEY=mysecretkey12345
      - CROWDSEC_AGENT_HOST=crowdsec:8080
    depends_on:
      - crowdsec
   
```

And then i check the bouncers list in the crowdsec container, I have nothing.

TheSpad:
> Hello, I have tested your code like this:
> 
> ```
> services:
>   crowdsec:
>     image: docker.io/crowdsecurity/crowdsec:latest
>     container_name: crowdsec
>     environment:
>       - BOUNCER_KEY_traefik=mysecretkey12345
> 
>   bouncer-traefik:
>     image: docker.io/fbonalair/traefik-crowdsec-bouncer:latest
>     container_name: crowdsec-bouncer-traefik
>     environment:
>       - CROWDSEC_BOUNCER_API_KEY=mysecretkey12345
>       - CROWDSEC_AGENT_HOST=crowdsec:8080
>     depends_on:
>       - crowdsec
>    
> ```
> 
> And then i check the bouncers list in the crowdsec container, I have nothing.

AFAIK there hasn't been an updated release of the image since this was merged so it won't work yet.

karkazzul:
thanks for your answer i have using the github repos and it's works !
it's just the docker image that's it's not updated

karkazzul:
Do you have an idea when the image on docker hub was updated with these modifications ?

TheSpad:
It'll presumably be whenever they next do a point release, not sure what their release cycle is like.

PR #1342, opened 2022-03-10
dummy plugin

mmetc:


sabban:
dummy plugin shouldn't be included to packages.

sabban (approved):
lgtm

Issue #1206, opened 2022-01-25
Bug/PIDFile shouldn't be used in systemd service


FinweVI:
On the systemd service file you are using the `PIDFile` directive while it's not really recommended anymore (https://www.freedesktop.org/software/systemd/man/systemd.service.html#PIDFile=)
Also the `pid_dir` configuration option seems to default to `/var/run`

**Describe the bug**
It can lead to errors such as: `systemd[1]: crowdsec.service: Can't open PID file /run/crowdsec.pid (yet?) after reload: No such file or directory`, when reloading the agent.

**Expected behavior**
No error when reloading the agent.

**Technical Information (please complete the following information):**
 - OS: Ubuntu Bionic
 - Version: 1.2.3

Athanasius:
As best I can tell the contents of `pid_dir` *aren't even used*, at least not in the core code in this repo.

There's:

`pkg/csconfig/common.go: PidDir       string     `yaml:"pid_dir"``

and then:

> pkg/csconfig/common.go: PidDir       string     `yaml:"pid_dir"`
> pkg/csconfig/common.go:         &c.Common.PidDir,

But nothing else referencing PidDir.  So no wonder systemd is complaining there's no /run/crowdsec.pid, because it appears nothing is writing it.

I'm mostly interested in this due to finding that a `kill -HUP` of the main crowdsec process does indeed cause it to reload its configuration.  So of course I need the PID in order to send the signal.  This is for setting up a logrotate script (I'll open another issue about the need for such in the Debian packages at least).

Athanasius:
Ah, OK, the systemd unit *does* define ExecReload, so I can just `systemctl reload crowdsec.service` for this.

Still, if `pid_dir` isn't even used then it should be removed from the code and the documentation (save a note about how it was removed).  Else the code should be fixed to actually use the directive, and write the PID file.

buixor:
sorry, this has been merged :)

Athanasius:
I note that current release, `crowdsec_1.4.1`, at least the Debian package, *still* includes:

```yaml
pid_dir: /var/run/
```
in `config.yaml`.  It's also still present at https://github.com/crowdsecurity/crowdsec/blob/master/config/config.yaml#L3

And that then causes:
```
Jul 28 13:07:36 river crowdsec[3552657]: time="2022-07-28T13:07:36+01:00" level=warning msg="Deprecation warning: the pid_dir config can be safely removed and is not required"
```

[...]

PR #1346, opened 2022-03-14
Deprecate pid_file config

sbs2001:
Addresses https://github.com/crowdsecurity/crowdsec/issues/1206

### DIFF
--- a/cmd/crowdsec/main.go
+++ b/cmd/crowdsec/main.go
@@ -251,6 +251,10 @@ func LoadConfig(cConfig *csconfig.Config) error {
 		log.Infof("single file mode : log_media=%s daemonize=%t", cConfig.Common.LogMedia, cConfig.Common.Daemonize)
 	}
 
+	if !flags.TestMode && cConfig.Common.PidDir != "" {
+		log.Warn("Deprecation warning: the pid_dir config can be safely removed and is not required")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


cmd/crowdsec/main.go
```diff
@@ -251,6 +251,10 @@ func LoadConfig(cConfig *csconfig.Config) error {
 		log.Infof("single file mode : log_media=%s daemonize=%t", cConfig.Common.LogMedia, cConfig.Common.Daemonize)
 	}
 
+	if !flags.TestMode && cConfig.Common.PidDir != "" {
```
buixor, line 254: why the specific check on testmode ?

pkg/csconfig/common.go
```diff
@@ -11,7 +11,7 @@ import (
 /*daemonization/service related stuff*/
 type CommonCfg struct {
 	Daemonize    bool
-	PidDir       string     `yaml:"pid_dir"`
+	PidDir       string     `yaml:"pid_dir,omitempty"`
```
buixor, line 14: maybe add a commz here so we don't forget about backward compat

cmd/crowdsec/main.go
sbs2001: For some reason I though it would create duplicate logs during systemctl re(load/start). Removed the check on test mode in latest commit

pkg/csconfig/common.go
sbs2001: Done

sabban (approved):
lgtm

buixor (approved):

Issue #1348, opened 2022-03-14
Improvement/Having a way to disable a scenario without tainting a collection


buixor:
**Is your feature request related to a problem? Please describe.**
Currently, disabling a specific scenario that is part of a collection without tainting the whole collection isn't possible. The alternative (using simulation mode) might not be enough depending on your case (you might want to *really* disable the scenario, not just ignore the decisions)

mmetc:
This might be already possible depending on the user case.
Why is it important for the collection to be "not tainted"?

For example in the "hubops" branch, I could override a scenario with a local one of the same name, which does nothing. I can then install/remove any collection that depends on it without issues, but they'll still be tainted.

PR #1350, opened 2022-03-14
[wip] Disable a scenario/parser without tainting a collection

AlteredCoder:
Fix https://github.com/crowdsecurity/crowdsec/issues/1348

Note: Currently, `cscli <item_type> remove` doesn't really remove an item but disable it (we need to provide `--purge` to remove the item).

- Don't consider a collection as tainted if we have remove it without `--purge`

TODO: check behavior on item or collection upgrade

### DIFF
--- a/pkg/cwhub/cwhub.go
+++ b/pkg/cwhub/cwhub.go
@@ -217,6 +217,10 @@ func ItemStatus(v Item) (string, bool, bool, bool) {
 		strret = "enabled"
 	}
 
+	if v.Downloaded && !v.Installed {
+		strret += ",downloaded"
+	}
+
 	Managed := true
 	if v.Local {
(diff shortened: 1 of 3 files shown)


buixor:
@AlteredCoder : yes I think upgrade needs to be tested

Lamera:
What happens with CAPI decisions, from removed scenarios? Are they applied anyway?

he2ss:
@Lamera if you decide to disable a scenario, so there will not be a new decisions triggered by this scenario anymore.
This PR purpose is just to avoid having a collection is state `tainted` when a scenario is only disabled and not customized.

buixor:
we're going to revamp cwhub and hence we don't want to invest more time on fixing the current code

PR #1356, opened 2022-03-15
bit of doc + tests for lapi stream mode

buixor:
- [x] Minimal tests for stream mode
 - [x] Add tests for `scope` parameter

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -49,13 +49,30 @@ repositories).
 | :----------------- | :--------------- | :----------- |
 | alerts GET/POST    | `9[78]_ipv[46]*` |              |
 | decisions GET/POST | `9[78]_ipv[46]*` |              |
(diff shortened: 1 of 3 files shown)


mmetc:
dependencies: also bash, python3, openbsd-netcat. daemonize is available as rpm too, and the equivalent "daemon" is preinstalled on freebsd.

jq_out should be "api_out"

If you want to test both new and deleted together:

run jq -c '[.deleted[0].value, [.new[].value]]' <(output)
assert_output '["1.2.3.4",["1111:2222:3333:4444:5555:6666:7777:8888","1.2.4.0/24","1.2.3.5"]]'

we should renumber the tests sometime, to group them together in a sensible way. or subdirectories but I like it flat when there are few files

mmetc (approved):

PR #1358, opened 2022-03-15
deal with misconfigured post-overflow stages

### LINKED ISSUE #1357: Bug/Crash on postoverflow stages misconfiguration
**Describe the bug**
 - Have a setup with postoverflows stages `s00-xxx` `s01-xxx` and `s01-zzzz`
 - Have whitelists (or not) in `s01-xxx` but that are not going to trigger
 - Have at least one whitelist in `s01-zzzz`

Crowdsec will crash when an event reaches `s01-zzzz` because `Parse` will see that an event reaches stage `s01-zzz` while its previous stage is `s00-xxx` (so it "jumped" a stage and is thus consider not to have parsed correctly). It will return an empty object.

In the meanwhile, it will try to print and access various fields back in output.go and trigger a null deref :

https://github.com/crowdsecurity/crowdsec/blob/master/cmd/crowdsec/output.go#L155

(reported by @TheSpad )

buixor:
- fix #1357

### DIFF
--- a/pkg/parser/runtime.go
+++ b/pkg/parser/runtime.go
@@ -269,7 +269,7 @@ func Parse(ctx UnixParserCtx, xp types.Event, nodes []Node) (types.Event, error)
 				StageParseCache[stage] = make(map[string][]ParserResult)
 			}
 		}
-		/* if the node is forward in stages, seek to its stage */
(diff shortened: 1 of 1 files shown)


sabban (approved):
lgtm

PR #1366, opened 2022-03-17
use yq instead of envsubst (and don't damage plugin config)

mmetc:
the long term object is to have minimal, well defined fixtures

but also required because it doesn't use config-templates (so we always start from the up-to-date configuration) and avoids a broken interpolation in the plugin configurations (there where $stuff that was wrongly interpreted as references to envvar)


sabban (approved):
lgtm

PR #1372, opened 2022-03-18
Mysql enabled tests

sabban:
add the ability to do func tests against mysql database.

This is done by adding needed cleanup, setup, backup, restore and configure functions for mysql, and call them when required (We're using DB_BACKEND env var for this). It'll be easy to add other backend the exact same way.

### DIFF
--- a/.github/workflows/ci_bats_mysql.yaml
+++ b/.github/workflows/ci_bats_mysql.yaml
@@ -17,7 +17,7 @@ jobs:
       mariadb:
         image: mysql:latest
         env:
-          MARIADB_ROOT_PASSWORD: mysql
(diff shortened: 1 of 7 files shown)


mmetc:
- don't use CONFIG_DIR, you have CONFIG_YAML now
 - I think setup_database() and setup_configuration(), especially creating database and user, should be moved outside instance_mysql but we can go now with a version that works and I can rearrange stuff later when I add postgres. Maybe have instance-mysql take care of the setup, and fixture/mysql manage backup/restore.
 - wrong about()
 - can you fix the failing test with sqlite or do I look into it?

mmetc (approved):

PR #1376, opened 2022-03-20
add dialect to handle pgx correctly

umglurf:
#1273 broke pgx database driver, as it needs to set dialect postgresql for ent to work. This adds a dialect flag to getEntDriver so that the dialect can be set correctly.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


buixor:
Thanks @umglurf ! We're adding functional tests for various DB backends to avoid this kind of issues, we'll merge soon :+1:

mmetc (approved):

mmetc:
Done. Thanks again!

PR #1377, opened 2022-03-20
bats changes to db tests

mmetc:
split CI/bats in sqlite+mysql+hub; renamed db and user 'crowdsec' to 'crowdsec-test'; moved backend-specific scripts to lib/db, etc.


mmetc:
it does not seem that the mysql and postgres tests are failing for the same reason, but both succeed outside CI. need to investigate :-/

sabban (approved):
lgtm

sabban (approved):
lgtm

PR #1384, opened 2022-03-23
allow capi register in functional tests

mmetc:
This allows to set a recognizable value for tests, in all possible contexts including where machine-id cannot be changed (non-root) or could break systems (dev machines). When all code that creates test instances is converted we can set new users with machineid.protectedID() which sends a hash instead for increased privacy.

### DIFF
--- a/cmd/crowdsec-cli/machines.go
+++ b/cmd/crowdsec-cli/machines.go
@@ -60,28 +60,36 @@ func generatePassword(length int) string {
 	return string(buf)
 }
 
-// Generate a unique identifier, composed by a unique string for each OS
-// installation where possible (prefix), and a random suffix. The prefix can be
-// random if it cannot be derived, and can also be a parameter to allow for test
-// environments.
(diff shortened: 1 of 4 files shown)


cmd/crowdsec-cli/machines.go
```diff
@@ -60,20 +60,27 @@ func generatePassword(length int) string {
...
+// environments.
+func generateID(prefix string) (string, error) {
+	var err error
+	if prefix == "" {
+		prefix, err = machineid.ID()
 		if err != nil {
-			return "", errors.Wrap(err, "generating machine id")
+			log.Debugf("failed to get machine-id with usual files : %s", err)
+		}
+		if prefix == "" || err != nil {
```
blotus, line 74: Redundant check on `prefix == ""`

mmetc: how about now?

sabban (approved):
lgtm

PR #1405, opened 2022-03-31
Update bouncer pull in rupture mode too

sbs2001:


pkg/apiserver/controllers/v1/decisions.go
```diff
@@ -54,7 +66,16 @@ func (c *Controller) GetDecision(gctx *gin.Context) {
 	} else {
 		PrometheusBouncersHasEmptyDecision(gctx)
 	}
-
+	go func() {
```
blotus, line 69: I've move this call after the check on HEAD requests, because we don't send anything to the bouncer if the method is HEAD, I wouldn't really count it as a real pull (and I'm not sure bouncers will do HEAD requests by themselves ?)

pkg/apiserver/controllers/v1/decisions.go
```diff
@@ -54,7 +66,16 @@ func (c *Controller) GetDecision(gctx *gin.Context) {
...
+		bouncerKey := gctx.Request.Header.Get(c.APIKeyHeader)
+		if bouncerInfo, err := APIKeyToBouncer(bouncerKey, c.DBClient); err == nil {
+			c.BouncerPullUpdator.UpdateBouncerPullEntry(database.BouncerKey{
+				ID: bouncerInfo.ID, Name: bouncerInfo.Name,
+			})
+		} else {
+			log.Errorf("unable to fetch bouncer for key %s: %v", bouncerKey, err)
```
blotus: Don't log the API key, it won't be really be useful for debug.

sbs2001: Fixed

pkg/apiserver/controllers/v1/decisions.go
sbs2001: Fixed

pkg/database/pull_updator.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	for {
+		select {
+		case <-ticker.C:
+			lpt.Lock()
+			for bouncer, pullTime := range lpt.bouncerPullTime {
+				if err := dbClient.UpdateBouncerLastPull(pullTime.UTC(), bouncer.ID); err != nil {
+					log.Errorf("unable to update bouncer '%s' pull: %v", bouncer.Name, err)
+				} else {
+					log.Debugf("Updated pull for %s bouncer", bouncer.ID)
```
blotus, line 50: Please fix linter warning

AlteredCoder:
I think that this is too overkill for what we want.

[Here](https://github.com/crowdsecurity/crowdsec/blob/master/pkg/apiserver/middlewares/v1/api_key.go#L49) we already get the bouncer from the database. I think we can just check if `bouncer.last_pull` is greater than 1 minute for example and if yes update it.

PR #1424, opened 2022-04-06
Fix plugin default config to support group threshold

sbs2001:
Fixes https://discourse.crowdsec.net/t/http-plugin-configuration-group-wait-and-group-threshold-problem/649/3

This error was caused due to somewhat wrong default config.
Old logic was:

If group_wait  is absent then set group_wait = 1s . This is fine in majority of cases. However when group_wait is absent and group_threshold is present, then also the code was setting group_wait = 1s . Thus rectifying group_threshold semantic.

New logic is:

Set group_threshold=1 only if group_threshold and group_wait is absent. Otherwise use provided config.

### DIFF
--- a/pkg/csplugin/broker.go
+++ b/pkg/csplugin/broker.go
@@ -356,8 +356,8 @@ func setRequiredFields(pluginCfg *PluginConfig) {
 		pluginCfg.TimeOut = time.Second * 5
 	}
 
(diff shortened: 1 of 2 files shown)


buixor:
@sbs2001 can we add a functional or unit test to ensure non regression ? thanks !

buixor:
dup

PR #1437, opened 2022-04-13
Improve plugins grouping (alternative to #1424)

buixor:
Fixes https://discourse.crowdsec.net/t/http-plugin-configuration-group-wait-and-group-threshold-problem/649/3

This error was caused due to somewhat wrong default config.

New logic is:
 - Simplify the watcher : only the ticker-based watcher can trigger events processing
 - Ensure w/ extra tests that treshold can be set by "time", "count", "time+count" or none

### DIFF
--- a/pkg/csplugin/broker_test.go
+++ b/pkg/csplugin/broker_test.go
@@ -519,7 +519,7 @@ func TestBrokerRunTimeThreshold(t *testing.T) {
 	}
 }
 
-func TestBrokerRun(t *testing.T) {
+func TestBrokerRunSimple(t *testing.T) {
 	DefaultEmptyTicker = 50 * time.Millisecond
 	buildDummyPlugin()
 	setPluginPermTo744()
--- a/pkg/csplugin/broker_test.go
+++ b/pkg/csplugin/broker_test.go
@@ -285,14 +285,14 @@ func TestBrokerNoThreshold(t *testing.T) {
 	setPluginPermTo744()
 	defer tearDown()
 	//init
-	procCfg := csconfig.PluginCfg{}
+	pluginCfg := csconfig.PluginCfg{}
 	pb := PluginBroker{}
 	profiles := csconfig.NewDefaultConfig().API.Server.Profiles
 	profiles = append(profiles, &csconfig.ProfileCfg{
 		Notifications: []string{"dummy_default"},
 	})
 	//default config
-	err := pb.Init(&procCfg, profiles, &csconfig.ConfigurationPaths{
+	err := pb.Init(&pluginCfg, profiles, &csconfig.ConfigurationPaths{
 		PluginDir:       testPath,
 		NotificationDir: "./tests/notifications",
 	})
@@ -336,7 +336,7 @@ func TestBrokerRunGroupAndTimeThreshold_TimeFirst(t *testing.T) {
 	defer tearDown()
 
 	//init
-	procCfg := csconfig.PluginCfg{}
+	pluginCfg := csconfig.PluginCfg{}
 	pb := PluginBroker{}
 	profiles := csconfig.NewDefaultConfig().API.Server.Profiles
 	profiles = append(profiles, &csconfig.ProfileCfg{
@@ -347,7 +347,7 @@ func TestBrokerRunGroupAndTimeThreshold_TimeFirst(t *testing.T) {
 	cfg.GroupThreshold = 4
 	cfg.GroupWait = 1 * time.Second
 	writeconfig(t, cfg, "tests/notifications/dummy.yaml")
-	err := pb.Init(&procCfg, profiles, &csconfig.ConfigurationPaths{
+	err := pb.Init(&pluginCfg, profiles, &csconfig.ConfigurationPaths{
 		PluginDir:       testPath,
 		NotificationDir: "./tests/notifications",
 	})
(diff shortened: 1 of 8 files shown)


pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+	log.Printf("second send")
+	pb.PluginChannel <- ProfileAlert{ProfileID: uint(0), Alert: &models.Alert{}}
+	log.Printf("third send")
+	pb.PluginChannel <- ProfileAlert{ProfileID: uint(0), Alert: &models.Alert{}}
+	time.Sleep(500 * time.Millisecond)
+	//because of group threshold, we shouldn't have data yet
+	assert.NoFileExists(t, "./out")
+	log.Printf("before sleep")
+	time.Sleep(1 * time.Second)
+	//after 8 seconds, we should have data
```
sbs2001, line 371: Comment doesn't seem to relevant anymore

pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+
+func TestBrokerNoThreshold(t *testing.T) {
+	var alerts []models.Alert
+	DefaultEmptyTicker = 50 * time.Millisecond
+
+	buildDummyPlugin()
+	setPluginPermTo744()
+	defer tearDown()
+	//init
+	procCfg := csconfig.PluginCfg{}
```
sbs2001, line 288: :nit Name and value don't match.

pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+
+func TestBrokerRunGroupAndTimeThreshold_TimeFirst(t *testing.T) {
+	//test grouping by "time"
+	DefaultEmptyTicker = 50 * time.Millisecond
+	buildDummyPlugin()
+	setPluginPermTo744()
+	defer tearDown()
+
+	//init
+	procCfg := csconfig.PluginCfg{}
```
sbs2001, line 340: :nit Name and value don't match.

pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+	}
+}
+
+func TestBrokerRunGroupAndTimeThreshold_CountFirst(t *testing.T) {
+	DefaultEmptyTicker = 50 * time.Millisecond
+	buildDummyPlugin()
+	setPluginPermTo744()
+	defer tearDown()
+	//init
+	procCfg := csconfig.PluginCfg{}
```
sbs2001, line 392: :nit Name and value don't match.

pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+
+func TestBrokerRunGroupThreshold(t *testing.T) {
+	//test grouping by "size"
+	DefaultEmptyTicker = 50 * time.Millisecond
+	buildDummyPlugin()
+	setPluginPermTo744()
+	defer tearDown()
+
+	//init
+	procCfg := csconfig.PluginCfg{}
```
sbs2001, line 447: :nit Name and value don't match.

pkg/csplugin/broker_test.go
```diff
@@ -252,7 +255,284 @@ func TestBrokerInit(t *testing.T) {
...
+	}
+}
+
+func TestBrokerRunTimeThreshold(t *testing.T) {
+	DefaultEmptyTicker = 50 * time.Millisecond
+	buildDummyPlugin()
+	setPluginPermTo744()
+	defer tearDown()
+	//init
+	procCfg := csconfig.PluginCfg{}
```
sbs2001, line 495: :nit Name and value don't match.

sbs2001 (changes requested):
Some nitpicks about style. Tests cases look good to me

pkg/csplugin/broker_test.go
```diff
@@ -276,10 +556,16 @@ func TestBrokerRun(t *testing.T) {
...
-	assert.FileExists(t, "./out")
-	assert.Equal(t, types.GetLineCountForFile("./out"), 2)
+	content, err := ioutil.ReadFile("./out")
+	if err != nil {
+		log.Errorf("Error reading file: %s", err)
+	}
+	var alerts []models.Alert
+	err = json.Unmarshal(content, &alerts)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(alerts))
```
sbs2001, line 568: Just thinking out loud. This means Threshold=1 is not honored and the watcher went with ticker event. Had it followed threshold = 1 then plugin would've fired 2 separate events and `assert.Equal(t, types.GetLineCountForFile("./out"), 2)` would pass.

pkg/csplugin/watcher.go
```diff
@@ -42,15 +50,62 @@ func (pw *PluginWatcher) Start(tomb *tomb.Tomb) {
...
+		//only time is set
+		watchTime = interval
+	} else if interval != 0 && threshold != 0 {
+		//both are set
+		watchTime = DefaultEmptyTicker
+		watchCount = threshold
+	} else {
+		//none are set, we sent every event we receive
+		watchTime = DefaultEmptyTicker
+		watchCount = 1
```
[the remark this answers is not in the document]
buixor: which one you think is best?

pkg/csplugin/broker_test.go
buixor: It's because both arrive in burst, we already have another test checking that one item .. sleep ...  one item leads to two different writes

sabban (approved):
lgtm

sbs2001 (approved):

PR #1443, opened 2022-04-14
improve test packaging

sabban:
* few minor bug fixes


tests/bats.mk
```diff
@@ -63,8 +63,10 @@ endef
 bats-all: bats-clean bats-build bats-fixture bats-test bats-test-hub
 
 # Source this to run the scripts outside of the Makefile
+# Old version of make doesn't have $(file) directive
```
mmetc, line 66: don't

tests/bats/02_nolapi.bats
```diff
@@ -76,7 +76,7 @@ declare stderr
 
 @test "$FILE lapi status shouldn't be ok without api.server" {
     yq 'del(.api.server)' -i "${CONFIG_YAML}"
-    ./instance-crowdsec start
+    ./instance-crowdsec start || true
```
mmetc, line 79: if crowdsec start fails, shouldn't we fail the test too?

tests/lib/config/config-global
```diff
@@ -51,9 +51,13 @@ make_init_data() {
     ./instance-db config-yaml
     ./instance-db setup
 
+    [ $DB_BACKEND != sqlite ] && ${CSCLI} machines add --auto
```
mmetc, line 54: use the function is_db_mysql

mmetc (approved):

PR #1447, opened 2022-04-15
Reduce the query unescape helper verbosity

AlteredCoder:
- According to the net/url document, `QueryUnescape` returns an error if any % is not followed by two hexadecimal digits.


sabban (approved):
lgtm

buixor:
While it can be a band-aid, the fact that the parser isn't resilient is a real issue and it means we cannot rely on it on the long-term ...

PR #1466, opened 2022-04-20
support yq prior to 4.18.1

mmetc:


sabban (approved):
lgtm but imho we could have done with only recent yq

PR #1470, opened 2022-04-21
cleanup container state if the reader tomb dies by itself

### LINKED ISSUE #1469: Bug/Docker datasource: logs are not read if not data for 10 minutes
When using [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy), if a container that we are tailing does not emit any logs for 10 minutes, we won't see any of the logs that are written later.

This is because the haproxy configuration of the proxy has a 10 minutes timeout on the connections (see https://github.com/Tecnativa/docker-socket-proxy/blob/master/haproxy.cfg#L21).

We probably ignore an error somewhere, as `docker logs` properly handles this case and make another request to start the tail again.

blotus:
Closes #1469


sabban (approved):
lgtm

PR #1477, opened 2022-04-25
Improve MySQL performance

AlteredCoder:
- Change the `TIMESTAMP` type to `DATETIME` for the `decisions.until` column .


buixor:
did you run ent codegen ? surprised by the low code change amount

AlteredCoder:
> did you run ent codegen ? surprised by the low code change amount

Yes, the `pkg/database/ent/migrate/schema.go` file has been autogenerated

buixor (approved):

PR #1478, opened 2022-04-25
Improve distinct/uniq behaviour

buixor:
In the existing implementation, `distinct` effectiveness is directly related to the `cache_size` parameter. The `distinct` parses in "live" the queue of objects to determine the uniqueness. For a scenario such as `crawl-http-non-statics`, this can have a "huge" impact in terms of potential false-positives. This PR adds its own cache to the `distinct` filters, so that it's not impacted by the cache_size.


AlteredCoder (approved):

sabban (approved):
lgtm

PR #1497, opened 2022-04-29
fix #1385: .yaml.local

### LINKED ISSUE #1385: Improvement/Allow .local config file
Hi,

It could be usefull to allow usage of config file with .local to allow to override some value in the config file without touching the "official" file...
Because when crowdsec config file (or the bouncer) is updated this can cause issue.
Or you keep your current config file and you may loose some new value.
Or you overwrite with the maintainer version and then you loose your custom change...

The idea is to thave a file like config.yaml.local (or config.local.yaml as you wish) where you can place custom value.
Like changing the log dir or the backend. The same apply for the bouncer ofc.

mmetc:
This allows several configuration files (config.yaml, profiles.yaml, local_api_credentials..) to have a second .yaml.local file with values that override (or in the profiles case, complement) the base configuration.

### DIFF
--- a/pkg/csconfig/api.go
+++ b/pkg/csconfig/api.go
@@ -55,7 +55,7 @@ func (o *OnlineApiClientCfg) Load() error {
 }
 
 func (l *LocalApiClientCfg) Load() error {
-	patcher := yamlpatch.NewPatcher(l.CredentialsFilePath)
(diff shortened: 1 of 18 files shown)


buixor (approved):
lgtm

buixor (changes requested):
conflicts to be fixed + adding a meaningful description to the PR <3

buixor (approved):
great :)

PR #1501, opened 2022-05-02
error reporting

mmetc:
* Favor errors.Wrap over fmt.Error
 * Have crowdsec print critical errs on stderr regardless of log configuration (fixes a 1.3.3 regression)
 * Add context where useful
 * Add tests

### DIFF
--- a/cmd/crowdsec-cli/alerts.go
+++ b/cmd/crowdsec-cli/alerts.go
@@ -18,6 +18,7 @@ import (
 	"github.com/crowdsecurity/crowdsec/pkg/models"
 	"github.com/go-openapi/strfmt"
 	"github.com/olekukonko/tablewriter"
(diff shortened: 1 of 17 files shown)


cmd/crowdsec/main.go
```diff
@@ -308,7 +308,8 @@ func main() {
 
 	if exitCode, err := Serve(cConfig); err != nil {
 		if err != nil {
-			log.Errorf(err.Error())
+			log.Error(err)
+			fmt.Fprintln(os.Stderr, err)
```
buixor, line 312: as discussed, it overlaps with the existing code that adds a hook on logger : `log.AddHook` (cf. main)

buixor (changes requested):

buixor (approved):
lgtm

Issue #826, opened 2021-06-02
Improvement/`cscli dashboard setup` : check available resources and warn user 


buixor:
**Is your feature request related to a problem? Please describe.**
When the user is running `cscli dashboard setup` and the available resources are obviously lacking (people tend to think that the dashboard might not require more resources than crowdsec itself), we should warn the user.

**Describe the solution you'd like**
Warn the user and/or ask for a confirmation if the available resources aren't enough to run metabase

buixor:
I think we should use something like https://github.com/pbnjay/memory and check that > 1Gb of ram is present (as per metabase req's).

nanikjava:
@buixor Interested to work on this. 

So what we need to check is only memory ? like in free memory ? anything else we want to check ?

buixor:
Yes, I think RAM is enough !

LaurenceJJones:
Hey I want to get involved in a "good first issue", this still a feature you want to implement?

LaurenceJJones:
@buixor  Just to let you know I started working on implementing this however in the docs it states 1-2GB of ram [here](https://doc.crowdsec.net/docs/getting_started/crowdsec_tour/#deploy-dashboard=)? we should stay consistent

buixor:
solved and shipped in 1.4 !

PR #1513, opened 2022-05-07
Memory check for cscli dashboard setup

LaurenceJJones:
If total memory < 1GB and force-yes is not provided the user can answer Y/n to continue running the Metabase container if the force-yes flag is provided it will just warn the user that Metabase container requires 1-2GB

Please note the current calculation of memory is set to 1GB as per issue, however, in crowdsec documentation it states 1-2GB

EDIT: The free system memory function was acting a bit strange on my VM it listed only 57MB free instead of 570MB so for the first implementation I thought total memory can be an improvement.

I have done some more testing and the free memory is correct for my VM after system allocation. I can update this PR with the check for free system memory check as well, however, I will wait for feedback.

#826

### DIFF
--- a/cmd/crowdsec-cli/dashboard.go
+++ b/cmd/crowdsec-cli/dashboard.go
@@ -112,16 +112,20 @@ cscli dashboard setup -l 0.0.0.0 -p 443 --password <password>
 				}
 			}
 			var answer bool
-			if valid, err := checkSystemMemory(); err == nil && !valid && !forceYes {
-				prompt := &survey.Confirm{
-					Message: "Metabase requires 1-2GB of RAM, your system is below this requirement continue ?",
-					Default: true,
-				}
-				if err := survey.AskOne(prompt, &answer); err != nil {
-					log.Fatalf("unable to ask about RAM check: %s", err)
-				}
-				if !answer {
-					log.Fatal("Unable to continue due to RAM requirement")
+			if valid, err := checkSystemMemory(); err == nil && !valid {
+				if !forceYes {
+					prompt := &survey.Confirm{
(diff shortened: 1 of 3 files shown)


cmd/crowdsec-cli/dashboard.go
```diff
@@ -304,3 +323,14 @@ func passwordIsValid(password string) bool {
 	return true
 
 }
+
+func checkSystemMemory() (bool, error) {
+	totMem := memory.TotalMemory()
+	if totMem == 0 {
+		return true, errors.New("Unable to get system total memory")
+	}
+	if uint64(math.Pow(2, 30)) >= totMem {
```
LaurenceJJones, line 332: Don't know math.Pow is needed here as the value will always resolve to ```1073741824``` so we could swap this out and hard code the value

cmd/crowdsec-cli/dashboard.go
```diff
@@ -109,6 +112,22 @@ cscli dashboard setup -l 0.0.0.0 -p 443 --password <password>
...
 			}
 			var answer bool
+			if valid, err := checkSystemMemory(); err == nil && !valid {
+				if !forceYes {
+					prompt := &survey.Confirm{
+						Message: "Metabase requires 1-2GB of RAM, your system is below this requirement continue ?",
+						Default: true,
+					}
+					if err := survey.AskOne(prompt, &answer); err != nil {
+						log.Fatalf("unable to ask about RAM check: %s", err)
```
buixor, line 122: we might want to warn/error on this but not fatal : if for any reason the check errors, we're preventing the user from setting up the dashboard

buixor (changes requested):
minor suggestion, overall lgtm :+1:

he2ss (approved):
lgtm

PR #1517, opened 2022-05-10
Crowdsec/decisions_stream bug fix

he2ss:
This PR fixes 2 bugs:

- In decision stream, we updated the bouncer.last_pull before doing stream, which can lead to a race condition
- We updated the bouncer.last_pull 2 times: one time in the decisionStream function directly, another time in the Middleware if the bouncer.last_pull was older than 1 minutes. which also lead to bouncer not getting decisions.


buixor (approved):
LGTM

PR #1523, opened 2022-05-13
fix #1283: update and enable error reports from golangci

### LINKED ISSUE #1283: Bug/lint messages not reported by CI
If we execute "golangci-lint run" by hand we can see several errors that are not caught by CI probably due to the "only-new-issues: true" option.

Some of them are trivial, some are definitely worth a look (shadowed variables, unchecked errors..).

mmetc:
The golancgi-lint configuration file lists all available linters.

They are divided in lists: deprecated (we just ignore them), enabled (common sense, could find serious bugs, etc), too strict (we like our snake case in yaml), etc.

I have put some linters that I like in the "Recommended" list. Comment each line to enable them and see what happens. After fixing, we can move from Recommended to Enabled (or too strict, if we don't like the result).

etc.etc.

### DIFF
--- a/.github/workflows/ci_golangci-lint.yml
+++ b/.github/workflows/ci_golangci-lint.yml
@@ -19,14 +19,22 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/setup-go@v3
       - uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


buixor (approved):
lgtm

PR #1536, opened 2022-05-18
avoid double output (error + log fatal) and automatic --help after each error

mmetc:
Errors being in log format may be harder to read or parse while using interactive / scripted tools, so if we want to change I'm not opposed. But this PR keeps the behavior consistent with bin/crowdsec. 

Related: https://github.com/spf13/cobra/issues/914


buixor (approved):
lgtm. Agreed we might change it in the future, but at least it's now consistent

PR #1539, opened 2022-05-18
test multiple notification events with the same plugin

mmetc:
While looking how to apply yaml.local to the plugin configurations, it occurred to me that we support multiple documents in the same YAML files. It makes sense but I was not aware.
I am adding tests and updating the configuration files to be explicit about it. We could also mention it in the docs.


buixor (approved):
lgtm

Issue #402, opened 2020-11-10
Improvement/Automatically extend ipv6 IP to range when taking decision (ie. /56 /64 etc.)


buixor:
We should automatically extend decisions on a given IPV6 to the range it might be part of :
 - It will prevent attackers constantly switching IPs within the block to avoid getting blocked
 - It will as well prevent attackers from trying to pollute the ban base

registergoofy:
Some ISPs do not provide /48 or /56 as they should. Somehow these should get an exemption and only get to ban /64 ....

buixor:
:partying_face: https://github.com/crowdsecurity/crowdsec/pull/1540 has been merged and solves this for 1.4

buixor:
and this is the associated post-overflow to automate the process : https://github.com/crowdsecurity/hub/pull/475

PR #1540, opened 2022-05-18
Allow to set static to a pointer and add IsIPV6 helper

AlteredCoder:
This PRs add 2 things:

 - We can now set a static where the `target` is a Pointer
 - Adding an expr helper to check if an IP address is in V6 format
 
This will allow to solve this issue: https://github.com/crowdsecurity/crowdsec/issues/402

Example:

We can use the following postoverflow that will replace the `Alert.Source.Value` with the IPV6 range (`/64` for our example):

```yaml
onsuccess: next_stage
filter: "evt.Overflow.Alert.Remediation == true && IsIPV6(evt.Overflow.Alert.Source.IP)"
name: crowdsecurity/ipv6
description: ""
statics:
  - target: evt.Overflow.Alert.Source.Value
    expression: IpToRange(evt.Overflow.Alert.Source.IP, "/64")
  - target: evt.Overflow.Alert.Source.Scope
    value: Range
```

And then the use following profile to generate decisions on ranges:
```yaml
name: default_range_remediation
debug: true
filters:
 - Alert.Remediation == true && Alert.GetScope() == "Range"
decisions:
 - type: ban
   duration: 4h
   scope: range
on_success: break
```

### DIFF
--- a/pkg/parser/runtime.go
+++ b/pkg/parser/runtime.go
@@ -17,7 +17,6 @@ import (
 
 	"strconv"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/mohae/deepcopy"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/parser/runtime.go
```diff
@@ -65,11 +64,17 @@ func SetTargetByName(target string, value string, evt *types.Event) bool {
 		case reflect.Struct:
 			tmp := iter.FieldByName(f)
 			if !tmp.IsValid() {
-				log.Debugf("%s IsValid false", f)
+				log.Errorf("%s IsValid false", f)
```
buixor, line 67: if we want to set it to error level, I think we need something more explicit, maybe include `target` in the error message ?

buixor (changes requested):
overall lgtm, I think we might want to add the doc for the expr helper too before we merge ;)

PR #1549, opened 2022-05-20
[WIP: need some tests] add a way to test notifications

sabban:
* add a way to test notifications with cscli
* fix an antipattern in broker system with tomb
* kill an unneeded goroutine in the broker/plugin/atcher


cmd/crowdsec-cli/notifications.go
```diff
@@ -135,6 +143,81 @@ func NewNotificationsCmd() *cobra.Command {
...
 	}
 	cmdNotifications.AddCommand(cmdNotificationsInspect)
+	var cmdNotificationsReinject = &cobra.Command{
+		Use:               "reinject",
+		Short:             "reinject alerts into notifications system",
+		Long:              `Reinject alerts into notifications system`,
+		Example:           `cscli notifications reinject <alert_id> <plugin_name>`,
+		Args:              cobra.ExactArgs(2),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
```
mmetc, line 153: Can you use the RunE function and return errors instead of calling log.Fatal? So it can collect coverage output even from failures

PR #1554, opened 2022-05-20
[WIP]New syslog parser for syslog datasource

blotus:
- [ ] RFC3164
 - [ ] RFC5424


pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse.go
```diff
@@ -0,0 +1,256 @@
...
+
+	r.PRI = pri
+	return nil
+}
+
+func (r *RFC3164) parseTimestamp() error {
+	validTs := false
+	for _, layout := range VALID_TIMESTAMPS {
+		tsLen := len(layout)
+		if r.position+tsLen > r.len {
```
buixor, line 90: not sure it's  always right. Can it happen that an actual timestamp is shorter than the layout string ? what if hours are < 10 ? (not sure tho)

pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		hostname = append(hostname, c)
+		r.position++
+	}
+	if r.strictHostname {
+		if !utils.IsValidHostnameOrIP(string(hostname)) {
+			return fmt.Errorf("hostname is not valid")
+		}
+	}
+	if len(hostname) == 0 {
```
buixor, line 129: should this be only in the strictHostname case ? or it's invalid in any case ?

pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse.go
```diff
@@ -0,0 +1,256 @@
...
+			c = r.buf[r.position]
+			if c == ']' {
+				pidEnd = true
+				r.position++
+				break
+			}
+			if c < '0' || c > '9' {
+				return fmt.Errorf("pid inside tag must be a number")
+			}
+			pid = pid*10 + int(c-'0')
```
buixor, line 173: is this worth ?

pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse.go
```diff
@@ -0,0 +1,256 @@
...
+			r.position++
+		}
+	}
+
+	if hasPid && !pidEnd {
+		return fmt.Errorf("pid inside tag must be closed with ']'")
+	}
+
+	if hasPid {
+		r.PID = strconv.Itoa(pid)
```
buixor, line 183: as you Itoa it here ?

pkg/acquisition/modules/syslog/internal/parser/rfc5424/parse.go
```diff
@@ -0,0 +1,398 @@
...
+
+	if len(timestamp) == 0 {
+		return fmt.Errorf("timestamp is empty")
+	}
+
+	if r.position == r.len {
+		return fmt.Errorf("EOL after timestamp")
+	}
+
+	date, err := time.Parse(VALID_TIMESTAMP, string(timestamp))
```
buixor, line 123: from https://www.rfc-editor.org/rfc/rfc5424.html#section-6.2.3 and https://www.rfc-editor.org/rfc/rfc3339 I'm not sure that RFC3339Nano fits all the use cases ? Didn't look at the tests yet, but should at least be :

`1985-04-12T23:20:50.52Z`
`2003-10-11T22:14:15.003Z`
`1985-04-12T19:20:50.52-04:00`
`2003-08-24T05:14:15.000003-07:00`

pkg/acquisition/modules/syslog/internal/parser/rfc5424/parse.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+		msgid = append(msgid, c)
+		r.position++
+	}
+
+	if len(msgid) == 0 {
+		return fmt.Errorf("msgid is empty")
+	}
+
+	if len(msgid) > 128 {
```
buixor, line 251: from https://www.rfc-editor.org/rfc/rfc5424.html#section-6.2.3 seems MSGID can be only up to 32 CHARS

pkg/acquisition/modules/syslog/syslog.go
```diff
@@ -199,38 +183,24 @@ func (s *SyslogSource) handleSyslogMsg(out chan types.Event, t *tomb.Tomb, c cha
...
 			if err != nil {
-				logger.Debugf("could not parse as RFC5424 (%s)", err)
-				p = rfc3164.NewParser(rfc3164.WithYear(rfc3164.CurrentYear{}))
-				m, err = p.Parse(syslogLine.Message)
+				logger.Debugf("could not parse as RFC3164 (%s)", err)
+				p2 := rfc5424.NewRFC5424Parser()
+				err = p2.Parse(syslogLine.Message)
 				if err != nil {
 					logger.Errorf("could not parse message: %s", err)
 					logger.Debugf("could not parse as RFC3164 (%s) : %s", err, syslogLine.Message)
```
buixor, line 194: might not be related but given that we're pushing our parser, we want to allow users to log unparsed messages by setting the log_source to debug level so that they can open us issues :)

buixor (dismissed):
some suggestions but looks great overall

pkg/acquisition/modules/syslog/syslog.go
blotus: Raw messages are logged at the trace level because it can get very verbose very quickly (too much for debug level IMO)

pkg/acquisition/modules/syslog/internal/parser/rfc5424/parse.go
blotus: RFC3339Nano handles all those timestamps.

pkg/acquisition/modules/syslog/syslog.go
AlteredCoder, line 194: i think it is `rfc5424` here in the debug message.

PR #1556, opened 2022-05-23
Add duration expr to add duration formula

he2ss:
This feature will allow users to control the duration as en expr. So we can have something like this in profile :

```
name: default_ip_remediation
filters:
 - Alert.Remediation == true && Alert.GetScope() == "Ip"
decisions:
 - type: ban
   duration: 4h
duration_expr: sprintf('%dh', (GetDecisionsCount(Alert.GetValue()) + 1) * 4)
on_success: break
```

This configuration will allow the user to increase the duration by the number of times an IP was seen in the DB.
If a duration is given in `decisions` and there is `duration_expr` then the expr will be priority.

This also open possibilities on expression as the pkg exprhelpers has a dbclient now.


cmd/crowdsec/serve.go
```diff
@@ -227,6 +229,23 @@ func Serve(cConfig *csconfig.Config) error {
...
+		dbClient, err := database.NewClient(cConfig.API.Server.DbConfig)
+		if err != nil {
+			return errors.Wrap(err, "Failed to get database client")
+		}
+		err = exprhelpers.Init(dbClient)
+		if err != nil {
+			return errors.Wrap(err, "Failed to init expr helpers")
+		}
+	} else {
+		err := exprhelpers.Init(nil)
```
buixor, line 243: should we have an info or warning here ? this situation isn't too usual ?

pkg/csprofiles/csprofiles.go
```diff
@@ -36,7 +115,27 @@ func GenerateDecisionFromProfile(Profile *csconfig.ProfileCfg, Alert *models.Ale
 		}
 		/*some fields are populated from the reference object : duration, scope, type*/
 		decision.Duration = new(string)
-		*decision.Duration = *refDecision.Duration
+		if Profile.Cfg.DurationExpr != "" && Profile.RuntimeDurationExpr != nil {
+			duration, err := expr.Run(Profile.RuntimeDurationExpr, exprhelpers.GetExprEnv(map[string]interface{}{"Alert": Alert}))
+			if err != nil {
+				log.Warningf("failed to run duration_expr : %v", err)
```
buixor, line 121: we created a custom logger above (xlog / clog) shouldn't we use it ? (and same for further logging in this function)

pkg/csprofiles/csprofiles.go
```diff
@@ -12,10 +14,87 @@ import (
...
+func (Profile *Runtime) GenerateDecisionFromProfile(Alert *models.Alert) ([]*models.Decision, error) {
 	var decisions []*models.Decision
 
-	for _, refDecision := range Profile.Decisions {
+	if clog == nil {
+		xlog := log.New()
+		if err := types.ConfigureLogger(xlog); err != nil {
+			log.Fatalf("While creating profiles-specific logger : %s", err)
+		}
+		xlog.SetLevel(log.TraceLevel)
```
buixor, line 91: why set this one to trace ? and it doesn't seem to be used further ?

pkg/csprofiles/csprofiles_test.go
```diff
@@ -72,27 +146,44 @@ func TestEvaluateProfile(t *testing.T) {
...
 		},
+		{
+			name: "simple filter with decision_expr",
+			args: args{
+				profileCfg: &csconfig.ProfileCfg{
+					Filters: []string{"1==1"},
+					Decisions: []models.Decision{
+						{Type: &typ, Scope: &scope, Simulated: &boolFalse},
+					},
+					DurationExpr: "Sprintf('%dh', 4*4)",
```
buixor, line 169: is it possible to actually check the resulting duration we got ?

pkg/csprofiles/csprofiles.go
buixor: and we need to keep in mind that `profile` supports a `debug` flag that should influence logging

buixor (changes requested):
minor changes. LGTM

pkg/csprofiles/csprofiles.go
he2ss: Removed it from `GenerateDecisionFromProfile` func. We only need to init new logger when we evaluate profiles (for the expr debug part). And each profile have it own config, that's the reason why.

pkg/csprofiles/csprofiles.go
he2ss: Nope, I removed the custom logger from `GenerateDecisionFromProfile`. We need to create a new logger only when we evaluate each profile (as each profile has it own log level).

cmd/crowdsec/serve.go
```diff
@@ -227,6 +229,24 @@ func Serve(cConfig *csconfig.Config) error {
 	apiTomb = tomb.Tomb{}
 	crowdsecTomb = tomb.Tomb{}
 	pluginTomb = tomb.Tomb{}
+
+	if cConfig.API.Server != nil && cConfig.API.Server.DbConfig != nil {
+		dbClient, err := database.NewClient(cConfig.API.Server.DbConfig)
+		if err != nil {
+			return errors.Wrap(err, "Failed to get database client")
```
buixor: Sorry to be annoying, but we shouldn't capitalize error messages : https://github.com/golang/go/wiki/CodeReviewComments#error-strings (applies to bellow error messages as well)

pkg/csprofiles/csprofiles.go
```diff
@@ -12,10 +14,85 @@ import (
...
+	for _, profile := range profilesCfg {
+		var runtimeFilter, runtimeDurationExpr *vm.Program
+		var debugFilter, debugDurationExpr *exprhelpers.ExprDebugger
+		runtime := &Runtime{}
+		xlog := log.New()
+		if err := types.ConfigureLogger(xlog); err != nil {
+			log.Fatalf("While creating profiles-specific logger : %s", err)
+		}
+		xlog.SetLevel(log.InfoLevel)
+		runtime.Logger = xlog.WithFields(log.Fields{
```
buixor, line 41: the name of the profile itself in the logger fields would be cool :)

pkg/csprofiles/csprofiles.go
```diff
@@ -12,10 +14,85 @@ import (
...
+			}
+			runtime.RuntimeFilters[fIdx] = runtimeFilter
+			if profile.Debug != nil && *profile.Debug {
+				if debugFilter, err = exprhelpers.NewDebugger(filter, expr.Env(exprhelpers.GetExprEnv(map[string]interface{}{"Alert": &models.Alert{}}))); err != nil {
+					log.Debugf("Error compiling debug filter of %s : %s", profile.Name, err)
+					// Don't fail if we can't compile the filter - for now
+					//	return errors.Wrapf(err, "Error compiling debug filter of %s", profile.Name)
+				}
+				runtime.DebugFilters[fIdx] = debugFilter
+				runtime.Logger.Logger.SetLevel(log.TraceLevel)
```
buixor, line 61: maybe DebugLevel instead and just be sure we log at the appropriate level later on

buixor (changes requested):
minor changes, but starts to look really good :)

buixor (approved):
LGTM

blotus (changes requested):
LGTM, please fix the failing bats test by changing the error message to match.

blotus (approved):

PR #1569, opened 2022-05-31
Fix #1552

### LINKED ISSUE #1552: Bug/Local API: `include_scenarios_containing` works only for the latest decision
**Describe the bug**
Currently, when we have multiple decisions concerning an IP address (with different scenarios), and when we query the local API to get decisions with `scenarios_containing`, it will returns the IP address only if the decisions with the longest duration match the scenario filter.  

**To Reproduce**
1. `cscli decisions add -i 1.2.3.4 --reason test`
2. `cscli decisions add -i 1.2.3.4 --reason test1234`
3. Query the local API with `scenarios_containing=test` and you will not receive decisions from the local API

**Expected behavior**
Getting the decisions whatever the decision duration

 - Version: v1.3.4

AlteredCoder:
- Fix #1552 
- Refactor decisions_test.go for a better lisibility
- Fix `DELETE /decisions` route response (return the real number of items deleted)


pkg/apiserver/decisions_test.go
```diff
@@ -251,181 +257,902 @@ func TestStreamStartDecisionDedup(t *testing.T) {
...
+			TestName:      "check that 127.0.0.1 is not in deleted IP",
+			Method:        "GET",
+			Route:         "/v1/decisions/stream?startup=true",
+			CheckCodeOnly: false,
+			Code:          200,
+			LenNew:        2,
+			LenDeleted:    0,
+			AuthType:      APIKEY,
+			DelChecks:     []DecisionCheck{},
+			NewChecks:     []DecisionCheck{},
```
buixor, line 629: I think we should check that we get the expected decisions (duration + scenario)

buixor:
I think we should add functional tests, it will be useful to check other DB backends.
great work

mmetc (approved):

PR #1595, opened 2022-06-16
enabled linters and fixes for: misspell, predeclared, unconvert, ineffassign, gosimple, govet

mmetc:


cmd/crowdsec/output.go
```diff
@@ -125,8 +125,6 @@ LOOP:
 			if len(cache) > 0 {
 				cacheMutex.Lock()
 				cachecopy := cache
-				newcache := make([]types.RuntimeAlert, 0)
-				cache = newcache
```
buixor, line 129: suppressing this is going to skip "emptying" the cache no ?

mmetc: If I'm not mistaken, we are exiting the function after this block (break LOOP), and cache disappears

buixor (changes requested):
see my comment, but overall lgtm

buixor (approved):
lgtm

Issue #1594, opened 2022-06-15
Improvement/cscli explain


LaurenceJJones:
**Is your feature request related to a problem? Please describe.**
A user may want to parse a log file but only so many number of lines if its a big file.

**Describe the solution you'd like**
--number argument to be added to ```cscli explain``` OR allow cscli explain to handle pipes ```tail -n 50 /var/log/nginx/access.log | cscli explain --type nginx```

Allowing pipes will make it work with head and tail instead of cscli explain needing to offer a `mode` argument

**Describe alternatives you've considered** (Optional) 
outside of cscli ecosystem: tail -n 50 /var/log/nginx/access.log > /tmp/tmp.log && cscli explain --file  /tmp/tmp.log --type nginx && rm  /tmp/tmp.log

LaurenceJJones:
I started working on implementation here https://github.com/LaurenceJJones/crowdsec/tree/cs-explain-stdin

Going to test edge cases as ```-f -```, plus there might be better ways to do it!

PR #1597, opened 2022-06-16
Cscli Explain Stdin

LaurenceJJones:
#1594 
Allow user to provide stdin to ```cscli explain```:
```bash
tail -n 10 /var/log/nginx/access.log | cscli explain --type nginx
head -n 10 /var/log/nginx/access.log | cscli explain --type nginx
cscli explain --type nginx < /var/log/nginx/access.log
```


mmetc:
Thanks

Maybe we can implement reading from stdin without temporary files, by wrapping it in a datasource that is then picked up by pkg/acquisition/modules/file. I don't know if we are following a specific standard for the dsn string

AlteredCoder:
Hello,

To be more compliant with other programs, can we use `-f -` to specify that we read from stdin please?
For example:

```
cat log.file.txt | cscli explain --type nginx -f -

**Review by AlteredCoder — CHANGES_REQUESTED** [review]
Hello,

To be more compliant with other programs, can we use `-f -` to specify that we read from stdin please?
For example:

```
cat logfile.txt | cscli explain --type nginx -f -

cmd/crowdsec-cli/explain.go
```diff
@@ -85,11 +115,11 @@ cscli explain --dsn "file://myfile.log" --type nginx
 				log.Fatalf("fail to run crowdsec for test: %v", err)
 			}
 
-			// rm the temporary log file if only a log line was provided
-			if logLine != "" {
+			// rm the temporary log file if only a log line/stdin was provided
+			if tmpFile != "" {
 				f.Close()
```
LaurenceJJones, line 120: Hey! thanks @buixor ! just wondering if we need to close the file handler twice?

AlteredCoder: yes i think the file is already close before

PR #1598, opened 2022-06-17
Cscli explain use temp dir

LaurenceJJones:
@blotus had a user via discord that ran ```cscli explain``` inside parsers folders which led to crowdsec failing to load as the YAML files left over are invalid format. Best to have all file output go-to temp directory of the host OS, I did add a flag for CWD if a user needs it.

### DIFF
--- a/cmd/crowdsec-cli/explain.go
+++ b/cmd/crowdsec-cli/explain.go
@@ -19,6 +19,7 @@ func NewExplainCmd() *cobra.Command {
 	var logLine string
 	var logType string
 	var opts cstest.DumpOpts
+	var useCWD bool
 
 	var cmdExplain = &cobra.Command{
 		Use:   "explain",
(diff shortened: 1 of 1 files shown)


mmetc:
Thanks for the PR

Just a thing, I don't know if a useCWD option is useful, but please keep flags lowercase.

LaurenceJJones:
I made that change to lowercase the flag. Yeah, the idea was in case the user had no temp directory even though it is mandatory under FHS. Open to removing this if needs be, I would rather give the user a fail safe option if the command is not runnable by using the temp folder as the default.

mmetc:
> I would rather give the user a fail safe option if the command is not runnable by using the temp folder as the default.

From the documentation of os.Tempdir():

```
On Unix systems, it returns $TMPDIR if non-empty, else /tmp. On Windows, it uses GetTempPath, returning the first non-empty value from %TMP%, %TEMP%, %USERPROFILE%, or the Windows directory. On Plan 9, it returns /tmp.

The directory is neither guaranteed to exist nor have accessible permissions.
``` 

So in case of non-existing directory or permission issues, the user can run "TMPDIR=./ cscli explain..."

LaurenceJJones:
Removed cwd flag as like you stated if user has issues we can advised them to set varibles, once this is merged I will add an info section to ```cscli explain``` documentation.

AlteredCoder (approved):

PR #1613, opened 2022-06-24
[wip] serve metrics only after agent and/or lapi are ready; fixed some func tests

mmetc:


buixor:
I like the idea, one of the questions is if we want to put a time-out around https://github.com/crowdsecurity/crowdsec/pull/1613/files#diff-0a396844b06ce082dd3e7bd50d3c0b41adab660bd4d578bb1251fb7360060a76R184 so we can avoid the user hanging if something goes wrong (even I don't really see how this could happen).

mmetc:
> I like the idea, one of the questions is if we want to put a time-out around https://github.com/crowdsecurity/crowdsec/pull/1613/files#diff-0a396844b06ce082dd3e7bd50d3c0b41adab660bd4d578bb1251fb7360060a76R184 so we can avoid the user hanging if something goes wrong (even I don't really see how this could happen).

I'd say no because the sync happens inside the metrics goroutine, it doesn't block anything else. The tests already have a timeout for this case so we would spot it

blotus (approved):

PR #1616, opened 2022-06-28
Properly handle expired decisions with different scopes

blotus:


pkg/database/decisions.go
```diff
@@ -234,37 +238,37 @@ func (c *Client) QueryExpiredDecisionsSinceWithFilters(since time.Time, filters
 	now := time.Now().UTC()
 
 	query := c.Ent.Decision.Query().Where(
-		decision.UntilLT(now),
 		decision.UntilGT(since),
 	)
-	query, predicates, err := BuildDecisionRequestWithFilter(query, filters)
+	query, _, err := BuildDecisionRequestWithFilter(query, filters)
```
buixor, line 243: Please add a comment like:

For "reasons", we're picking up the following approach :
 - we select all the decisions that didn't expire, including those that expired between last pull and now
 - we sort them from the sooner to expire to the one that will expire in the longer time
 - we iterate on the list to ensure that we only return expired decision if not other decision on the same value/scope/type is still valid

buixor (approved):
plz add comms but OK

PR #1628, opened 2022-07-01
Add no-capi flag and review some logs

AlteredCoder:


he2ss (approved):
maybe add a bats test for this new param ?
Otherwise, all seems good :+1:

PR #1634, opened 2022-07-03
add `cscli support dump`

blotus:
This PR add `support` command to `cscli`.

The only subcommand available for now is `dump`. It will create a zip file containing various details about the crowdsec installation and the OS to help diagnose users issues more efficiently.

This collects:
- [x] Crowdsec version
- [x] OS version
- [x] Installed collections list
- [x] Installed parsers list
- [x] Installed scenarios list
- [x] Installed postoverflows list
- [x] Bouncers list
- [x] Machines list
- [x] CAPI status
- [x] LAPI status
- [x] Crowdsec config (sensitive information like username and password are redacted)
- [x] Crowdsec metrics


cmd/crowdsec-cli/bouncers.go
```diff
@@ -54,55 +108,11 @@ Note: This command requires database direct access, so is intended to be run on
 		Args:              cobra.ExactArgs(0),
 		DisableAutoGenTag: true,
 		Run: func(cmd *cobra.Command, arg []string) {
-			blockers, err := dbClient.ListBouncers()
+			bouncers, err := getBouncers(dbClient)
```
buixor, line 111: in csv/human mode, `getBouncers` displays table or write csv, but still returns the bytes it read. It will trigger double display in csv/human ?

mmetc:
Is it possible to avoid calling log.Fatal and just return errors, to check coverage? thanks

cmd/crowdsec-cli/machines.go
```diff
@@ -130,61 +190,11 @@ Note: This command requires database direct access, so is intended to be run on
 			}
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			machines, err := dbClient.ListMachines()
+			agents, err := getAgents(dbClient)
```
buixor, line 193: same as for `getBouncers`

cmd/crowdsec-cli/support.go
```diff
@@ -0,0 +1,393 @@
...
+
+func collectBouncers(dbClient *database.Client) ([]byte, error) {
+	return getBouncers(dbClient)
+}
+
+func collectAgents(dbClient *database.Client) ([]byte, error) {
+	return getAgents(dbClient)
+}
+
+func collectAPIStatus(login string, password string, endpoint string, prefix string) []byte {
```
he2ss, line 141: This func miss the tls support if I'm not mistaken.
We need to check if the user as setup tls and then try to authenticate and collect.

buixor: as seen it works oob because magic

he2ss (approved):

Issue #1624, opened 2022-06-30
Improvement/cscli config test


LaurenceJJones:
**Is your feature request related to a problem? Please describe.**
There is no issue just Crowdsec has the ability to test the configuration before starting, however, a user might not run this as this feature as it is not documented.
```bash
/usr/bin/crowdsec -c /etc/crowdsec/config.yaml -t
```

**Describe the solution you'd like**
Provide a cscli command to invoke this command and provide the input back to the user terminal or document / reinforce the need to run this command before restarting the service? A user may want to be informed they have made a mistake before trying to restart the service meaning Crowdsec will be down for some time whilst they troubleshoot.

**Example of what you imagine**
```bash
cscli config test
cscli -c /tmp/otherconfig.yaml config test
```

LaurenceJJones:
If we provide -t to crowdsec, it uses the configuration for the method of logging (file | stdout). Should we provide an argument EG ```-stdout``` to override this for programmatic purposes?

I was thinking of making -t make stdout the default, however, this may cause confusion when a user goes to restart through systemctl as there be no logs.

mmetc:
Thank you Laurence,

just my 2c re: -stdout. I'm a strong believer of having as many configuration options as possible available through flags, as well as environment variables. However, in crowdsec we don't have a unified mapping between the two, it requires code each time and we decide it case by case. We could have a generic way to do it but it would need to work with two tools that don't use the same flag library (crowdsec and cscli) and it would be hard to do that without breaking compatibility.

Having said that, a "-log-media stdout" flag would be a clear reference to the configuration option it's overriding. It's not extremely useful without -log-dir and friends, but at least it's clear. There's also the incorrect reference to stdout when the logs are actually going to stderr, but it's a minor inconvenience.

LaurenceJJones:
@mmetc I agree that multiple configuration levels allow for "no configuration file" input! Would you like the configurations to be added where applicable for the overall PR or collate all of them into a single PR so they can be merged as one?

mmetc:
I'm not sure what you mean, I think for bin/crowdsec we should keep them to the useful minimum for now. Overriding log_media has its uses, but the other log options may not be worth it. Keep in mind you can also reference environment variables in config.yaml, or (from 1.4) override them with config.yaml.local

PR #1635, opened 2022-07-03
cscli config test

LaurenceJJones:
#1624 Implementation

Looking forward to hearing some feedback on this.


sabban:
Hi @LaurenceJJones,

Sorry  for the delay.

I guess you want to way for cscli to check crowdsec configuration which is a good idea. OTOH this relies on the PATH and the "crowdsec" name to work, which is I guess, not possible to assume in every case.

For this feature I would have used the same code as in crowdsec using the `StartRunSvc` function with the `LintOnly`flag of Crowdsec configuration struct set to true. But it would raised another issue: that we duplicate code between crowdsec and cscli. So I would say this needs more thinking.

@buixor @sblot any more thoughts ?

Thanks for the contribution !

sabban:
Maybe we should document `crowdsec -t` as well

buixor:
As discussed, it might make more sense to simply improve `crowdsec -t` rather than adding some glue around it.
Keeping it opened till we all agree to close it, but removing milestone.

LaurenceJJones:
Conclusion update docs around `crowdsec -t`, install into users before `systemctl restart crowdsec` best practice to run this command and check logs.

PR #1662, opened 2022-07-12
Properly handle service shutdown on windows

blotus:


cmd/crowdsec/serve.go
```diff
@@ -304,10 +304,10 @@ func Serve(cConfig *csconfig.Config, apiReady chan bool, agentReady chan bool) e
 		select {
 		case <-apiTomb.Dead():
 			log.Infof("api shutdown")
-			os.Exit(0)
```
buixor, line 307: let's check this properly on linux / BSD for side-effects

PR #1669, opened 2022-07-20
Generic dateparse approach

buixor:
Currently, `dateparse-enrich` relies on a static, closed list of supported time formats. It makes creating parsers for software using a non-standard format overwhelming.

This approach allows a parser to "suggest" a format string for the date to be parsed.

Missing :
 - [x] tests
 - [x] doc : https://github.com/crowdsecurity/hub/pull/517

### DIFF
--- a/pkg/parser/enrich_date.go
+++ b/pkg/parser/enrich_date.go
@@ -7,6 +7,23 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+func parseDateWithFormat(date, format string) (string, time.Time) {
+	t, err := time.Parse(format, date)
+	if err == nil && !t.IsZero() {
+		//if the year isn't set, set it to current date :)
+		if t.Year() == 0 {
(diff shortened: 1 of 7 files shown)


pkg/parser/enrich_date.go
```diff
@@ -29,40 +46,43 @@ func GenDateParse(date string) (string, time.Time) {
...
 
 	var ret map[string]string = make(map[string]string)
-	tstr, tbin := GenDateParse(in)
-	if !tbin.IsZero() {
-		ret["MarshaledTime"] = tstr
+	var strDate string
+	var parsedDate time.Time
+
+	//log.Printf("received events : %+v", p.Parsed)
+	if v, ok := p.Parsed["DateFormat"]; ok && v != "" {
```
blotus, line 64: I think it would be cleaner to have a `StrTimeFormat` at the root of the event object (and this will be consistent with `StrTime`).

AlteredCoder (approved):

Issue #1668, opened 2022-07-20
logging in the email plugin is buggy


sabban:
### What happened?

The logger name in the email plugin uses a global that is constantly changed when notifying, leading to weird naming in the logs:
```
time="20-07-2022 10:17:37" level=info msg="sent email to [manuel@crowdsec.net]" @module=email-plugin.email_default1.email_default.email_default1.email_default.email_default.email_default1.email_default.email_default1.email_default1.email_default.email_default.email_default1.email_default1.email_default
``` 
This is obtained when using two email plugins at the same time (email_default and email_default1. On top of that, the logger is another library than the one used in CrowdSec. This should be changed.

### What did you expect to happen?

Logging should work with a real plugin type+name

### How can we reproduce it (as minimally and precisely as possible)?

Using two instances of the email plugin at the same time.

### Crowdsec version

<details>

```console
❯ crowdsec -version
2022/07/20 11:12:38 version: v1.4.0-rc5-debian-pragmatic-24b1a99c42f16cb0e937406d90a54f0fbad3e7bd
2022/07/20 11:12:38 Codename: alphaga
2022/07/20 11:12:38 BuildDate: 2022-07-13_10:25:27
2022/07/20 11:12:38 GoVersion: 1.17.5
2022/07/20 11:12:38 Platform: linux
2022/07/20 11:12:38 Constraint_parser: >= 1.0, <= 2.0
2022/07/20 11:12:38 Constraint_scenario: >= 1.0, < 3.0
2022/07/20 11:12:38 Constraint_api: v1
2022/07/20 11:12:38 Constraint_acquis: >= 1.0, < 2.0

```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Enabled collections and parsers

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
#Generated acquisition file - wizard.sh (service: sshd) / files : /var/log/auth.log
filenames:
  - /var/log/auth.log
labels:
  type: syslog
---
#Generated acquisition file - wizard.sh (service: linux) / files : /var/log/syslog /var/log/messages
filenames:
  - /var/log/syslog
  - /var/log/messages
labels:
  ty
# paste output here

# On Windows:
[...]

PR #1670, opened 2022-07-21
fix logging for email plugin

sabban:
fix logging for email plugin (avoid naming issues, when using twice the plugin)

This will fix https://github.com/crowdsecurity/crowdsec/issues/1668


buixor (approved):
LGTM

PR #1686, opened 2022-07-26
add suggestion on cscli install items

AlteredCoder:
```bash
sudo cscli parsers install crowdsecurity/nginx-log
FATA[26-07-2022 11:38:07 PM] unable to find parser 'crowdsecurity/nginx-log', did you mean crowdsecurity/nginx-logs ? 
```

### DIFF
--- a/cmd/crowdsec-cli/collections.go
+++ b/cmd/crowdsec-cli/collections.go
@@ -62,17 +62,7 @@ func NewCollectionsCmd() *cobra.Command {
 				t := cwhub.GetItem(cwhub.COLLECTIONS, name)
 				if t == nil {
 					nearestItem, score := GetDistance(cwhub.COLLECTIONS, name)
-					errMsg := ""
-					if score < MaxDistance {
-						errMsg = fmt.Sprintf("unable to find collections '%s', did you mean %s ?", name, nearestItem.Name)
-					} else {
-						errMsg = fmt.Sprintf("unable to find collections '%s'", name)
-					}
-					if ignoreError {
-						log.Error(errMsg)
-					} else {
-						log.Fatalf(errMsg)
-					}
+					Suggest(cwhub.COLLECTIONS, name, nearestItem.Name, score, ignoreError)
 				}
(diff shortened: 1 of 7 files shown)


LaurenceJJones:
Hey! Great change, the only thing I could think to improve it would be to implement a Yes/No question instead of fatal erroring allows the user to self heal instead of exiting the program.
```go
var answer bool
prompt := &survey.Confirm{
	Message: "Unable to find parser 'crowdsecurity/nginx-log', did you mean crowdsecurity/nginx-logs?",
	Default: true,
}
if err := survey.AskOne(prompt, &answer); err != nil {
	log.Warnf("Unable to ask", err)
}
if !answer {
	log.Error("Unable to find parser 'crowdsecurity/nginx-log'")
        continue
} else {
        name = nearestItem
}
```

cmd/crowdsec-cli/collections.go
```diff
@@ -59,6 +59,21 @@ func NewCollectionsCmd() *cobra.Command {
...
+						errMsg = fmt.Sprintf("unable to find collections '%s', did you mean %s ?", name, nearestItem.Name)
+					} else {
+						errMsg = fmt.Sprintf("unable to find collections '%s'", name)
+					}
+					if ignoreError {
+						log.Error(errMsg)
+					} else {
+						log.Fatalf(errMsg)
+					}
+				}
```
sbs2001, line 76: Can we abstract this in a separate function with item type as a parameter ?  The code duplicated below multiple times.

cmd/crowdsec-cli/utils.go
```diff
@@ -21,9 +21,12 @@ import (
 	"github.com/prometheus/prom2json"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/texttheater/golang-levenshtein/levenshtein"
 	"gopkg.in/yaml.v2"
 )
 
+var MaxDistance = 7
```
sbs2001, line 28: nit: do we want to make this a constant ?

blotus (approved):

PR #1687, opened 2022-07-27
allow user to disable decision deduplication

buixor:
This change was induced after discussion with a user that is using crowdsec for multi-tenancy. I don't see any reason not to merge this ?

### DIFF
--- a/pkg/database/decisions.go
+++ b/pkg/database/decisions.go
@@ -13,6 +13,7 @@ import (
 	"github.com/crowdsecurity/crowdsec/pkg/database/ent/predicate"
 	"github.com/crowdsecurity/crowdsec/pkg/types"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/database/decisions.go
```diff
@@ -97,8 +97,12 @@ func BuildDecisionRequestWithFilter(query *ent.DecisionQuery, filter map[string]
 func (c *Client) QueryAllDecisionsWithFilters(filters map[string][]string) ([]*ent.Decision, error) {
 	query := c.Ent.Decision.Query().Where(
 		decision.UntilGT(time.Now().UTC()),
-		longestDecisionForScopeTypeValue,
 	)
+	//Allow a bouncer to ask for non-deduplicated results
```
he2ss, line 101: typo in the comment right ? (ask for dedup and not "non-dedup") ?

he2ss (approved):

PR #1702, opened 2022-08-04
Don't suggest an item which user already mentioned

sbs2001:
Old behaviour

```
❯ sudo cscli collection install crowdsecurity/apache2 crowdsecurity/
…security/apache2              …security/freebsd         …security/mariadb              …security/odoo          …security/traefik
```

New behaviour

```
❯ sudo cscli collection install crowdsecurity/apache2 crowdsecurity/
…security/asterisk             …security/haproxy         …security/modsecurity          …security/opnsense      …security/vsftpd
```

### DIFF
--- a/cmd/crowdsec-cli/utils.go
+++ b/cmd/crowdsec-cli/utils.go
@@ -112,12 +112,8 @@ func compAllItems(itemType string, args []string, toComplete string) ([]string, 
 	comp := make([]string, 0)
 	hubItems := cwhub.GetHubStatusForItemType(itemType, "", true)
 	for _, item := range hubItems {
-		if toComplete == "" {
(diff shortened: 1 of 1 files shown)


cmd/crowdsec-cli/utils.go
```diff
@@ -112,12 +112,8 @@ func compAllItems(itemType string, args []string, toComplete string) ([]string,
 	comp := make([]string, 0)
 	hubItems := cwhub.GetHubStatusForItemType(itemType, "", true)
 	for _, item := range hubItems {
-		if toComplete == "" {
```
sbs2001, line 115: strings.Contains(item.Name, toComplete) is always true if toComplete is empty string, so we can eliminate this condition branch

buixor (approved):

buixor (approved):

PR #1708, opened 2022-08-08
Blackhole new behaviour and add support for WAL when using sqlite

blotus:
- Allow to enable WAL when using sqlite
 - Change `blackhole` behaviour: previously, a bucket would have been created even if the overflow would be discarded in the end. This could lead to a lot of useless buckets being created and tracked.
 
The new behaviour prevents the bucket creation if the scenario is blackholed.
 
This slightly change the meaning of blackhole: before, a new bucket would be created, events poured to it, if the blackhole expired before it was full enough to overflow, it would be allowed to overflow normally.
Now, the bucket will not be created in the first place, which means that some overflows that would have happened with the old behaviour will not happen now (this is likely not an issue in real life). 

* [ ] tests needed


ornicar:
We (lichess.org) are running this in production right now. I'll report if anything unexpected happens.

buixor:
let's make blackhole cleanup rely on latest event's timestamp

buixor (changes requested):
let's make blackhole cleanup rely on latest event's timestamp

cmd/crowdsec/crowdsec.go
```diff
@@ -67,6 +70,15 @@ func runCrowdsec(cConfig *csconfig.Config, parsers *parser.Parsers) error {
 	parserWg.Wait()
 
 	bucketWg := &sync.WaitGroup{}
+	log.Infof("BucketsGCEnabled: %v", cConfig.Crowdsec.BucketsGCEnabled)
+	if !cConfig.Crowdsec.BucketsGCEnabled {
```
buixor, line 74: it's a bit counter-intuitive, might be worth a comment ;)

buixor (approved):
lgtm

sabban:
Test needed to be merged.

PR #1715, opened 2022-08-18
Polling API Integration

buixor:
Add an opt-in feature: support for polling API from LAPI to CAPI.

This allows real-time communication from CAPI to the agent : 
 - short term :  viewing / adding / removing decisions from the console
 - medium/long term : deploying scenarios/parsers/whitelists from console


pkg/apiserver/apic.go
```diff
@@ -400,6 +579,104 @@ func fillAlertsWithDecisions(alerts []*models.Alert, decisions []*models.Decisio
...
+
+	log.Infof("Starting PAPI pull (since:%s)", lastTimestamp)
+	for event := range c.Start(lastTimestamp) {
+		//update last timestamp in database
+		newTime := time.Now().UTC()
+		binTime, err := newTime.MarshalText()
+		if err != nil {
+			return errors.Wrap(err, "failed to marshal last timestamp")
+		}
+		if err := a.dbClient.SetConfigItem(PAPI_PULL_KEY, string(binTime)); err != nil {
```
AlteredCoder, line 636: maybe we want to do this after we sent the alert in the channel ?

pkg/apiserver/controllers/v1/decisions.go
```diff
@@ -93,11 +93,17 @@ func (c *Controller) DeleteDecisionById(gctx *gin.Context) {
...
 	if err != nil {
 		c.HandleDBErrors(gctx, err)
 		return
 	}
+	//transform deleted decisions to be sendable to capi
+	deletedDecisions, err := FormatDecisions(deletedFromDB, false)
+	if err != nil {
+		log.Warningf("failed to format decisions: %v", err)
+	}
+	c.DecisionDeleteChan <- deletedDecisions
```
AlteredCoder, line 106: do we want to wait for this operation since the channel is not buffered

pkg/apiserver/controllers/v1/decisions.go
```diff
@@ -109,11 +115,18 @@ func (c *Controller) DeleteDecisionById(gctx *gin.Context) {
...
```
AlteredCoder, line 128: do we want to wait for this operation since the channel is not buffered

pkg/csconfig/console.go
```diff
@@ -15,16 +15,18 @@ const (
 	SEND_CUSTOM_SCENARIOS  = "custom"
 	SEND_TAINTED_SCENARIOS = "tainted"
 	SEND_MANUAL_SCENARIOS  = "manual"
+	RECEIVE_DECISIONS      = "decisions_poll"
```
AlteredCoder, line 18: doesn't `decisions_sync` make more sense ?

pkg/csconfig/console.go
```diff
@@ -15,16 +15,18 @@ const (
...
-var CONSOLE_CONFIGS = []string{SEND_CUSTOM_SCENARIOS, SEND_MANUAL_SCENARIOS, SEND_TAINTED_SCENARIOS}
+var CONSOLE_CONFIGS = []string{SEND_CUSTOM_SCENARIOS, SEND_MANUAL_SCENARIOS, SEND_TAINTED_SCENARIOS, RECEIVE_DECISIONS}
 
 var DefaultConsoleConfigFilePath = DefaultConfigPath("console.yaml")
 
 type ConsoleConfig struct {
 	ShareManualDecisions  *bool `yaml:"share_manual_decisions"`
 	ShareTaintedScenarios *bool `yaml:"share_tainted"`
 	ShareCustomScenarios  *bool `yaml:"share_custom"`
+	ReceiveDecisions      *bool `yaml:"receive_decisions"`
```
AlteredCoder, line 29: same here: `sync_decisions` ?

pkg/csconfig/console.go
```diff
@@ -34,6 +36,7 @@ func (c *LocalApiServerCfg) LoadConsoleConfig() error {
 		c.ConsoleConfig.ShareCustomScenarios = types.BoolPtr(true)
 		c.ConsoleConfig.ShareTaintedScenarios = types.BoolPtr(true)
 		c.ConsoleConfig.ShareManualDecisions = types.BoolPtr(false)
+		c.ConsoleConfig.ReceiveDecisions = types.BoolPtr(false)
```
AlteredCoder, line 39: is this should be enabled by default ?

pkg/database/decisions.go
```diff
@@ -361,32 +366,33 @@ func (c *Client) DeleteDecisionsWithFilter(filter map[string][]string) (string,
...
 
-	nbDeleted, err := decisions.Exec(c.CTX)
+	toDelete, err := decisions.All(c.CTX)
 	if err != nil {
 		c.Log.Warningf("DeleteDecisionsWithFilter : %s", err)
-		return "0", errors.Wrap(DeleteFail, "decisions with provided filter")
+		return "0", nil, errors.Wrap(DeleteFail, "decisions with provided filter")
 	}
-	return strconv.Itoa(nbDeleted), nil
+	count, err := c.BulkDeleteDecisions(toDelete, false)
```
AlteredCoder, line 377: check err

pkg/database/ent/schema/decision.go
```diff
@@ -37,6 +37,7 @@ func (Decision) Fields() []ent.Field {
 		field.String("value"),
 		field.String("origin"),
 		field.Bool("simulated").Default(false),
+		field.String("uuid").Optional(), //this uuid is mostly here to ensure that CAPI/PAPI has a unique id for each decision
```
AlteredCoder, line 40: do we want an index on this field ?

pkg/database/decisions.go
```diff
@@ -474,28 +480,86 @@ func (c *Client) SoftDeleteDecisionsWithFilter(filter map[string][]string) (stri
...
-func (c *Client) SoftDeleteDecisionByID(decisionID int) (int, error) {
-	nbUpdated, err := c.Ent.Decision.Update().Where(decision.IDEQ(decisionID)).SetUntil(time.Now().UTC()).Save(c.CTX)
-	if err != nil || nbUpdated == 0 {
-		c.Log.Warningf("SoftDeleteDecisionByID : %v (nb soft deleted: %d)", err, nbUpdated)
-		return 0, errors.Wrapf(DeleteFail, "decision with id '%d' doesn't exist", decisionID)
+func (c *Client) SoftDeleteDecisionByID(decisionID int) (int, []*ent.Decision, error) {
+	toUpdate, err := c.Ent.Decision.Query().Where(decision.IDEQ(decisionID)).All(c.CTX)
+
+	if err != nil || len(toUpdate) == 0 {
+		c.Log.Warningf("SoftDeleteDecisionByID : %v (nb soft deleted: %d)", err, toUpdate)
```
AlteredCoder, line 554: use `len(toUpdate`) in the args

pkg/types/constants.go
```diff
@@ -3,3 +3,6 @@ package types
 const ApiKeyAuthType = "api-key"
 const TlsAuthType = "tls"
 const PasswordAuthType = "password"
+
+const PAPIBaseURL = "https://papi.dev.crowdsec.net/v1/decisions/stream/poll"
```
AlteredCoder, line 7: should we handle the URL with the version as we do for CAPI ?

pkg/apiserver/apic.go
```diff
@@ -118,13 +164,111 @@ func NewAPIC(config *csconfig.OnlineApiClientCfg, dbClient *database.Client, con
...
+				a.mu.Lock()
+				cacheCopy := cache
+				cache = make([]*models.AddSignalsRequestItemDecisionsItem, 0)
+				a.mu.Unlock()
+				log.Infof("Signal push: %d signals to push", len(cacheCopy))
+				go a.SendDeletedDecisions(&cacheCopy)
+			}
+		case deletedDecisions := <-a.DecisionDeleteChan:
+
+			if a.consoleConfig.ShareManualDecisions != nil && *a.consoleConfig.ShareManualDecisions {
```
AlteredCoder, line 205: We need to be careful with this flag and the check overall.

`ShareManualDecisions` configuration option was initially used to send the decisions that a user added manually (`cscli decisions add ..`).
Since we have now the `decisions_poll` configuration option, a user might want to send deleted decisions (to have the synchronization in the console) but not the `decisions` that he added manually.

AlteredCoder (changes requested):

pkg/apiserver/controllers/v1/decisions.go
buixor: On the other end of the chan, we have a routine dedicated to polling the chan and building a cache `SyncDecisions`.
We can throw this in a tombless go-routine maybe, wdyt ?

pkg/apiserver/controllers/v1/decisions.go
buixor: same as above

pkg/csconfig/console.go
buixor: yes this was my first thought too. I'm afraid it will lead to confusion given the current behaviour of "send_manual_scenaruios" (that sends "manual_decisions"). We can change it, but let's keep in mind backward compat. What do you suggest ?

pkg/csconfig/console.go
buixor: solve this once we agreed with the above

pkg/csconfig/console.go
buixor: I don't think so, especially it communicates w/ a new endpoint. I think we might want to have it opt-in for now.

AlteredCoder:
Note:

There is a bug with `cscli alerts list` :
```
sudo cscli alerts list --all
No active alerts
```
While I have at least an alert for CAPI and one for PAPI

buixor:
- [ ] Do not add papi url by default to online_credentials as it's opt-in
- [x] Console status doesn't display the new flag (decisions receive)
- [ ] Ensure `cscli decisions` list correctly display decisions from PAPI
- [ ] Ensure `cscli alerts list` correctly displays alerts from PAPI


```
#ok
./cscli -c dev.yaml alerts list --origin console

#not ok 
 ./cscli -c dev.yaml alerts list -a
```

cmd/crowdsec-cli/alerts.go
```diff
@@ -108,6 +153,7 @@ func DisplayOneAlert(alert *models.Alert, withDetail bool) error {
 		fmt.Printf(" - AS         : %s\n", alert.Source.AsName)
 		fmt.Printf(" - Begin      : %s\n", *alert.StartAt)
 		fmt.Printf(" - End        : %s\n\n", *alert.StopAt)
+		fmt.Printf(" - UUID	      : %s\n", alert.UUID)
```
buixor, line 156: maybe we should align tabs here

pkg/csconfig/api.go
```diff
@@ -211,8 +214,25 @@ func (c *Config) LoadAPIServer() error {
 		log.Warning("crowdsec local API is disabled from flag")
 	}
 
-	if c.API.Server == nil {
-		log.Warning("crowdsec local API is disabled because its configuration is not present")
+	if c.API.Server != nil {
+		logLevel := log.InfoLevel
```
buixor, line 218: shouldn't this be inherited from `api->server->log_level` or at least `common->log_level` ?

pkg/database/alerts.go
```diff
@@ -661,7 +828,11 @@ func AlertPredicatesFromFilter(filter map[string][]string) ([]predicate.Alert, e
 			predicates = append(predicates, alert.HasDecisionsWith(decision.OriginEQ(value[0])))
 		case "include_capi": //allows to exclude one or more specific origins
 			if value[0] == "false" {
-				predicates = append(predicates, alert.HasDecisionsWith(decision.Or(decision.OriginEQ("crowdsec"), decision.OriginEQ("cscli"))))
+				predicates = append(predicates, alert.HasDecisionsWith(
```
buixor, line 831: at which point are we going to switch to a "not" ? :sweat_smile:  (but probably out of scope for now)

AlteredCoder (approved):

AlteredCoder (approved):

PR #1728, opened 2022-09-01
[wip] cscli setup / wizard.sh

mmetc:


cmd/crowdsec-cli/setup.go
```diff
@@ -0,0 +1,312 @@
...
+		flags.String("detect-config", defaultServiceDetect, "path to service detection configuration")
+		flags.Bool("list-supported-services", false, "do not detect; only print supported services")
+		flags.StringSlice("force-unit", nil, "force detection of a systemd unit (can be repeated)")
+		flags.StringSlice("force-process", nil, "force detection of a running process (can be repeated)")
+		flags.StringSlice("skip-service", nil, "ignore a service, don't recommend hub/datasources (can be repeated)")
+		flags.String("force-os-family", "", "override OS.Family: one of linux, freebsd, windows or darwin")
+		flags.String("force-os-id", "", "override OS.ID=[debian | ubuntu | , redhat...]")
+		flags.String("force-os-version", "", "override OS.RawVersion (of OS or Linux distribution)")
+		flags.Bool("snub-systemd", false, "don't use systemd, even if available")
+		flags.Bool("yaml", false, "output yaml, not json")
```
LaurenceJJones, line 51: Any reason we would default to json rather than yaml?

LaurenceJJones: Answered my own question once I saw bats-tests

mmetc: I don't have strong opinions, but here's the rationale. JSON is easier to copy/paste (no corruption due to indentation) and programmatically manipulate or validate in case of IaC automation. In the other cases, it's an intermediary file that does not get stored anywhere, it doesn't need to be readable or to have comments. If it does (as in wizard.sh) there's the yaml option. As a rule of thumb, for machine-to-machine communication: json; machine-human-machine: yaml

sbs2001:
Is the old wizard broken somehow ?
On a fresh instance I get:
```
ubuntu@primary:~/Home/work/crowdsec/crowdsec-v1.4.2-rc4$ sudo ./wizard.sh --install --unattended 
INFO[10/28/22:19:08:29] crowdsec_wizard: checking if crowdsec is installed
INFO[10/28/22:19:08:29] crowdsec_wizard: installing crowdsec
INFO[28-10-2022 07:08:31 PM] Wrote new 500185 bytes index to /etc/crowdsec/hub/.index.json 
WARN[28-10-2022 07:08:31 PM] can't load CAPI credentials from '/etc/crowdsec/online_api_credentials.yaml' (missing field) 
INFO[28-10-2022 07:08:31 PM] push and pull to Central API disabled        
WARN[0000] you are using sqlite without WAL, this can have an impact of performance. If you do not store the database in a network share, set db_config.use_wal to true. Set explicitly to false to disable this warning.  context=ent
INFO[28-10-2022 07:08:31 PM] Machine '9ebc77f81364448c9d54d53a95ac7148' successfully added to the local API 
INFO[28-10-2022 07:08:31 PM] API credentials dumped to '/etc/crowdsec/local_api_credentials.yaml' 
WARN[28-10-2022 07:08:31 PM] can't load CAPI credentials from '/etc/crowdsec/online_api_credentials.yaml' (missing field) 
INFO[28-10-2022 07:08:31 PM] push and pull to Central API disabled        
INFO[28-10-2022 07:08:33 PM] Successfully registered to Central API (CAPI) 
INFO[28-10-2022 07:08:33 PM] Central API credentials dumped to '/etc/crowdsec/online_api_credentials.yaml' 
WARN[28-10-2022 07:08:33 PM] Run 'sudo systemctl reload crowdsec' for the new configuration to be effective. 

FATA[28-10-2022 07:08:33 PM] detecting services: invalid datasource for gitea-systemd: source is empty 
Installing hub objects....
FATA[28-10-2022 07:08:33 PM] EOF 
```

And after this I did:
```
ubuntu@primary:~/Home/work/crowdsec/crowdsec-v1.4.2-rc4$ sudo cscli setup detect 
FATA[28-10-2022 07:22:03 PM] detecting services: invalid datasource for dovecot-systemd: source is empty 
ubuntu@primary:~/Home/work/crowdsec/crowdsec-v1.4.2-rc4$ sudo cscli setup detect 
FATA[28-10-2022 07:22:04 PM] detecting services: invalid datasource for asterisk-systemd: source is empty 
ubuntu@primary:~/Home/work/crowdsec/crowdsec-v1.4.2-rc4$ sudo cscli setup detect 
FATA[28-10-2022 07:22:05 PM] detecting services: invalid datasource for suricata-systemd: source is empty
```

Looks like the default detect.yaml is missing some config field which has been made mandatory by the code.

config/detect.yaml
```diff
@@ -0,0 +1,456 @@
...
+  #
+
+  lemonldap-ng-systemd:
+    when:
+      - UnitFound("lemonldap-ng-fastcgi-server.service")
+    install:
+      collections:
+        - firewallservices/lemonldap-ng
+    datasource:
+      # XXX todo where are the logs?
```
sbs2001, line 167: We need to find this, as this won't pass the validator if file path is missing.

From https://lemonldap-ng.org/documentation/2.0/logs.html#default-loggers  it seems it uses same location as Apache ? 

Do you think maybe we could make the validator pass in such situations while emitting a warning ?

mmetc: The datasource of each application must be verified manually before going in production, since I did some guesswork. In the bats-detect directory you find the tests to install on .deb, .rpm platforms (I deploy them in ansible fist) and check the detection worked correctly, but logs are not automatically verified.
If we are not sure about a datasource, we can just skip the whole section. I'm not sure warnings are useful with apt/dnf, but we could collect them with something like:

```
 - when: blabla
   install:
      collections: bla/bla
   message: |
      The collection bla/bla was installed, but it has no datasource. Enjoy!
```

Then it would be printed together at the end.

We could even use the mechanism to have parts with when/message and nothing else. Could be useful.
Or maybe allow an empty datasource if there's a comment: field, which would go in its acquis.d/*.yaml file

PR #1732, opened 2022-09-05
Fix static build

blotus:
We were not specifying the proper build tags when building crowdsec statically, which meant a static crowdsec was still trying to open the libc and this led to errors when running on  systems with a libc older than the one used on the build machine.

This PR makes sure we use the Go version of `os.user` (this means that crowdsec will not able to perform lookup against LDAP for example when checking the plugins permissions when built statically, but this is most likely a very rare usecase)  and `net` (see https://pkg.go.dev/net#hdr-Name_Resolution for the impact) to avoid any dependencies of the libc when building statically.
It also disable SQLite extension support to remove compile time warnings (but this was not used anyway)

### DIFF
--- a/.github/workflows/bats-mysql-static.yml
+++ b/.github/workflows/bats-mysql-static.yml
@@ -1,0 +1,96 @@
+name: Functional tests (MySQL - Static Build)
+
+on:
+  workflow_call:
+    inputs:
+      database_image:
+        required: true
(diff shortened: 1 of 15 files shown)


blotus:
Replaced by https://github.com/crowdsecurity/crowdsec/pull/1744

PR #1740, opened 2022-09-07
Abuseipdb notification plugin

katzeprior:
I created a notification plugin that sends the alerts to abuseipdb with the imo correct tags and a nice message. [here](https://www.abuseipdb.com/user/91534) is my abuseipdb account that has been collecting alerts for a week now, and still no bugs, only error would be if an IP has been reported in the last 15 minutes.

Thanks @LaurenceJJones for brainstorming different ideas.


itsvrl:
This needs to be looks at. This is great!

mathieuHa:
This looks like a must to have in Crowdsec.

Could you take a look @LaurenceJJones @mmetc ?

Looking forward for this notification plugin

LaurenceJJones:
i am interested in this plugin, however, the score system needs some work. Cant have a static list cause it wont future proof it. So I have on my list to create an abuseipdb enricher, so inside the meta will be the score for people to use in scenarios. However I also want this enricher to take the scenario and try to classify using abuseip numbers so then inside the plugin we can get the detail. (Havent worked the last bit out yes as alert object doesnt store that info) but will sort some thing out

PR #1741, opened 2022-09-08
Replace shodan with crowdsec cti

sbs2001:


LaurenceJJones:
Just an extra FYI if the user is not pre authenticated before clicking the link they are redirect to
```
https://app.crowdsec.net/login?next=/cti/[ip]
```
Which loses the IP then they have to click the link again after authenticating.

LaurenceJJones:
I can confirm the redirection is working, cheers, this will fix my cti shortcuts

buixor (approved):

PR #1759, opened 2022-09-18
Fix misspelling of instantiate participles

xconverge:
Just a mechanical find and replace with case preserved spelling fix.


buixor:
thanks for your contribution :)

he2ss (approved):
:+1:

PR #1763, opened 2022-09-23
new tables

mmetc:
This depends on https://github.com/crowdsecurity/crowdsec/pull/1762

I kept the styling as similar as possible to the old ones, the terminal width is only detected when the io.Writer which is passed to the constructor is os.Stdout - I have modified support.go accordingly. Wrapping is not perfect, but similar to the old tables.


buixor (approved):
lgtm

Issue #1315, opened 2022-03-04
Improvement/Add a forward_dns method for associated parser/enrichment plugin 


buixor:
cf. https://github.com/crowdsecurity/hub/issues/405

LaurenceJJones:
@buixor since #1775 this can be closed on release?

LaurenceJJones:
Classing as closed completed, reopen if you dont feel the fix is appropriate

PR #1775, opened 2022-09-30
Add LookupHost expr lib func

ThinkChaos:
Makes it possible to allow-list a domain whose IPs are not known in advance.
A practical is example is whitelisting a home IP with a dynamic DNS setup without having to resort to allowing the whole country or ISP.

Example:
```yaml
name: me/my_cool_whitelist
description: lets whitelist our own IP
whitelist:
  reason: dont ban my IP
  expression:
    - evt.Overflow.Alert.Source.IP in LookupHost('my-home-domain.invalid')
```

This is my solution to #1315 and https://github.com/crowdsecurity/hub/issues/405. I'll let you decide if it fixes those requests, but my use case is the exact same so I think it should suffice.  
I didn't implement it via a enrichment because it's easier this way: for this use case the param to the function is constant, so it doesn't need to be via enrich. Also enrich functions can only return a string (and no error).


buixor:
Hello,

I like it, I'll keep you posted soon.

LaurenceJJones (approved):
Will add some documentation to docs.crowdsec.net to advise this should only be used in postoverflows as its very slow.

PR #1782, opened 2022-10-04
Bulk delete alert optimization

mmetc:


mmetc:
I don't see a way to improve error messages since I cannot list/log all the IDs and they messages are not returned as-is by the API anyway.

buixor (approved):

Issue #1794, opened 2022-10-07
TLS is forced even when -e USE_TLS="false"


stephdl:
### What happened?

I use crowdsec with podman and at the start of the container I use the environment var `-e USE_TLS="false"` however the TLS is forced inside the config.yaml. The documentation states we can use true or false

https://hub.docker.com/r/crowdsecurity/crowdsec

`USE_TLS - Enable TLS on the API Server (default: false) : -e USE_TLS="<true|false>"`

### What did you expect to happen?

I expect at the start of the container to use either true or false to enable or disable TLS between agent and local api

when I read the code [ docker_start.sh](https://github.com/crowdsecurity/crowdsec/blob/master/docker/docker_start.sh#L93) I can see that you only look after if the var is initialized, you should look after true or false

### How can we reproduce it (as minimally and precisely as possible)?
```
/usr/bin/podman run \
    --detach \
    --replace --name=crowdsec \
    --publish 127.0.0.1:8080:8080 \
    --publish 6060:6060 \
    --volume /run/log/journal/:/run/log/journal:Z \
   --env USE_TLS="false" \
    docker.io/crowdsecurity/crowdsec:v1.4.1-debian
```

### Crowdsec version

<details>

```
root@8f0160795aa2:/# cscli version
2022/10/07 07:42:01 version: v1.4.1-e1954adc325baa9e3420c324caabd50b7074dd77
2022/10/07 07:42:01 Codename: alphaga
2022/10/07 07:42:01 BuildDate: 2022-07-25_09:26:12
2022/10/07 07:42:01 GoVersion: 1.17.12
2022/10/07 07:42:01 Platform: linux
2022/10/07 07:42:01 Constraint_parser: >= 1.0, <= 2.0
2022/10/07 07:42:01 Constraint_scenario: >= 1.0, < 3.0
2022/10/07 07:42:01 Constraint_api: v1
2022/10/07 07:42:01 Constraint_acquis: >= 1.0, < 2.0
```

</details>

### OS version

<details>
```
[root@ns8loc2 ~]#  cat /etc/os-release
NAME="CentOS Stream"
VERSION="9"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
PRETTY_NAME="CentOS Stream 9"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:centos:9"
[...]

PR #1795, opened 2022-10-07
Check is TLS == "true" before to enable in LAPI

stephdl:
We need to check if TLS is true before to enable in LAPI, see https://github.com/crowdsecurity/crowdsec/issues/1794

### DIFF
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -90,7 +90,7 @@ if [ "$GID" != "" ]; then
     fi
 fi
 
-if [ "$USE_TLS" != "" ]; then
+if [ "${USE_TLS,,}" == "true" ]; then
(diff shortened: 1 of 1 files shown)


mmetc:
Thanks!

I suggest using parameter expansion: "${USE_TLS,,}"  which converts the variable to lowercase, and works with True|False

mmetc:
Almost, it's missing the brackets ->  "${USE_TLS,,}"

stephdl:
thank for your patience @mmetc

mmetc:
Thank you for your contribution :)

mmetc (approved):

stephdl:
newby zone, what I need to merge the PR ?

mmetc:
After submitting a PR, it has to be reviewed (done), tested (running now) and then it can be merged anytime. All three operations are done by us. Once reviewed, you don't need to do anything else.

PR #1797, opened 2022-10-07
unit tests: always capture testcase variable -> allow parallel testing

mmetc:
well explained here https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

I am not using Parallel() now, we could and it's a good habit to capture the variable just in case.


sabban (approved):
lgtm

PR #1806, opened 2022-10-13
enable all static checks + minor fixes and typos

mmetc:
- explicitly ignore returned parameters
 - replace Walk with faster WalkDir
 - log path error during hub dir sync
 - colorize static unit tests
 - removed duplicate import in crowdsec/main.go
 - typos
 - func tests: default datasource in tests/var/log instead of /tmp
 - action setup-go v3


sabban (approved):
lgtm

PR #1811, opened 2022-10-15
rename pkg/cstest -> pkg/hubtest

mmetc:
keep pkg/cstest for generic helper functions
this also avoids circular imports in test files


sabban (approved):
lgtm

PR #1813, opened 2022-10-16
force postgres 14 for func tests

mmetc:
the postgres 15 image (released two days ago) has a permission issue. proper fix will follow


sabban (approved):
lgtm

PR #1815, opened 2022-10-17
refactor broker_test.go, extract cstest/filenotfound*.go

mmetc:
explicit error comparison, use helpers, reduce line count


sabban (approved):
lgtm

PR #1816, opened 2022-10-17
warn if no acquisition files are found, acquisition_test refactoring, func tests

mmetc:


cmd/crowdsec/main.go
```diff
@@ -153,7 +153,7 @@ func LoadAcquisition(cConfig *csconfig.Config) error {
 	} else {
 		dataSources, err = acquisition.LoadAcquisitionFromFile(cConfig.Crowdsec)
 		if err != nil {
-			return errors.Wrap(err, "while loading acquisition configuration")
```
sabban, line 156: any reason why you removed the Wrap here ? It feels like we loose information ? But maybe I missed sth ?

mmetc: Yes, there was already TMI :)

sabban (approved):

PR #1817, opened 2022-10-17
Hub auto update cronjob

LaurenceJJones:


mmetc:
This is what I do in opnsense:

https://github.com/opnsense/plugins/blob/master/security/crowdsec/src/etc/cron.d/oscrowdsec.cron

https://github.com/opnsense/plugins/blob/master/security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh

LaurenceJJones:
FYI my rm modifications dont seem to be working?

PR #1824, opened 2022-10-19
fix for collections delete

mmetc:


tests/bats/20_collections.bats
```diff
@@ -58,15 +58,49 @@ teardown() {
...
-@test "cannot remove a collection twice" {
-    run -0 cscli collections install crowdsecurity/mysql -o human
-    run -0 --separate-stderr cscli collections remove crowdsecurity/mysql
-    run -1 --separate-stderr cscli collections remove crowdsecurity/mysql -o json
-    run -0 jq -r '.level' <(stderr)
-    assert_output 'fatal'
-    run -0 jq -r '.msg' <(stderr)
-    assert_output --partial "unable to disable crowdsecurity/mysql"
-    assert_output --partial "doesn't exist"
+@test "cannot remove a collection that belongs to another" {
```
buixor, line 61: can you check `--force` works tho ?

mmetc:
> can we check that `--force` will work ? and if so, what happens if you upgrade another collection that depends on the removed one ?

1) is done in the last commit, what do you mean by 2) ? I remove sshd and then upgrade linux? it doesn't remove the taint if that's what you mean, must use install --force

tests/bats/20_collections.bats
mmetc: yes, as I said in the comment below. anything else?

buixor (approved):

PR #1831, opened 2022-10-21
make papi url configurable

he2ss:


pkg/apiserver/apic.go
```diff
@@ -147,6 +147,10 @@ func NewAPIC(config *csconfig.OnlineApiClientCfg, dbClient *database.Client, con
 	if err != nil {
 		return nil, errors.Wrapf(err, "while parsing '%s'", config.Credentials.URL)
 	}
+	papiURL, err := url.Parse(config.Credentials.PapiURL)
+	if err != nil {
+		return nil, errors.Wrapf(err, "while parsing '%s'", config.Credentials.PapiURL)
```
mmetc, line 152: please use fmt.Errorf with %w if you want to wrap, %v if you don't. errors.Wrapf is deprecated and archived by the author

PR #1835, opened 2022-10-24
Use explicit transaction when inserting community blocklist

blotus:
Create explicitly a transaction when inserting data we received from CAPI.
This greatly speeds up the insertion of the community blocklist (during testing, takes under 5s to insert 80k IPs vs 4 minutes previously).
It also helps with concurrent access to the database when using MySQL (no more locks when trying to query the decisions stream while the community blocklist is updated)


buixor (approved):
lgtm

PR #1838, opened 2022-10-24
fix(ci): create hub badges, repository dispatch only on crowdsecurity/crowdsec

mmetc:
This avoids workflow errors when working in a fork.


sabban (approved):
lgtm

PR #1843, opened 2022-10-26
add `cscli alerts delete --id`

buixor:
was added in pkg/apiclient and pkg/apiserver, but wasn't available via cmd line flags :man_shrugging:


blotus (approved):

mmetc:
It would be nice to consistently use the RunE method and return error instead of log.Fatal. This allows bincover to correctly register the coverage. Bonus points for func test, but I can do that

PR #1845, opened 2022-10-27
notify when community-blocklist starts pull

buixor:
minor change to notify blocklist pull update, will make eventual troubleshooting easier

### DIFF
--- a/pkg/apiserver/apic.go
+++ b/pkg/apiserver/apic.go
@@ -431,7 +431,7 @@ func (a *apic) PullTop() error {
 		return nil
 	}
 
-	log.Printf("Starting community-blocklist update")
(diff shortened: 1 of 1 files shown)


AlteredCoder (approved):

pkg/apiserver/apic.go
```diff
@@ -431,6 +431,8 @@ func (a *apic) PullTop() error {
 		return nil
 	}
 
+	log.Printf("Starting community-blocklist update")
```
mmetc, line 434: If it's the same as Infof, please use it. Otherwise there can be confusion between stdout/stderr

mmetc (approved):

PR #1846, opened 2022-10-27
randomize pull, push and metric intervals; reload crowdsec only when hub changed

mmetc:
Actually I think d +/- delta would be better, instead of delta/2

### DIFF
--- a/pkg/cwhub/helpers.go
+++ b/pkg/cwhub/helpers.go
@@ -197,6 +197,7 @@ func UpgradeConfig(csConfig *csconfig.Config, itemType string, name string, forc
 				log.Infof("%v %s is local", emoji.Prohibited, v.Name)
 			}
 		} else {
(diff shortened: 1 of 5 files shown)


pkg/cwhub/helpers.go
```diff
@@ -197,6 +197,9 @@ func UpgradeConfig(csConfig *csconfig.Config, itemType string, name string, forc
 				log.Infof("%v %s is local", emoji.Prohibited, v.Name)
 			}
 		} else {
+			// this is used while scripting to know if the hub has been upgraded
+			// and a configuration reload is required
+			fmt.Printf("updated %s\n", v.Name)
```
LaurenceJJones, line 202: If user ran update manually wouldn't they see it twice?

buixor (approved):
lgtm

PR #1855, opened 2022-11-03
fix docker_start without using jq

AlteredCoder:


docker/docker_start.sh
```diff
@@ -133,7 +133,7 @@ if [ "$DISABLE_POSTOVERFLOWS" != "" ]; then
 fi
 
 function register_bouncer {
-  if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o json | jq -r .[].name | grep -q "${NAME}"; then
+  if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o raw | cut -d "," -f1 | grep -q "${NAME}"; then
```
mmetc, line 136: If you want to support names with comma or space:
cscli bouncers list -o json | sed '/^ *"name"/!d;s/^ *"name": "\(.*\)",/\1/'

then apply grep -q
tested with busybox too

docker/docker_start.sh
```diff
@@ -133,7 +133,7 @@ if [ "$DISABLE_POSTOVERFLOWS" != "" ]; then
 fi
 
 function register_bouncer {
-  if ! cscli -c "$CS_CONFIG_FILE" bouncers list -o json | jq -r .[].name | grep -q "${NAME}"; then
+  if ! cscli bouncers list -o json | sed '/^ *"name"/!d;s/^ *"name": "\(.*\)",/\1/' | grep -q "${NAME}"; then
```
LaurenceJJones, line 136: -c "$CS_CONFIG_FILE" no longer need this?

mmetc:
I pushed a commit because you could have two bouncers named "foo" and "foo-something", you don't want a partial match.
Otherwise lgtm

mmetc (approved):

LaurenceJJones (approved):
lgtm

PR #1858, opened 2022-11-04
fix ticker in bucket

sabban:
Fix ticker in bucket


LaurenceJJones (approved):
LGTM under high load, no leaks

PR #1861, opened 2022-11-07
docker: avoid duplicating geoip database

mmetc:
this doesn't need changes for k8s or homeassistant, still saves 77mb (verify with "docker system df")

### DIFF
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -13,11 +13,17 @@ KEY_FILE="${KEY_FILE:-/etc/ssl/key.pem}"
 # Plugins directory default
 PLUGIN_DIR="${PLUGIN_DIR:-/usr/local/lib/crowdsec/plugins/}"
 
-#Check & prestage databases
-if [ ! -e "/var/lib/crowdsec/data/GeoLite2-ASN.mmdb" ] && [ ! -e "/var/lib/crowdsec/data/GeoLite2-City.mmdb" ]; then
-    mkdir -p /var/lib/crowdsec/data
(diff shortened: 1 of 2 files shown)


docker/docker_start.sh
```diff
@@ -13,11 +13,17 @@ KEY_FILE="${KEY_FILE:-/etc/ssl/key.pem}"
...
-fi
+# Check & prestage databases
+for geodb in GeoLite2-ASN.mmdb GeoLite2-City.mmdb; do
+    # We keep the pre-populated geoib databases in /staging instead of /var,
+    # because if the data directory is bind-mounted from the host, it will be
+    # empty and the files will be out of reach, requiring a runtime download.
+    # We link to them to save about 80Mb compared to cp/mv.
+    if [ ! -e "/var/lib/crowdsec/data/$geodb" ] && [ -e "/staging/var/lib/crowdsec/data/$geodb" ]; then
+        mkdir -p /var/lib/crowdsec/data
+        ln -s "/staging/var/lib/crowdsec/data/$geodb" /var/lib/crowdsec/data/
```
he2ss, line 24: Just need to check in case of updating the enrichment parser, when the new geodb files will be downloaded and overwritte the symlinks it will work.

mmetc: Yes. This was already merged with the slim PR.
When the parser is upgraded, crowdsec follows the symlink and updates the files in the /staging/var/lib/crowdsec/data directory.
I think replacing the links with real files would make more sense because they could point to a read-only partition, and I'm going to open an issue for this.

PR #1863, opened 2022-11-07
fix #1860 : Only repeat the WAL warning once

### LINKED ISSUE #1860: db_config.use_wal

The new message about use_wal doesn't really describe how to enable it well. We should include a commented out line so user doesnt need to think about un commenting it.
```
time="07-11-2022 10:04:34" level=warning msg="you are using sqlite without WAL, this can have an impact of performance. If you do not store the database in a network share, set db_config.use_wal to true. Set explicitly to false to disable this warning." context=ent
```

### Why is this needed?

For ease for users to enable it without needing to think what key is.

buixor:
fix #1860


AlteredCoder (approved):

LaurenceJJones (approved):
LGTM

PR #1878, opened 2022-11-21
set BUILD_VERSION to the correct tag

mmetc:
This fixes a regression introduced in 0449ec18 (Windows Support #1159) where the build system uses the last local tag that has been created (chonologically) instead of the most recent tag in the current branch. This problem is not detected while working with stable versions because they are built on a clean checkout.


sabban (approved):
lgtm

PR #1888, opened 2022-11-25
Force agent reauth on 403

blotus:
When LAPI starts, by default, a random key is generated to sign the JWT token used by the agent.

This means that in a multi-server setup, if LAPI restarts, all agents will fail to authenticate with their current JWT token until it expires (1h at the maximum) or they are restarted.

The previous code was only checking for a 401 error, check for 403 as well (as this is what is returned by LAPI when the token signature is invalid).


buixor (approved):
lgtm

PR #1895, opened 2022-11-28
Alert context

AlteredCoder:
Add the ability to embed some "context" to alerts. Allows rich alerts and is useful for multi-tenancy.


tests/bats/81_alert_context.bats
```diff
@@ -0,0 +1,52 @@
...
+    ./instance-crowdsec stop
+}
+
+#----------
+
+@test "$FILE 1.1.1.172 has context" {
+    tmpfile=$(TMPDIR="${BATS_TEST_TMPDIR}" mktemp)
+    touch "${tmpfile}"
+    ACQUIS_YAML=$(config_yq '.crowdsec_service.acquisition_path')
+    echo -e "---\nfilename: $tmpfile\nlabels:\n  type: syslog\n" >>"${ACQUIS_YAML}"
```
mmetc, line 36: `echo -e` is not portable, here for the newlines you can use `printf` or `cat <<EOT`.

tests/bats/81_alert_context.bats
```diff
@@ -0,0 +1,52 @@
...
+    ./instance-crowdsec start
+    sleep 2
+    fake_log >>"${tmpfile}"
+    sleep 2
+    rm -f -- "${tmpfile}"
+
+    run -0 cscli alerts inspect 2 -o json
+    run -0 jq -c '.meta | sort_by(.key) | map([.key,.value])' <(output)
+
+    assert_output '[["source_host","[\"sd-126005\"]"],["source_ip","[\"1.1.1.172\"]"],["target_user","[\"netflix\"]"]]'
```
mmetc, line 50: you can have a look at assert_json which is more compact and ignores map order

cmd/crowdsec-cli/alerts.go
```diff
@@ -112,6 +113,27 @@ func DisplayOneAlert(alert *models.Alert, withDetail bool) error {
...
+		sort.Slice(alert.Meta, func(i, j int) bool {
+			return alert.Meta[i].Key < alert.Meta[j].Key
+		})
+		table := newTable(color.Output)
+		table.SetRowLines(false)
+		table.SetHeaders("Key", "Value")
+		for _, meta := range alert.Meta {
+			var valSlice []string
+			if err := json.Unmarshal([]byte(meta.Value), &valSlice); err != nil {
+				log.Fatalf("unknown context value type '%s' : %s", meta.Value, err)
```
mmetc, line 126: please return fmt.Errorf instead of Fatal, so we keep coverage

cmd/crowdsec-cli/console.go
```diff
@@ -29,24 +34,35 @@ func NewConsoleCmd() *cobra.Command {
 		Args:              cobra.MinimumNArgs(1),
 		DisableAutoGenTag: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if err := csConfig.LoadAPIServer(); err != nil || csConfig.DisableAPI {
-				var fdErr *fs.PathError
-				if errors.As(err, &fdErr) {
-					log.Fatalf("Unable to load Local API : %s", fdErr)
+			parentCmd := cmd.Parent()
+
+			// if the command is "cscli console context", only the agent need to be loaded
```
buixor, line 39: *needs

cmd/crowdsec-cli/console.go
```diff
@@ -225,6 +242,193 @@ Disable given information push to the central API.`,
...
+			printHelp(cmd)
+		},
+	}
+
+	var keyToAdd string
+	var valuesToAdd []string
+	cmdContextAdd := &cobra.Command{
+		Use:               "add",
+		Short:             "Add context to send with alerts",
+		DisableAutoGenTag: true,
```
buixor, line 260: I think we will need more explanation on this as it can be confusing, including a "longer" example.

cmd/crowdsec-cli/console.go
```diff
@@ -270,9 +474,143 @@ func SetConsoleOpts(args []string, wanted bool) {
...
+			if !inSlice(fieldName, ret) {
+				ret = append(ret, fieldName)
+			}
+		}
+	}
+
+	return ret
+}
+
+func detectNode(node parser.Node, parserCTX parser.UnixParserCtx) []string {
```
buixor, line 526: we might want to extend to `evt.Meta.` and `evt.Enriched.` ?

cmd/crowdsec-cli/utils.go
```diff
@@ -732,3 +732,21 @@ func formatNumber(num int) string {
 	res := math.Round(float64(num)/float64(goodUnit.value)*100) / 100
 	return fmt.Sprintf("%.2f%s", res, goodUnit.symbol)
 }
+
+func removeFromSlice(val string, slice []string) []string {
```
buixor, line 736: won't it remove the last item if `val` isn't found ?

pkg/leakybucket/manager_load.go
```diff
@@ -348,6 +351,20 @@ func LoadBucket(bucketFactory *BucketFactory, tomb *tomb.Tomb) error {
 		return fmt.Errorf("invalid bucket from %s : %v", bucketFactory.Filename, err)
 	}
 	bucketFactory.tomb = tomb
+
+	bucketFactory.LabelsToSendCompiled = make(map[string][]*vm.Program)
```
buixor, line 355: no debug version on those ?

pkg/leakybucket/overflows.go
```diff
@@ -14,9 +15,85 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/antonmedv/expr"
+	"github.com/antonmedv/expr/vm"
 	"github.com/crowdsecurity/crowdsec/pkg/exprhelpers"
 )
 
+const (
+	maxContextValueLen = 4000
```
buixor, line 23: we most likely want this to be configurable :<

pkg/leakybucket/overflows.go
```diff
@@ -14,9 +15,85 @@ import (
...
+)
+
+func truncate(values []string) (string, error) {
```
buixor, line 26: IIRC we have a similar behavior for storing meta in DB, maybe we can use the same code to do it ?

pkg/parser/unix_parser.go
```diff
@@ -50,6 +52,45 @@ func Init(c map[string]interface{}) (*UnixParserCtx, error) {
 	return &r, nil
 }
 
+// Return new parsers
+// nodes and povfwnodes are already initialized in parser.LoadStages
+func NewParsers() *Parsers {
```
buixor, line 57: windows support (note)

buixor (changes requested):
wip, looks great

buixor (approved):
LGTM

cmd/crowdsec-cli/console.go
AlteredCoder: Its already work with evt.Meta. and can't work with evt.Enriched

mmetc:
> What about _not_ having a default value, and waiting for the user to explicitly enable it ? (While writing this I realize it might make the `cscli` more tricky and/or force the user to edit his config).
> 
> Let's discuss it.

Not the OP but you have a point. Part of the issue is that we allow configuration files outside $CONFIG_DIR which is not useful in practice.
In general, I lean towards "allow the user to set a directory but hardcode the file names".

cmd/crowdsec-cli/console.go
```diff
@@ -187,9 +187,17 @@ Disable given information push to the central API.`,
 	cmdConsole.AddCommand(cmdDisable)
 
 	cmdConsoleStatus := &cobra.Command{
-		Use:               "status [feature-flag]",
-		Short:             "Shows status of one or all feature flags",
-		Example:           "status tainted",
+		Use:   "status [feature-flag]",
+		Short: "Shows status of one or all feature flags",
+		Example: `sudo cscli console status
++-------------+-----------+---------------------------------------------------+
```
blotus, line 193: We do not have any output example in `cscli`, and the output can be somehow confusing when running `cscli console status -h`

mmetc: Another thing - we use the term "feature-flags" for both console and crowdsec/cscli. This is confusing especially when I merge "cscli config feature-flags". What can we do? @buixor ?

buixor: Removed the output not to confuse people. Agreed with @mmetc on the feature-flag confusion; hence replaced with "option" and "console option" in help messages. I believe it's more accurate and less confusing

buixor (approved):
LGTM

PR #1896, opened 2022-11-28
Suggest bouncers and machines to delete

sabban:
Original work from @sbs2001


sabban:
tested and working.

cmd/crowdsec-cli/machines.go
```diff
@@ -311,6 +311,25 @@ cscli machines add MyTestMachine --password MyPassword
...
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			var err error
+			dbClient, err = getDBClient()
+			if err != nil {
+				cobra.CompError("unable to create new database client: " + err.Error())
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			}
+			machines, err := dbClient.ListMachines()
+			if err != nil {
+				cobra.CompError("unable to list machines " + err.Error())
```
LaurenceJJones, line 323: You dont return early here? guessing machines would be nil?

cmd/crowdsec-cli/utils.go
```diff
@@ -732,3 +733,18 @@ func formatNumber(num int) string {
 	res := math.Round(float64(num)/float64(goodUnit.value)*100) / 100
 	return fmt.Sprintf("%.2f%s", res, goodUnit.symbol)
 }
+
+func getDBClient() (*database.Client, error) {
```
buixor, line 737: there is already two implementations of `getDBClient` : in pkg/apiserver and pkg/exprhelpers. Can you either either of those ?

go.sum
```diff
@@ -611,6 +612,8 @@ github.com/nxadm/tail v1.4.6 h1:11TGpSHY7Esh/i/qnq02Jo5oVrI1Gue8Slbq0ujPZFQ=
 github.com/nxadm/tail v1.4.6/go.mod h1:+ncqLTQzXmGhMZNUePPaPqPvBxHAIsmXswZKocGu+AU=
 github.com/oklog/run v1.0.0 h1:Ru7dDtJNOyC66gQ5dQmaCa0qIsAUFY3sFpK1Xk8igrw=
 github.com/oklog/run v1.0.0/go.mod h1:dlhp/R75TPv97u0XWUtDeV/lRKWPKSdTuV0TZvrmrQA=
+github.com/olekukonko/tablewriter v0.0.5 h1:P2Ga83D34wi1o9J6Wh1mRuqd4mF/x/lgBS7N7AbDhec=
```
buixor, line 615: why does it touch the go.mod ?

buixor (commented):
few remarks, overall LGTM

cmd/crowdsec-cli/bouncers.go
```diff
@@ -153,6 +154,25 @@ cscli bouncers add MyBouncerName -k %s`, generatePassword(32)),
...
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			var err error
+			dbClient, err = getDBClient()
+			if err != nil {
+				cobra.CompError("unable to create new database client: " + err.Error())
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			}
+			bouncers, err := dbClient.ListBouncers()
+			if err != nil {
+				cobra.CompError("unable to list bouncers " + err.Error())
```
LaurenceJJones, line 166: Same again dont return early here?

cmd/crowdsec-cli/utils.go
sabban: I am afraid I either I don't understand what you meant, or I miss something here.

`pkg/exprhelpers` db stuff seems to me completely unrelated as it seems to require something already initialized. I don't think there's any implementation of this in `pkg/apiserver` There's one implementation in `pkg/database` that is used here.

In fact this function check that there's a database then it returns a client to it.

go.sum
sabban: It shouldn't, fixed.

cmd/crowdsec-cli/bouncers.go
sabban: same here, cobra.CompError doesn't fatal, but this is used for completion for the shell to complete, so I guess we don't want to fatal here...

cmd/crowdsec-cli/utils.go
buixor: nevermind I read way too fast :<

buixor (approved):
lgtm

PR #1902, opened 2022-11-30
use a copy of bucket processors in LeakRoutine

blotus:
Fix a memory leak caused by sharing bucket processors across all buckets: this could lead to race conditions when using multiple buckets routines and having a bucket using `distinct`, which could create a bucket that discards its first event and not setting up its underflow timer, preventing it from ever being destroyed.


buixor (approved):
thx

PR #1903, opened 2022-11-30
add -error flag to crowdsec binary

mmetc:
This change makes the log initialization happen in a single place, and as soon as possible. This means any info-level messages, like "push and pull blah blah" are not shown in stderr anymore, by default. They end up in the regular logs.


sabban (approved):
lgtm

sabban (approved):
lgtm

PR #1908, opened 2022-12-02
Api server changes

LaurenceJJones:
Since the capacity is going to round the decision or alert length we might as well set that as the capacity to minimize the chances of mem or copy operations. Also the make string could be simplified since it can be initialised within the same line.


mmetc:
Oh yes I wanted to do this purely for the null output. Good that it has performance benefits too.While you are at it, can you have a look at alerts/machines/bouncers? See https://github.com/opnsense/plugins/blob/master/security/crowdsec/src/opnsense/service/conf/actions.d/actions_crowdsec.conf#L54

buixor:
I think this branch will be killed in favour of the stream-based approach that is WIP for decisions/stream. Please keep in mind and comment on upcoming PR @mmetc if you want to change the behaviour w/ empty/null arrays.

PR #1919, opened 2022-12-11
docker/README: automatic registration with tls (#1909)

mmetc:


docker/README.md
```diff
@@ -163,13 +171,20 @@ https://docs.crowdsec.net/docs/user_guides/bouncers_configuration/
 
 ### Automatic Bouncer Registration
 
-You can automatically register bouncers with the crowdsec container at startup, using environment variables or Docker secrets. You cannot use this process to update an existing bouncer without first deleting it.
+Without TLS:
```
sabban, line 174: I would write TLs authentatication here, because one can have a tls lapi without having tls auth, hence the statement might be a bit misleading.

mmetc: ok! I'll change it in the master as well

sabban (approved):
lgtm

PR #1927, opened 2022-12-12
Stream decisions from db

blotus:
When querying `/decisions/stream`, write the decisions as soon as possible to the client.

This allows to greatly reduce the memory usage of crowdsec in some "extreme" situations (eg, lots of bouncers querying LAPI at the same time to get a lot of decisions).

For comparison with the master branch, with around 300k decisions in the database and 100 bouncers (ie, 100 requests in parallel to `/decisions/stream?startup=true`):
 - Master: Crowdsec used around 44G of RAM (and apparently never released it)
 - This PR: With a page size of 10k (configurable in a future commit), memory usage never goes above 1G.
 
 Response time is around the same as before for startup request (if not a little bit faster), but for delta requests it seems to be quite longer (1ms -> 150ms, still under investigation).


buixor (changes requested):
PR looks cool to me. Some minor cleanup (comments) to do. Should we put this behind a feature flag ?

buixor (changes requested):
can you FF it plz

mmetc (approved):

PR #1929, opened 2022-12-15
docker: separate CLIENT_* and LAPI_* variables for tls certificates

mmetc:
Legacy configurations (using CERT_FILE, KEY_FILE and CACERT_FILE) still work as long as DISABLE_AGENT=true, otherwise the container crashes looking for the client certificate (in the default location which is now /etc/ssl/crowdsec-client/)

### DIFF
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -91,9 +91,14 @@ cscli_if_clean() {
 #-----------------------------------#
 
 if [ -n "$CERT_FILE" ] || [ -n "$KEY_FILE" ] || [ -n "$CACERT_FILE" ]; then
-    echo "Cannot start crowdsec container: the variables CERT_FILE, KEY_FILE and CACERT_FILE" >&2
-    echo "are not supported since 1.4.4, please replace them with the LAPI_* and CLIENT_* equivalents." >&2
-    exit 1
+    printf '%b' '\033[0;33m'
+    echo "Warning: the variables CERT_FILE, KEY_FILE and CACERT_FILE have been deprecated." >&2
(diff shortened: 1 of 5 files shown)


mmetc:
I hesitate because both client and server certificates default to /etc/ssl/.
In helm we can point them to /etc/ssl/crowdsec-agent/ and /etc/ssl/crowdsec-lapi/; what should we do here with respect to previous installations?

mmetc:
There must be a simpler way to set default values in yq

docker/docker_start.sh
```diff
@@ -122,37 +128,39 @@ elif [ -n "$USE_WAL" ] && isfalse "$USE_WAL"; then
...
-# regenerate local agent credentials (ignore if agent is disabled)
-if isfalse "$DISABLE_AGENT"; then
-    if isfalse "$DISABLE_LOCAL_API"; then
-        echo "Regenerate local agent credentials"
-        cscli machines delete "$CUSTOM_HOSTNAME" 2>/dev/null || true
-        cscli machines add "$CUSTOM_HOSTNAME" --auto --url "$LOCAL_API_URL"
-    fi
+# regenerate local agent credentials (even if agent is disabled, cscli needs a
+# connection to the API)
+cscli machines delete "$CUSTOM_HOSTNAME" 2>/dev/null || true
```
LaurenceJJones, line 133: Shouldnt we only delete if disable lapi is false?

he2ss (approved):

docker/docker_start.sh
LaurenceJJones: We also might be dealing with a remote lapi and the credentials make us remove ourselves?

LaurenceJJones (approved):
Tried to see if I could break it but couldnt so LGTM

PR #1933, opened 2022-12-19
feature flags

mmetc:
Package fflag provides a simple feature flag system.

 Feature names are lowercase and can only contain letters, numbers, undercores
 and dots.

 good: "foo", "foo_bar", "foo.bar"
 bad: "Foo", "foo-bar"

 A feature flag can be enabled by the user with an environment variable
 or by adding it to {ConfigDir}/feature.yaml

 I.e. CROWDSEC_FEATURE_FOO_BAR=true
 or in feature.yaml:
```
 ---
 - foo_bar
```

 If the variable is set to false, the feature can still be enabled
 in feature.yaml. Features cannot be disabled in the file.

 A feature flag can be deprecated or retired. A deprecated feature flag is
 still accepted but a warning is logged. A retired feature flag is ignored
 and an error is logged.

 A specific deprecation message is used to inform the user of the behavior
 that has been decided when the flag is/was finally retired.

### DIFF
--- a/cmd/crowdsec/main.go
+++ b/cmd/crowdsec/main.go
@@ -330,7 +330,12 @@ func LoadFeatureFlags(cConfig *csconfig.Config, logger *log.Logger) error {
 			enabledFeatures = append(enabledFeatures, feature)
 		}
 	}
-	logger.Infof("Enabled features: %s", strings.Join(enabledFeatures, ", "))
+
+	msg := "<none>"
+	if len(enabledFeatures) > 0 {
+		msg = strings.Join(enabledFeatures, ", ")
+	}
+	logger.Infof("Enabled features: %s", msg)
 
 	return nil
 }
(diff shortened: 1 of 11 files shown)


cmd/crowdsec/main.go
```diff
@@ -295,9 +297,45 @@ func LoadConfig(cConfig *csconfig.Config) error {
...
+	if err != nil {
+		return fmt.Errorf("unable to get feature status: %s", err)
+	}
+
+	for feature, enabled := range featureStatus {
+		if enabled {
+			enabledFeatures = append(enabledFeatures, feature)
+		}
+	}
+	logger.Infof("Enabled features: %s", strings.Join(enabledFeatures, ", "))
```
buixor, line 333: maybe only log if at least one feature was enabled ?

buixor (approved):
minor remark, overall LGTM !

cmd/crowdsec/main.go
mmetc: yes I forgot to put the <none>. I prefer to be explicit

PR #1934, opened 2022-12-21
Cscli config refactoring

mmetc:
Moved cobra boilerplate and run functions to their own blocks, and separate files according to the command to be called. Avoid the use of global variables by defining and consuming the flags in separate places. No functional change otherwise.


LaurenceJJones:
kind/chore
area/cscli

buixor (approved):

PR #1937, opened 2022-12-21
runtime feature flag initialization

mmetc:
without an explicit initializer, some pointer is missing which causes segfault down the road


sabban (approved):
lgtm

PR #1945, opened 2022-12-23
separate cobra constructors:  lapi, machines, bouncers, postoverflows

mmetc:
I just moved around stuff like I did for "cscli config". The aim is to have

 - smaller functions, fewer indentation levels
 - fewer global variables
 - more testable code (eventually in pkg/ or cmd/internal/)


buixor (approved):
lgtm

PR #1954, opened 2022-12-29
Improve warnings around lack of `evt.StrTime` field

### LINKED ISSUE #1951: Improve cold log / time machine warning about missing timestamps

/kind enhancement

We currently have a warning when a line does not have a timestamp in cold log mode (as it will not go into buckets)

### Why is this needed?

Make sure it clear to the user why this behaviour is happening.

buixor:
- fix #1951
 - Improve the message at runtime when trying to pour some events lacking `evt.StrTime` to buckets (in forensic mode)
 - Make `cscli hubtest run` warn you when you're generating logs **without** `evt.StrTime` field


LaurenceJJones (approved):
LGTM

blotus (approved):

PR #1957, opened 2022-12-29
retry with backoff requests to CAPI

nitescuc:


mmetc:
Functional tests are failing because they don't wait enough before asserting that lapi is available or not (and we can't wait 3 minutes for the 5 attempts).
Can we make this configurable? Even an on/off switch would fix the issue.

he2ss (approved):

PR #1960, opened 2022-12-30
fix func tests

mmetc:


.github/workflows/go-tests-windows.yml
```diff
@@ -16,7 +16,6 @@ on:
 
 env:
   RICHGO_FORCE_COLOR: 1
-  CROWDSEC_FEATURE_DISABLE_HTTP_RETRY_BACKOFF: true
```
nitescuc, line 19: maybe leave the variable here to avoid some tests taking forever

.github/workflows/go-tests.yml
```diff
@@ -30,7 +30,6 @@ env:
   # and to override our endpoint in aws sdk
   AWS_ENDPOINT_FORCE: http://localhost:4566
   KINESIS_INITIALIZE_STREAMS: "stream-1-shard:1,stream-2-shards:2"
-  CROWDSEC_FEATURE_DISABLE_HTTP_RETRY_BACKOFF: true
```
nitescuc, line 33: maybe leave the variable here to avoid some tests taking forever

PR #1962, opened 2023-01-02
Add conditional bucket

blotus:
Add a new bucket type that will only overflow if an expr returns true.

It allows for fail2ban style scenarios (where we count the number of events over a fixed time window):
```
type: conditional
name: demo/conditional
description: "pouet"
filter: "evt.Meta.log_type == 'http_access-log'"
groupby: evt.Meta.source_ip
conditional_overflow: len(queue.Queue) >= 5
leakspeed: 2m
capacity: -1
labels:
  type: scan
```

or to detect more complex behaviours:
```
conditional_overflow: any(queue.Queue, {.Meta.http_path == "/"}) and any(queue.Queue, {.Meta.http_path == "/foo"})
```


pkg/leakybucket/manager_load.go
```diff
@@ -108,6 +109,19 @@ func ValidateFactory(bucketFactory *BucketFactory) error {
 		if bucketFactory.Capacity != 0 {
 			return fmt.Errorf("trigger bucket must have 0 capacity")
 		}
+	} else if bucketFactory.Type == "conditional" {
+		if bucketFactory.ConditionalOverflow == "" {
+			return fmt.Errorf("conditional bucket must have a condition")
+		}
+		if bucketFactory.Capacity != -1 {
```
buixor, line 116: please warn on capacity that is not `-1` but do not error.

buixor (changes requested):
don't make -1 mandatory

buixor (approved):
lgtm

PR #1968, opened 2023-01-04
fix yq behavior with bind-mount config.yaml

mmetc:
yq cannot replace a file if it's bind-mounted (i.e. as a file, not inside a volume or directory). This allows it to open for writing without replacing the file.


LaurenceJJones (approved):
LGTM

Sorry I tested this last week should of ticked

PR #1971, opened 2023-01-04
fix parser test 2k23

mmetc:


buixor (approved):

CyrilBrulebois:
Hello,

Any chance the current year could be injected automatically, instead of hardcoding 2022, then 2023, then…? While it's rather low maintenance, that means stable releases will fail to build without changes during the life cycle of a Debian stable release. And it also means that reproducible builds fail, as they tend to build in the past and in the future to see how time affects the resulting build products.

Thanks for considering! I might try and come up with a patch myself, but I wanted to check with you what you thought about that idea.

Cheers,
Cyril.

CyrilBrulebois:
The #2119 fix looks very good, tests are good even when performed on 2022-01-01, thanks!

PR #1986, opened 2023-01-12
improve feature flag logging

mmetc:
For cscli: it should provide a terse output, not nag users with configuration details. Although it's usually important that cscli and crowdsec have the same enabled features, having it list them every time the command is invoked can be too much.

For crowdsec: when features are set from the environment, it's too early to log where we should. So we can use log.Debug at activation time, and list them again once logging is configured.

 - wrap some functions in csconfig for convenience and DRY
 - for each enabled feature, log.Debug
 - log all enabled features once as Info (crowdsec) or Debug (cscli)
 - file does not exist -> log.Trace

### DIFF
--- a/cmd/crowdsec-cli/main.go
+++ b/cmd/crowdsec-cli/main.go
@@ -66,9 +66,14 @@ func initConfig() {
 		csConfig = csconfig.NewDefaultConfig()
 	}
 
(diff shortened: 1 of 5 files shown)


buixor (changes requested):
lgtm

buixor (approved):

PR #1989, opened 2023-01-13
Store go module name in var in Makefile

junnhy5:
For a question of cleanliness and if the project will evolve in the future at the level of the not (even if there is little chance that it is the case), to store the name of the module in a variable and call it then. 😊


LaurenceJJones:
/are configuration
/kind enhancement

Thank you for the pull request! yes this can be merged. Thank you!

junnhy5:
@LaurenceJJones merging is blocked by the CI 😌

mmetc:
GO_MODULE_NAME = "github.com/crowdsecurity/crowdsec"

you have to remove the quotes: GO_MODULE_NAME=github...

mmetc (approved):

PR #1999, opened 2023-01-17
Fix reference to ghcr.io

benscobie:
Very minor thing I spotted.


LaurenceJJones (approved):
Thanks for typo fix! LGTM

PR #2005, opened 2023-01-19
docker: add cri-logs collection by default to support CRI log format

he2ss:


LaurenceJJones:
Some kubernetes user may change `container_runtime` but do not realise the base s00 parser is not included unless they add it to the environment args

LaurenceJJones (approved):
LGTM

mmetc (approved):

PR #2010, opened 2023-01-20
Option to disable remote lapi registration

LaurenceJJones:
We currently do not offer the ability to disable remote lapi registration, so if a user exposes the port without firewalling anyone could send a request which will be added to the database.

We could add `trustedips` but this doesn't make much sense since we are turning it off for remote completely/


pkg/apiserver/controllers/controller.go
```diff
@@ -83,7 +84,13 @@ func (c *Controller) NewV1() error {
 	})
 
 	groupV1 := c.Router.Group("/v1")
-	groupV1.POST("/watchers", c.HandlerV1.CreateMachine)
+	groupV1.POST("/watchers", func(gctx *gin.Context) {
```
mmetc, line 87: can you extract and give a name to this handler?

AlteredCoder (approved):

PR #2011, opened 2023-01-20
Add machines prune command

LaurenceJJones:
Easy way to prune machines if user has alot of unvalidated or machines with a lastheartbeat greater than a duration `10 minutes` by default.

Will add tests later, creating draft PR for opinions.


mmetc:
I think we should display a table of the machines to delete (with machines_table.go), and pass the list to the prune function instead of calling the query again. If there are 0 machines, no prompt at all

LaurenceJJones:
I dont agree with displaying the machines if the users has lots then it doesnt add value to show it. I do agree with passing the list to prune then we can use the same func maybe for bouncers.

mmetc:
So there is no way for the users to see what they are deleting

LaurenceJJones:
True, I was going for the feel of what other prune feel like, however, they are not consistent
```
[loz@unknown-dev ~]$ podman volume prune
WARNING! This will remove all volumes not used by at least one container. The following volumes will be removed:
0d883d60c21590b948affae15433270de75072d8a6d65f5e758e264a7c4842d6
57b5fc9ea32f232d0e5456bdfdfeb8d02cc9bad1947617c93b020c1c0c0bf820
7240a203e1fcf33194de5863818fa2fd5471fe8ee2a0911f3c9b99d2dc090e96
cd622d98f7fb33791905702ea9405d83f05af3e2967d06bdd4a733c7dd80e429
Are you sure you want to continue? [y/N] N
[loz@unknown-dev ~]$ docker volume prune
WARNING! This will remove all local volumes not used by at least one container.
Are you sure you want to continue? [y/N] 
```

LaurenceJJones:
> 

There is a slight overlap but a user may want to only run the command once since they generate too many machines / bouncers and doesnt want to add the config just to restart crowdsec to flush then remove config and restart crowdec. For one flush that's a bit long winded when we can provide a single prune command.

mmetc (approved):

Issue #2008, opened 2023-01-20
expanded env tweaks


LaurenceJJones:
### What happened?

Following #2007 

We have found the underlying issue is within os.expandEnv func that will treat the `$` within the any field as an environment var even if the variable is not found. Purposed idea is to write our own func that will return the original string if the env var is not found.

### What did you expect to happen?

To pass the original string without replacing the not found env var

### How can we reproduce it (as minimally and precisely as possible)?

Look at #2007

### Crowdsec version

N/A

LaurenceJJones:
Fixed in main branch, closing as completed

PR #2012, opened 2023-01-21
allow use of literal $ in config.yaml

mmetc:
With this change, references to undefined $VARIABLES in a configuration file are not replaced by empty strings (which damages passwords with $ signs) but left unchanged. This can also help users to spot incomplete configurations.

For now only config.yaml but it can be used by others and bouncers as well.


LaurenceJJones (approved):
LGTM

LaurenceJJones:
#2008

PR #2019, opened 2023-01-25
v3 capi and blocklists links support

nitescuc:


buixor (approved):
LGTM, ensure that we do tests with a very short pull period to check for leaks

PR #2043, opened 2023-02-09
use the tag name from the predefined variable for azure pipeline (#2041)

sabban:
use the tag name from the predefined variable for azure pipeline (#2041)


LaurenceJJones (approved):
LGTM as it merged in main

PR #2069, opened 2023-02-22
fix the way acquisition is stopped

sabban:
fix the way acquisition is stopped: drain the inputLineChan before terminating the acquisition routine to make sure that the acquisition routine doesn't get stuck filling a chan that isn't read anywhere.


sabban:
I ensured that there's no lock added of any kind

sabban:
I may come with a better way to do this.

blotus (approved):

PR #2091, opened 2023-03-02
Add `cscli papi status` and `cscli papi sync`

blotus:
- `cscli papi status`: allow to see if crowdsec is allowed to interact with PAPI, the current console plan if allowed, and the categories the instance is subscribed to
 - `cscli papi sync`: Force a sync of all PAPI orders. All non-expired orders will be pulled (even if the instance already received them).
 - Remove `/v1/decisions/stream/poll` from papi_url in the configuration


buixor (changes requested):
minor changes but overall LGTM !

buixor (approved):
lgtm

PR #2095, opened 2023-03-03
add papi_url in credentials file when enabling console_management, and remove it when disabling console_management

blotus:


buixor (approved):
lgtm

PR #2097, opened 2023-03-06
Load lapi config for `config show output`

mmetc:
This adds URL and login parameters as it was intended.
Also rewrite configShow and displayOneAlert to use an embedded text/template for shorter code.


buixor (approved):
lgtm

blotus (approved):

PR #2100, opened 2023-03-06
add ToString() helper

blotus:
Add a `String()` helper to cast an interface to a string.
This *does not* convert any type to string, it only validate a value can be used as a string (for example, the `UnmarshalJSON` helper returns `map[string]interface{}`, which prevents to use its output in other helpers)


buixor (approved):
I don't know if the name is really good, but I don't have a better name to suggest neither :shrug:

buixor (approved):
lgtm, but fix tests and labels

PR #2103, opened 2023-03-07
Run lint after tests instead of separate workflow

mmetc:
This should consume slightly less resources on the CI, but lint is run only if the tests are passing.


sabban (approved):
lgtm

PR #2118, opened 2023-03-13
Only use pgx for postgresql database

blotus:
- Always use the pgx driver when using a postgres DB
 - pq has shown to provoke huge memory leak in previous under some conditions that were very hard to debug (almost no trace in a pprof heap dump for example)
 - Supporting both drivers add some complexity for the functional tests (only pgx is tested currently)

### DIFF
--- a/pkg/csconfig/database.go
+++ b/pkg/csconfig/database.go
@@ -102,9 +102,10 @@ func (d *DatabaseCfg) ConnectionDialect() (string, string, error) {
 		return "sqlite3", dialect.SQLite, nil
 	case "mysql":
 		return "mysql", dialect.MySQL, nil
-	case "postgres", "postgresql":
-		return "postgres", dialect.Postgres, nil
-	case "pgx":
+	case "pgx", "postgresql", "postgres":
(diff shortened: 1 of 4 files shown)


pkg/csconfig/database.go
```diff
@@ -102,9 +102,10 @@ func (d *DatabaseCfg) ConnectionDialect() (string, string, error) {
 		return "sqlite3", dialect.SQLite, nil
 	case "mysql":
 		return "mysql", dialect.MySQL, nil
-	case "postgres", "postgresql":
-		return "postgres", dialect.Postgres, nil
-	case "pgx":
+	case "pgx", "postgresql", "postgres":
+		if d.Type != "pgx" {
+			log.Debugf("database type '%s' is deprecated, switching to 'pgx' instead", d.Type)
```
buixor, line 107: why debug then ?

buixor (changes requested):
maybe make the warning more obvious ? can we ensure doc is up to date too ?

buixor (approved):
lgtm

PR #2119, opened 2023-03-13
try to make reproducible build work

sabban:
Small change to make debian reproducible build work

### DIFF
--- a/pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse_test.go
+++ b/pkg/acquisition/modules/syslog/internal/parser/rfc3164/parse_test.go
@@ -1,6 +1,7 @@
 package rfc3164
 
 import (
+	"fmt"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


blotus (approved):

CyrilBrulebois:
Without this, I'm getting a failure to build (during the test suite) when setting back the clock to 2022-01-01; with this patch applied on top of the 1.4.6-1 Debian package, the test suite passes. I'm not confident (yet) this is going to be sufficient to fix build failures in the reproducible-builds.org environment, but this is definitely going to help! I should have more data once 1.4.6-2 gets uploaded, and build-tested there.

CyrilBrulebois:
https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/crowdsec.html says things are looking good now:

> reproducible 1.4.6-3 in unstable
> at 2023-03-19 05:38:00 UTC

PR #2122, opened 2023-03-15
properly update the time structure within event

buixor:
This change is to ensure it works in time-machine, allowing users to do scenarios like:

```yaml
# ssh bruteforce
type: trigger
name: crowdsecurity/ssh-bf
description: "Detect ssh bruteforce outside of office hours"
filter: "evt.Meta.log_type == 'ssh_failed-auth' && (evt.Time.Hour() >= 20 || evt.Time.Hour() < 6) || (evt.Time.Weekday().String() == 'Saturday' || evt.Time.Weekday().String() == 'Sunday')"
groupby: evt.Meta.source_ip
blackhole: 1m
reprocess: true
```

### DIFF
--- a/pkg/parser/enrich_date.go
+++ b/pkg/parser/enrich_date.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	expr "github.com/crowdsecurity/crowdsec/pkg/exprhelpers"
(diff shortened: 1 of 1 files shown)


LaurenceJJones:
Just checking this wont break all of the hubtests since its a new property? @blotus ? but the property will be missing from all tests?

PR #2130, opened 2023-03-20
S3 acquisition datasource

blotus:
Add support for reading logs from a S3 bucket.

The datasource supports has 2 ways to detect new objects:
 - SQS: the bucket is configured to send a notification to a SQS queue. Crowdsec will subscribe to the queue and get the new objects in real time.
 - Listing: Crowdsec will list the content of the bucket at specific interval and read new objects. This method is slow, and does not scale with big buckets (S3 listing operation is slow). It is mostly intended for testing purposes.

### DIFF
--- a/pkg/acquisition/modules/s3/s3.go
+++ b/pkg/acquisition/modules/s3/s3.go
@@ -35,7 +35,7 @@ type S3Configuration struct {
 	Key                               string  `yaml:"-"` //Only for DSN acquisition
 	PollingMethod                     string  `yaml:"polling_method"`
 	PollingInterval                   int     `yaml:"polling_interval"`
-	SQSARN                            string  `yaml:"sqs_arn"`
+	SQSName                           string  `yaml:"sqs_name"`
 }
 
 type S3Source struct {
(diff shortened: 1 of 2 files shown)


pkg/acquisition/modules/s3/s3.go
```diff
@@ -283,23 +308,29 @@ func (s *S3Source) sqsPoll() error {
 			logger.Debugf("Received %d messages from SQS", len(out.Messages))
 			for _, message := range out.Messages {
 				sqsMessagesReceived.WithLabelValues(s.Config.SQSName).Inc()
-				eventBody := S3Event{}
-				err := json.Unmarshal([]byte(*message.Body), &eventBody)
+				bucket, key, err := s.extractBucketAndPrefix(message.Body)
 				if err != nil {
 					logger.Errorf("Error while parsing SQS message: %s", err)
+					//Always delete the message to avoid infinite loop
+					_, err = s.sqsClient.DeleteMessage(&sqs.DeleteMessageInput{
```
he2ss, line 315: In the documentation, we may need to specify to use a specific queue only for crowdsec. Because sometimes, users will use one queue for multiple needs, and as crowdsec deletes the messages, it can be a mess.

he2ss (approved):

pkg/acquisition/modules/s3/s3.go
```diff
@@ -0,0 +1,594 @@
...
+				}
+			}
+			if newObject {
+				lastObjectDate = *bucketObjects[len(bucketObjects)-1].LastModified
+			}
+		}
+	}
+}
+
+func (s *S3Source) extractBucketAndPrefix(message *string) (string, string, error) {
```
he2ss, line 264: Here it's better to check the type of message and set it to avoid checking each time the message type.

pkg/acquisition/modules/s3/s3.go
```diff
@@ -0,0 +1,594 @@
...
+var sqsMessagesReceived = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "cs_s3_sqs_messages_total",
+		Help: "Number of SQS messages received per queue.",
+	},
+	[]string{"queue"},
+)
+
+func (s *S3Source) newS3Client() error {
+	var sess *session.Session
```
mmetc, line 117: Maybe like this?

```
options := session.Options{
	SharedConfigState: session.SharedConfigEnable,
}

if s.Config.AwsProfile != nil {
	options.Profile = *s.Config.AwsProfile
}

sess := session.Must(session.NewSessionWithOptions(options))
```

buixor (approved):

PR #2138, opened 2023-03-23
Add experimental support for re2

blotus:
This PR adds support (behind a feature flag) for using re2 instead of the default golang regexp package for GROK patterns and the `RegexpInFile` helper.

The feature flags are:
 - re2_grok_support
 - re2_regexp_in_file_support

Early benchmarks show:
 - a ~100% performance improvement when parsing a nginx log file with 200k in replay mode without the bad-user-agent scenario (default config, 1 parser routine / 1 bucket routine)
 - If using the bad UA scenario, and with re2 support for expr helpers,  a ~40% improvement (default config, 1 parser routine / 1 bucket routine)
 - On a 16 core machine, with 8 parser routines and 8 bucket routine, without bad UA, the time to parse the file is the same but CPU usage is halved when using re2.


buixor:
(note: all tests are in one-shot mode, so account far agent startup and shutdown)


1. **MASTER** nginx collection without `http-bad-user-agent` + 1 parser/bucket

- nginx-10k: `23,98s user 0,57s system 185% cpu 13,233 total`
- nginx-20k: `41,89s user 0,68s system 214% cpu 19,839 total`
- nginx-50k: `101,25s user 1,34s system 236% cpu 43,406 total`

2. **RE2-FFLAG** nginx collection without `http-bad-user-agent` + 1 parser/bucket

- nginx-10k: `20,40s user 0,54s system 155% cpu 13,453 total`
- nginx-20k: `36,04s user 0,77s system 182% cpu 20,133 total`
- nginx-50k: `74,51s user 1,25s system 232% cpu 32,546 total`


Notes:

  - with RE2, the startup (between first print and crowdsec reading the file) is significantly slower
    - on master, startup takes ~1 second
    - on RE2, startup takes ~5 seconds (mainly the compilation of the "default" patterns that seems to take a good 4s)

buixor:
image

PR #2149, opened 2023-03-29
CI: test with postgres 15

mmetc:
Postgres 15 restricts the default privileges for the public schema. We set crowdsec_test as owner which is shorter than grant this and grant that.


ankit-gautam23 (commented):
It would be great if you could enclose the test result screenshot here.

mmetc:
> It would be great if you could enclose the test result screenshot here.

Oh yes, thanks. I have not found the time to dig further

image

PR #2150, opened 2023-03-29
CI: Static builds by default; replace bincover with go -cover from 1.20 

mmetc:
This should also fix functional tests coverage with static builds


sabban (approved):
got a wakthrough from mmetc
seems fine to me: remove a lot of code as well love it

PR #2167, opened 2023-04-12
Make it more obvious that parser succeeded but was whitelisted

### LINKED ISSUE #1781: "Parser failure" on whitelisted line
When getting users to use `cscli explain` it can be quite confusing to some that the end result is `parser failure` when the value is whitelisted? Maybe moving whitelisted out of the inner for loop to check if whitelisted? as it didnt fail to parse

https://github.com/crowdsecurity/crowdsec/blob/af30ba0e3b3393402ee8c1533d7d5f9e6507bca7/pkg/cstest/parser_assert.go#L439

LaurenceJJones:
Fix #1781


buixor (approved):
lgtm ! thanks

PR #2189, opened 2023-05-10
make hostname available in notification template ;-)

Cyrille37:
systemd define %H as the hostname. The hostname is human readable in notification message ;-)


Cyrille37:
Here is the discussion about different possible solution to make available the hostname in notification go_template:
https://discourse.crowdsec.net/t/notification-another-identifiing-machine-than-id/1284/

After different tries, this commit seems to be the simplest & efficient solution.

PR #2190, opened 2023-05-10
make hostname available in notification template ;-)

Cyrille37:
systemd define %H as the hostname. The hostname is human readable in notification message ;-)

Here is the discussion about different possible solution to make available the hostname in notification go_template:
https://discourse.crowdsec.net/t/notification-another-identifiing-machine-than-id/1284/

After different tries, this commit seems to be the simplest & efficient solution.

/area configuration
/kind enhancement

### DIFF
--- a/debian/crowdsec.service
+++ b/debian/crowdsec.service
@@ -5,6 +5,7 @@ After=syslog.target network.target remote-fs.target nss-lookup.target
 [Service]
 Type=notify
 Environment=LC_ALL=C LANG=C
+Environment=HOSTNAME=%H
 ExecStartPre=/usr/bin/crowdsec -c /etc/crowdsec/config.yaml -t
 ExecStart=/usr/bin/crowdsec -c /etc/crowdsec/config.yaml
 #ExecStartPost=/bin/sleep 0.1


Cyrille37:
A Slack template can start like :
```yaml
format: |
  from ({{ env "HOSTNAME" }}):
  ...
```

buixor:
Hello @Cyrille37 !

Thanks for the PR. However, if I'm correct, [as sprig already supports env](http://masterminds.github.io/sprig/os.html) it should do the job.

Can you let me know if it does?

Cyrille37:
Hello @buixor 

But `systemd` has a sandbox feature that the service doesn’t have access to all environments so you will have to edit the service file to add the environment key.

buixor:
ah my bad, I get your point.

The PR has too much adherence to systemd debian. 

If it sounds good to you, we would add a expr + sprig helper so that it doesn't involve systemd etc.

blotus:
We've merged https://github.com/crowdsecurity/crowdsec/pull/2193, which provides a new `Hostname()` helper in both expr and the notification templates.

In the meantime, if you still want to get the hostname from the env, you can use `systemctl edit crowdsec` to add an override for the unit file and add the env var here (it will stay across upgrades).

Cyrille37:
Thanks all :-) That's great !

PR #2214, opened 2023-05-21
generate asserts for evt.Unmarshaled in hubtest

blotus:
Handle `evt.Unmarshaled` in hubtest: Some parsers now use either `ParseKV` or `UnmarshalJSON`, which will write to `evt.Unmarshaled`.
Generate asserts for them if it is non-nil.
We only generate asserts for maps, and basic types. Slices are ignored at the moment.


buixor (approved):
lgtm

PR #2220, opened 2023-05-23
Log info capi whitelists

### LINKED ISSUE #2219: INFO log capi whitelist
https://github.com/crowdsecurity/crowdsec/blob/eddb994c0b48d77b34a3f22b719dc5716670d2ae/pkg/csconfig/api.go#L361

Currently the user only gets notified if the file is:

- Not in the location provided
- Invalid YAML format
- Invalid IP in `ips`
- Invalid CIDR in `cidr`

User should be notified that Crowdsec saw the configuration change and have loaded it.

LaurenceJJones:
Fix #2219


pkg/csconfig/api.go
```diff
@@ -252,6 +252,8 @@ func (c *Config) LoadAPIServer() error {
 
 		if err := c.API.Server.LoadCapiWhitelists(); err != nil {
 			return err
+		} else if c.API.Server.CapiWhitelistsPath != "" {
+			log.Infof("loade capi whitelist from %s: %d IP's, %d CIDR's", c.API.Server.CapiWhitelistsPath, len(c.API.Server.CapiWhitelists.Ips), len(c.API.Server.CapiWhitelists.Cidrs))
```
mmetc: loaded?

mmetc: also s/'//

buixor (approved):

mmetc (approved):

PR #2241, opened 2023-05-29
Auto detect if reading logs or storing sqlite db on a network share

blotus:


pkg/acquisition/modules/file/file.go
```diff
@@ -330,7 +330,21 @@ func (f *FileSource) StreamingAcquisition(out chan types.Event, t *tomb.Tomb) er
...
+		if f.config.PollWithoutInotify != nil {
+			inotifyPoll = *f.config.PollWithoutInotify
+		} else {
+			networkFS, fsType, err := types.IsNetworkFS(file)
+			if err != nil {
+				f.logger.Warningf("Could not get fs type for %s : %s", file, err)
+			}
+			f.logger.Debugf("fs for %s is network: %t (%s)", file, networkFS, fsType)
+			if networkFS {
+				inotifyPoll = false
```
buixor, line 343: Is this worth explicit logging ? "Disabling inotify poll on %s as it is a network share. You can manually set inotify to false to make this message disappear, or to true to enforce inotify poll"

pkg/csconfig/database.go
```diff
@@ -62,7 +65,30 @@ func (c *Config) LoadDBConfig() error {
...
 	if c.DbConfig.Type == "sqlite" {
 		if c.DbConfig.UseWal == nil {
-			log.Warning("You are using sqlite without WAL, this can have a performance impact. If you do not store the database in a network share, set db_config.use_wal to true. Set explicitly to false to disable this warning.")
+			dbDir := filepath.Dir(c.DbConfig.DbPath)
+			isNetwork, fsType, err := types.IsNetworkFS(dbDir)
+			if err != nil {
+				log.Warnf("unable to determine if database is on network filesystem: %s", err)
+				log.Warning("You are using sqlite without WAL, this can have a performance impact. If you do not store the database in a network share, set db_config.use_wal to true. Set explicitly to false to disable this warning.")
+				return nil
+			}
```
buixor, line 74: same as above, we want logging if we override the users parameters based on our detection :)

buixor (changes requested):
moar explicit logging

buixor (approved):

mmetc:
@blotus conflicts are fixed, keep in mind that on master we don't care about at all (no warning, no check) about WAL from cscli. I kept the same behavior in this PR.

PR #2248, opened 2023-05-31
expose the FormatAlert function to other packages

blotus:
Needed to allow the playground to test notification templates with a behavior as close as possible to crowdsec.


buixor (approved):
lgtm

PR #2252, opened 2023-06-01
CI: publish dev-debian docker image (amd64 only)

mmetc:


.github/workflows/publish_docker-image_on_master-debian.yml
```diff
@@ -0,0 +1,70 @@
+name: Publish Docker image on Push to Master
```
he2ss, line 1: publish Debian Docker image

he2ss (approved):

PR #2266, opened 2023-06-05
Implement "crowdsec -fatal" flag; change help message

mmetc:
The -trace...-error flags do not change the log destination but only the verbosity. This change reflects that, and implements "-fatal" which was missing. If it's not the behavior we think users expect, we'll have to change the configuration setup.


sabban (approved):
lgtm

PR #2267, opened 2023-06-05
don't pre-create log files (not required anymore)

mmetc:
The lumberjack package fixed the issue in https://github.com/natefinch/lumberjack/pull/83 (tested with umask 002) and this code is now redundant since we updated the dependency to v2.2.1.


LaurenceJJones:
Should we also catch the crowdsec_api.log also?

buixor (approved):

PR #2269, opened 2023-06-05
Move grok_pattern.go away from pkg/types to reduce bouncer dependencies

mmetc:
The right place seems pkg/parser, and all files using grok_pattern already import it.


sabban (approved):
lgtm

Issue #2121, opened 2023-03-14
Docker startup script does not use local config


ToshY:
### What happened?

I have setup Crowdsec and a MySQL database in my docker compose configuration (see reproduction steps), and when upgrading from `v1.4.2-debian` to `v1.4.3-debian`, I keep getting the following error message in my Crowdsec container:

```
chown: cannot access '/var/lib/crowdsec/data/crowdsec.db': No such file or directory
```

After extensive debugging, I've actually found out that starting from `v1.4.3`, the `docker_start.sh` had been changed to include `shopt -s inherit_errexit`, which results on exit when something fails in the script, which in this case is `chown` operation. Making this stricter is fine and not the issue, but it made me realise that it only read `/etc/crowdsec/config.yaml` and doesn't do anything with possible local configuration files that are present, e.g. `/etc/crowdsec/config.yaml.local`.

#### TLDR

The `docker_start.sh` does not take into consideration `config.yaml.local`, unlike stated in the [documentation](https://github.com/crowdsecurity/crowdsec/blob/master/docker/README.md#full-configuration) (which was changed in [PR 2115](https://github.com/crowdsecurity/crowdsec/pull/2115)).

### What did you expect to happen?

Local config gets merged with the standard config.

* Check if `config.yaml.local` exists, if so merge it with the standard config (local values overriding standard config).
* Use [yq](https://mikefarah.gitbook.io/yq/operators/multiply-merge#merge-two-files-together) for merging.

### How can we reproduce it (as minimally and precisely as possible)?

`./crowdsec/Dockerfile`
```Dockerfile
ARG CROWDSEC_IMAGE_VERSION

FROM crowdsecurity/crowdsec:${CROWDSEC_IMAGE_VERSION}

COPY ./config/config.yaml.local \
     /etc/crowdsec/
```

`./crowdsec/config/config.yaml.local`
```yaml
db_config:
  log_level: info
  type: ${DATABASE_TYPE}
  user: ${DATABASE_USER}
  password: ${DATABASE_USER_PASSWORD}
  db_name: ${DATABASE_NAME}
  host: ${DATABASE_HOST}
  port: ${DATABASE_PORT}
api:
  server:
    log_level: info
    listen_uri: 0.0.0.0:8081
    trusted_ips:
      - 172.17.0.0/16
```

`./.env`
```dotenv
CROWDSEC_IMAGE_VERSION="v1.4.3"
MYSQL_HOST="mysql"
[...]

PR #2272, opened 2023-06-06
docker: always merge .yaml.local in conf_get()

mmetc:
With this change, all queries to the configuration will return the values from .local if they are set. However, conf_set will only write to .yaml and never to .local. This means users can potentially override values that are supposed to be under control of the entrypoint (credentials and things set from envvars).

See https://github.com/crowdsecurity/crowdsec/issues/2121


mmetc:
@LaurenceJJones can I have some feedback if this solves your use cases

buixor (approved):
code-wise ok, let's wait for @LaurenceJJones feedback :)

LaurenceJJones (approved):
LGTM

LaurenceJJones (approved):

PR #2278, opened 2023-06-10
Syslog Notification Plugin

LaurenceJJones:
Initial POC for syslog plugin notifications

Code looks very basic, however, I don't believe there is any additional features we need to add.


plugins/notifications/syslog/syslog.yaml
```diff
@@ -0,0 +1,25 @@
...
+# max_retry:          # Number of attempts to relay messages to plugins in case of error
+# timeout:            # Time to wait for response from the plugin before considering the attempt a failure, eg "10s"
+
+#-------------------------
+# plugin-specific options
+
+# The following template receives a list of models.Alert objects
+# The output goes in the logs and to a text file, if defined
+format: |
+  {{.|toJson}}
```
LaurenceJJones, line 18: note to self, add a better format example for syslog

buixor:
Looks cool!

What about making tag an expression? (or use the same templating approach as for the message itself)

LaurenceJJones:
> What about making tag an expression? (or use the same templating approach as for the message itself)

Well, unfortunately we dont get access to the raw alert object from within the plugin so I don't know how this would be achieved.

PR #2281, opened 2023-06-13
allow running rootless docker tests

mmetc:
Taint scenario from inside the container, it's simpler and works when tests are run with rootless docker


buixor (approved):

sabban (approved):
lgtm

PR #2288, opened 2023-06-19
append vendor.tgz to each release

mmetc:
This can be used by ports that don't allow internet access during builds, like freebsd, gentoo etc.


buixor (approved):
LGTM

PR #2291, opened 2023-06-19
support for stdin with "cscli decision import" and raw values

mmetc:
some refactoring and tests too


cmd/crowdsec-cli/decisions.go
```diff
@@ -504,46 +576,50 @@ func NewDecisionsImportCmd() *cobra.Command {
 duration,scope,value
 24h,ip,1.2.3.4
 
-cscsli decisions import -i decisions.csv
+cscli decisions import -i decisions.csv
```
buixor, line 579: we should update the help to:
 - mention it supports stdin with `-i i`
 - mention is support the "plain" format that is simply the ip, one per line

buixor (changes requested):
let's update the help, but looks good :)

cmd/crowdsec-cli/decisions.go
```diff
@@ -478,191 +481,259 @@ cscli decisions delete --type captcha
...
+	if format == "" {
+		if strings.HasSuffix(input, ".json") {
+			format = "json"
+		} else if strings.HasSuffix(input, ".csv") {
+			format = "csv"
+		}
+	}
+
+	if format == "" {
+		return fmt.Errorf("unable to guess format from file extension, please provide a format with --format flag")
```
LaurenceJJones, line 585: Could this line just sit at the bottom of the upper if statement

LaurenceJJones (commented):
LGTM, the only comment is just because we don't need two if statements in succession

cmd/crowdsec-cli/decisions.go
mmetc: No because it needs to check if it was detected. The alternative is not one if, but two nested ifs which is less readable.

buixor (approved):

LaurenceJJones:
Just note that your bats test may fail as origin is no longer overwritable

mmetc:
> Just note that your bats test may fail as origin is no longer overwritable

What do you mean "no longer"? It can be overwritten from csv or json, but not from a cli flag. Same as the previous version. Should I add the flag?

LaurenceJJones:
> > Just note that your bats test may fail as origin is no longer overwritable
> 
> What do you mean "no longer"? It can be overwritten from csv or json, but not from a cli flag. Same as the previous version. Should I add the flag?

It can be since 1.5 there was a merge to revert it to always be cscli-import

https://github.com/crowdsecurity/crowdsec/blob/507da49b5ad6f4f7c4eda328ae17b9ff3e627a17/cmd/crowdsec-cli/decisions.go#L556

PR #2294, opened 2023-06-21
spellcheck/style leakybucket readme

mmetc:


pkg/leakybucket/README.md
```diff
@@ -126,17 +133,17 @@ leakspeed and an infinite capacity [capacity set to -1 for now].
 
 [This is not dry enough to have many details here, but:]
 
-The bucket code is triggered by `InfiniBucketify` in main.go.
-There's one struct called buckets which is for now a
+The bucket code is triggered by `InfiniBucketify` in `main.go`.
```
sabban, line 136: This is not true anymore, it's run from `runPour` in `pour.go`. So I suggest 

The bucket code is triggered by `runPour` in `pour.go`, by calling the `leaky.PourItemToHolders` function.

sabban (commented):
One assertion isn't true anymore

PR #2295, opened 2023-06-21
Build with libre2 by default, options for wasm and static; add mk/gmsl

mmetc:
By default build with re2-dev, dynamic link. Error if re2 is not found.

If BUILD_RE2_WASM=1, build with WebAssembly.
If BUILD_STATIC=1, link everything statically (either wasm or re2) which may fail on fedora, etc.
Provide an example in Dockerfile with re2 download, static build and install.


LaurenceJJones (approved):
LGTM

PR #2306, opened 2023-06-24
Loki integration #2

lperdereau:
Refer to : Loki integration #1574

- Code updated
- All old tests pass
- Work with latest Loki version


lperdereau:
Strangely, when I exec:
- `make release`
- `cd <build>`
- `mkdir tests`
- `./test_env.sh`
- `cd tests`
- `crowdsec -c dev.yaml -type keycloak -dsn 'loki://localhost:3100/?query={filename="/var/log/keycloak.log"}'`

I get

image

mmoscher:
Interesting! Any news on this? What's missing and/or where is help needed?

lperdereau:
> Interesting! Any news on this? What's missing and/or where is help needed?

No news, I'm trying to continue work on it if you want to contribute we can try to list what's missing.
It's an existing branch who I updated and increase test by adding it to docker compose tests.

pschiffe:
Looking forward to this feature, hopefully you'll be able to finish it :crossed_fingers:

lperdereau:
I hope this feature can be review by a maintainer. Just few comment about the feature and keys to complete this PR.

lperdereau:
Thank you @LaurenceJJones for CI pipeline.
I've just patch the build error caused by :

```go
package loki_test
import "github.com/crowdsecurity/go-cs-lib/pkg/cstest"
```

Updated into:
```go
package loki_test
import "github.com/crowdsecurity/go-cs-lib/cstest"
```

buixor:
Hey,

Sorry for the delay, having a look at it now and in the coming days :)
Thanks a lot,

pkg/acquisition/modules/loki/loki.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/crowdsecurity/crowdsec/pkg/types"
+)
+
+const (
+	readyTimeout time.Duration = 3 * time.Second
+	readyLoop    int           = 3
+	readySleep   time.Duration = 10 * time.Second
+	lokiLimit    int           = 100
+)
+
```
buixor, line 32: I suggest we extend the metrics by adding a counter for dropped lines.
While it might be tempting to ask you to add the query, it will end up looking ugly. So I will wait for "pretty name" to become standard in datasource interface and add it afterwards.

pkg/acquisition/modules/loki/loki.go
```diff
@@ -0,0 +1,339 @@
...
+		Headers:  l.Config.Headers,
+		Limit:    l.Config.Limit,
+		Query:    l.Config.Query,
+		Since:    l.Config.Since,
+		Username: l.Config.Auth.Username,
+		Password: l.Config.Auth.Password,
+	}
+
+	l.Client = lokiclient.NewLokiClient(clientConfig)
+	l.Client.Logger = logger.WithField("component", "lokiclient")
```
buixor, line 132: I suggest adding the source as a field/label to the logger.

pkg/acquisition/modules/loki/loki.go
```diff
@@ -0,0 +1,339 @@
...
+			return errors.Wrap(err, "could not start loki tail")
+		}
+		for {
+			select {
+			case resp, ok := <-respChan:
+				if !ok {
+					ll.Warnf("loki channel closed")
+					return err
+				}
+				if len(resp.DroppedEntries) > 0 {
```
buixor, line 308: This is where we will track the amount of dropped lines for the prom metrics. Logging here the full info of source (including the query) will make sense, wdyt ?

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
```diff
@@ -0,0 +1,255 @@
...
+
+	Since        time.Duration
+	Until        time.Duration
+	WaitForReady time.Duration
+
+	Limit int
+}
+
+func (lc *LokiClient) tailLogs(conn *websocket.Conn, c chan *LokiResponse, ctx context.Context) error {
+	tick := time.NewTicker(100 * time.Millisecond)
```
buixor, line 45: If we don't read "fast enough" from loki, we will lose messages (as far as I understand). While we can track it with the metrics and log messages, having a dynamic ticker might help. I suggest the following approach:
 - we can start with a 50ms ticker
 - if we don't get a message when reading, we can increase the ticker duration (ie. double it, with 100ms being the upper bound)
 - on the other hand, if we miss messages, we lower the ticker duration (ie. divide it by two, and go as low as 5ms)

I'm not sure about the upper/lower bound we should pick with the ticker as I don't have XP with Loki, but I hope you get the idea :smile:

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func (lc *LokiClient) Tail(ctx context.Context) (chan *LokiResponse, error) {
+	responseChan := make(chan *LokiResponse)
+	dialer := &websocket.Dialer{} //TODO: TLS support
+	u := lc.getURLFor("loki/api/v1/tail", map[string]string{
+		"limit": strconv.Itoa(lc.config.Limit),
+		"start": strconv.Itoa(int(time.Now().Add(-lc.config.Since).UnixNano())),
+		"query": lc.config.Query,
```
buixor, line 185: it seems the `delay_for` parameter allows to do some magic to deal with log duplication and so on. I suggest we let the user decide how to configure it ?

cf. https://github.com/grafana/loki/issues/931 and https://github.com/grafana/grafana/issues/20346

buixor (changes requested):
Hello,

Here goes a quick review. It looks overall good to me; I have a few suggestions. Please let me know what you think. If we agree, it can go in the next release.

Thanks for your contribution :+1: 

Self note : When creating the associated documentation, we need to make it loud & clear that the user can "miss" some log lines depending on his config and the volume of logs he wants to ingest.

pkg/acquisition/modules/loki/loki.go
lperdereau: No problem to add a counter for dropped lines on metrics.

You are right we should wait a standard unique name in datasource interface.

pkg/acquisition/modules/loki/loki.go
lperdereau: I agree with you, it seems to be here for the prom metrics.

And yes, we should write most of the information we can send back to the user.

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
lperdereau: Totally we can add `delay_for` params

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
lperdereau: I don't have any information about this choice made by the previous developer.
I thought it might be a websocket connection keepalive.

I will say: If we don't read "fast enough" from loki, we will lose connection.

I did a search on grafana which allows the user to stream Loki's logs live ([feature documentation](https://grafana.com/docs/grafana/latest/explore/logs-integration/#live-tailing)).
Here the codebase [grafana/grafana](https://github.com/grafana/grafana/blob/5b63cdb5b0fed1c4c6df879e54b7adb96ac2711c/pkg/tsdb/loki/streaming.go#L62)

We find the same idea
```go
        ticker := time.NewTicker(time.Second * 60) //.Step)
	defer ticker.Stop()

	for {
		select {
		case <-done:
			logger.Info("Socket done")
			return nil
		case <-ctx.Done():
			logger.Info("Stop streaming (context canceled)")
			return nil
		case t := <-ticker.C:
			count++
			logger.Error("Loki websocket ping?", "time", t, "count", count)
		}
	}
```

I agree with we can have a dynamic ticker.

I'm not sure about this ticker, but I still think it's a connection keeper.

lperdereau (commented):
Thank you @buixor for the review and comment.
It's really good to see activity on this 😃

pkg/acquisition/modules/loki/loki.go
lperdereau: https://github.com/crowdsecurity/crowdsec/blob/master/pkg/acquisition/acquisition.go#L110-L112
This code doesn't do it?

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
lperdereau: Development: done
Tests: done

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
lperdereau: I dig into grafana code base, and I understand that they have an UI who allow stream aquisition for user via a proxy API who tail loki logs.
This ticker and and the above code (in grafana repository) is used to keep connection between user's browser and Grafana Proxy API.
In this way we don't need this for our usage. I simplified the implementation.

buixor:
Hey thanks,

Having a look ! Hopefully this time I won't keep you waiting ;)

buixor:
Hello,

I made a few minor changes in the code (bubbling or errors, not capitalizing errors and a few logging clarifications).
However during our internal testing we realized that the `tail` approach can lead to some missed messages. An easy way to reproduce it is like using `ab` (concurrency of 10 with few thousand messages) on a web-server plugged to loki: Loki tail is going to miss a significant part of the messages.

While investigating, we found out that, according to [this issue](https://github.com/grafana/loki/issues/4489#issuecomment-1015417405),  `The tail endpoint is not built with the intention of providing all logs for a stream` :disappointed: 

We are experimenting with the query approach [that another guy recommends](https://github.com/grafana/loki/issues/4489#issuecomment-1013821232). I will let you know if the PoC is conclusive.

Stay tuned, you can drop by on discord if you want to chat about it!

buixor (approved):
Thanks a lot @lperdereau for the contribution. Tests and code is fine for me. It shall ship in 1.5.6 :partying_face:

PR #2309, opened 2023-06-26
non-fatal error if some datasource can't be run (i.e. journalctl but systemd is missing)

mmetc:
This on the other hand, gives a new fatal error when there are no valid datasources. In the previous version, crowdsec kept running with just a warning if no acquisition yaml or dir were specified.


sabban (approved):
lgtm

sabban (approved):
lgtm

PR #2316, opened 2023-06-28
add journalctl for ssh by default

sabban:
add journalctl for ssh by default since crowdsec wont't return nor crash if the datasource is not usable.


LaurenceJJones (commented):
In any user environment this change will not affect anything, as when wizard.sh runs the auto detection it deletes the default acquis.yaml

PR #2319, opened 2023-06-28
Add merge helpers to merge maps into maps if the value is string

LaurenceJJones:
This is useful with the new unmarshaled on the event. You may want to move everything from `evt.Unmarshaled.traefik` to `evt.Parsed` I provide two methods one that will override any key on the target or a safe option.

Since `evt.Parsed` and `evt.Meta` expect the value to be a string. I have to use to ToString method, however, we should explore to expand this helper function as any other type it pretty much fails.

This above situation does **NOT** affect ParseKV since it always returns a string, however, UnmarshalJSON has the potential to return other values.

If PR is accepted I will add explicit documentation that if the values are not strings you will need to set them manually after the merge example:

```
onsuccess: next_stage
debug: true
name: crowdsecurity/kafka-logs
description: "Parse kafka json logs"
## 1.5.2 returns "" , 1.5.3 returns nil
filter: evt.Parsed.program == 'kafka' && UnmarshalJSON(evt.Parsed.message, evt.Unmarshaled, "kafka") in [
"", nil]
statics:
    - meta: service
      value: kafka
   -  meta: unused
      expression: MergeSafe(evt.Unmarshaled.kafka, evt.Meta)
 ## Since connectionScore is float/int it is ignore be Merge and MergeSafe so user must explicitly set the value since ToString (Our helper) does NOT type cast it correctly.
    - meta: connection_score
      expression: int(evt.Unmarshaled.kafka.connectionScore)
```

### DIFF
--- a/pkg/exprhelpers/expr_lib.go
+++ b/pkg/exprhelpers/expr_lib.go
@@ -419,6 +419,20 @@ var exprFuncs = []exprCustomFunc{
 			new(func() (string, error)),
 		},
 	},
+	{
+		name:     "Merge",
+		function: Merge,
+		signature: []interface{}{
+			new(func(map[string]interface{}, map[string]string) error),
+		},
+	},
+	{
+		name:     "MergeSafe",
+		function: MergeSafe,
+		signature: []interface{}{
+			new(func(map[string]interface{}, map[string]string) error),
+		},
+	},
 }
 
 //go 1.20 "CutPrefix":              strings.CutPrefix,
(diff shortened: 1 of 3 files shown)


LaurenceJJones:
My own comment: Maybe we should call MergeMap and MergeMapSafe since user may think it merges different things ?!?

LaurenceJJones:
Converting to draft as some dont feel this is needed as may cause issues if user does not use it correctly.

PR #2320, opened 2023-06-29
Add some extra string helpers

LaurenceJJones:
I will add a helper to convert keys on an object and apply a "modifier" this PR has those modifiers

### DIFF
--- a/pkg/exprhelpers/exprlib_test.go
+++ b/pkg/exprhelpers/exprlib_test.go
@@ -1456,9 +1456,9 @@ func TestToCamelCase(t *testing.T) {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			vm, err := expr.Compile(tc.expr, GetExprOptions(map[string]interface{}{"value": tc.value})...)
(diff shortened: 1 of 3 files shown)


pkg/exprhelpers/exprlib_test.go
```diff
@@ -1429,3 +1429,71 @@ func TestParseKv(t *testing.T) {
...
+			expected: "FooBar",
+			expr:     `SnakeToCamel(value)`,
+		},
+	}
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			vm, err := expr.Compile(tc.expr, GetExprOptions(map[string]interface{}{"value": tc.value})...)
+			assert.NoError(t, err)
+			output, err := expr.Run(vm, map[string]interface{}{"value": tc.value})
```
mmetc, line 1460: require.NoError() because if err!=nil the rest of the test is not useful

LaurenceJJones:
shelving PR atm till there a use case

PR #2323, opened 2023-07-02
Cscli CTI command

LaurenceJJones:
Draft POC for cscli CTI command


LaurenceJJones:
Converting to draft as internally we discussed `cscli` should only have commands to interact with crowdsec locally maybe we should provide a separate package for this

PR #2362, opened 2023-07-18
Add options to disable sharing and receiving from CAPI

LaurenceJJones:
Added two option to disable sharing and receiving from CAPI

This is useful if you want to send signals but not receive and the other way around for premium console users.


mmetc (commented):
Can you avoid using pointers for the bools? If you want the default to be zero value, which makes sense, you can reverse the logic and have
 NoShareSignals and NoCommunityBlocklist.

LaurenceJJones:
> NoShareSignals and NoCommunityBlocklist

DisableSharingSignals and DisableCommunityBlocklists rather than no but I agree.

LaurenceJJones:
Closing PR creating an issue

PR #2363, opened 2023-07-19
Reduce log verbosity at startup

mmetc:
A configuration syntax test is performed every time the service is started from systemd. The resulting error, if any, is shown on journalctl logs.
This PR removes the unnecessary output in crowdsec.log generated by the configuration test.


LaurenceJJones (approved):
LGTM

PR #2369, opened 2023-07-21
update metabase to v0.46.6.1

blotus:

PR #2370, opened 2023-07-21
Add metabase version override and update

LaurenceJJones:


LaurenceJJones:
Link to #2369 but adds cscli flag to override version so users are not hardcoded to a single version

mmetc:
Two things - 

1 - the name of the variable would be metabaseVersion, not metabaseDefaultVersion

2 - we should let the user override metabaseImage string instead

LaurenceJJones:
> Two things - 
> 
> 1 - the name of the variable would be metabaseVersion, not metabaseDefaultVersion

Agree

> 
> 2 - we should let the user override metabaseImage string instead

I disagree that the user should be able to pass the whole image name as if an admin allows the user to run as sudo they will be able to run any container

mmetc:
> > 2 - we should let the user override metabaseImage string instead
> 
> I disagree that the user should be able to pass the whole image name as if an admin allows the user to run just cscli as sudo they will be able to run any container

which can be done already with cscli explain

cmd/crowdsec-cli/dashboard.go
```diff
@@ -199,6 +199,7 @@ cscli dashboard setup -l 0.0.0.0 -p 443 --password <password>
 	cmdDashSetup.Flags().BoolVarP(&force, "force", "f", false, "Force setup : override existing files")
 	cmdDashSetup.Flags().StringVarP(&metabaseDbPath, "dir", "d", "", "Shared directory with metabase container")
 	cmdDashSetup.Flags().StringVarP(&metabaseListenAddress, "listen", "l", metabaseListenAddress, "Listen address of container")
+	cmdDashSetup.Flags().StringVarP(&metabaseDefaultVersion, "version", "v", metabaseDefaultVersion, "Metabase version to use")
```
blotus, line 202: `metabaseDefaultVersion` should be `metabaseVersion`

blotus (changes requested):
Looks mostly ok to me, but as @mmetc said, the user should be able to override the whole image. (I don't really consider this to be a security risk if they have sudo for cscli, as they can already execute arbitraty commands as root easily).

blotus (approved):

PR #2378, opened 2023-07-25
Build target for "make tidy"

mmetc:
The make tidy target runs "go mod tidy" in the root directory and all plugins.


LaurenceJJones (approved):
LGTM

Issue #2383, opened 2023-07-26
[Docker image] Setting disable_local_api to true


LaurenceJJones:
### What happened?

Setting `DISABLE_LOCAL_API` does not set `.API.Server.Enable` to false so when using `cscli` commands within the container that rely on being on the main node. it will bypass this check.

### What did you expect to happen?

Cscli commands that do

```go
if err := csConfig.LoadAPIServer(); err != nil || csConfig.DisableAPI {
   return fmt.Errorf("local API is disabled, please run this command on the local API machine: %w", err)
}
```

Should fail the PersistentPreRunE

### How can we reproduce it (as minimally and precisely as possible)?

```bash
podman run --rm -d -e DISABLE_LOCAL_API=true crowdsecurity/crowdsec:v1.5.2
podman -lit cscli machines add -a
## Should fail as machine add should only run on lapi
```

I am using modified code for notifications as that was the initial report by user which lead me to find this within the containers should not run outside the main node and this check fails with that code.

[branch](https://github.com/LaurenceJJones/crowdsec/tree/cscli_-notifications_pre_run_e)

### Crowdsec version

<details>

```console
$ cscli version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Enabled collections and parsers

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
$ cat /etc/crowdsec/acquis.yaml /etc/crowdsec/acquis.d/*
# paste output here

# On Windows:
C:\> Get-Content C:\ProgramData\CrowdSec\config\acquis.yaml
# paste output here
</details>


### Config show

<details>

```console
$ cscli config show
# paste output here
```

</details>

### Prometheus metrics

<details>

```console
$ cscli metrics
# paste output here
```

</details>

### Related custom configs versions (if applicable) : notification plugins, custom scenarios, parsers etc.

<details>

</details>

mmetc:
Since DISABLE_LOCAL_API is implemented by providing an argument to crowdsec, and this does not affect cscli - I see no other way than to remove the .api.server section from the configuration. We need to skip the other sections in the entrypoint that require lapi, like interaction with capi etc.

[...]

PR #2384, opened 2023-07-26
Add notifications precheck to bail if not main node

LaurenceJJones:


LaurenceJJones:
Initial fix for user report #2383, however, a bigger fix for correctly setting the cscli context is needed.

LaurenceJJones:
closing in favour of #2383

PR #2385, opened 2023-07-26
docker: remove .api.server from config.yaml if DISABLE_LOCAL_API is set

mmetc:
otherwise DISABLE_LOCAL_API has no effect on cscli


LaurenceJJones (approved):
LGTM

cmd/crowdsec-cli/capi.go
LaurenceJJones, line 1: Should we revert these changes in favour of require?

cmd/crowdsec-cli/notifications.go
LaurenceJJones, line 1: Same again revert these changes in favour of require PR?

PR #2388, opened 2023-07-27
Check cscli preconditions with crowdsec-cli/require package

mmetc:


cmd/crowdsec-cli/require/require.go
```diff
@@ -0,0 +1,70 @@
+package require
```
LaurenceJJones, line 1: Add a require PAPI?

```
LAPI()
CAPI()
if csConfig.API.Server.OnlineClient.Credentials.PapiURL == "" {
 return fmt......
}

Comment (mmetc) [reply cmd/crowdsec-cli/require/require.go:1]: done

File: cmd/crowdsec-cli/papi.go [hunk 2]
```diff
@@ -12,6 +11,8 @@ import (
 
 	"github.com/crowdsecurity/crowdsec/pkg/apiserver"
 	"github.com/crowdsecurity/crowdsec/pkg/database"
+
+	"github.com/crowdsecurity/crowdsec/cmd/crowdsec-cli/require"
 )
 
 func NewPapiCmd() *cobra.Command {
```
LaurenceJJones, line 18: PreRunE only needs PAPI as the others are run within PAPI

LaurenceJJones: So it can be as simple as 

```
return require.PAPI(csConfig)
```

LaurenceJJones (approved):
LGTM

PR #2391, opened 2023-07-28
[cscli] notifications test command and slight re write

LaurenceJJones:
Rework some of the cscli notification stuff

- Add a generic test which works with active and non active configurations
- Split the top function so can reuse some code 😄 
- Found an issue which notification configuration were not set the default max retry so if it failed the first time then it would just exit rather than retrying which is crowdsec natural configuration. Had to expose the func as it was private before


LaurenceJJones:
Assign @sabban to review upon return

sabban (approved):
lgtm

cmd/crowdsec-cli/notifications.go
```diff
@@ -224,75 +233,144 @@ func NewNotificationsInspectCmd() *cobra.Command {
...
+			return pluginBroker.Init(csConfig.PluginConfig, []*csconfig.ProfileCfg{
+				{
+					Notifications: []string{
+						cfg.Name,
+					},
+				},
+			}, csConfig.ConfigPaths)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			origin := types.CscliOrigin
```
mmetc, line 273: can you just use the ptr package to avoid all the temporary variables?

cmd/crowdsec-cli/notifications.go
```diff
@@ -224,75 +233,144 @@ func NewNotificationsInspectCmd() *cobra.Command {
...
 func NewNotificationsReinjectCmd() *cobra.Command {
-	var remediation bool
 	var alertOverride string
+	var alert *models.Alert
 
 	var cmdNotificationsReinject = &cobra.Command{
 		Use:   "reinject",
-		Short: "reinject alert into notifications system",
-		Long:  `Reinject alert into notifications system`,
+		Short: "reinject a alert into profiles to trigger notifications",
```
mmetc, line 346: "an alert"

PR #2394, opened 2023-07-31
Fix SSHD_FAIL pattern.

vamega:
Previously auth_type was expected to be a WORD, however this failed to match auth types like `keyboard-interactive` or `keyboard-interactive/pam`.

Example log line that was not being parsed before:

```
Failed keyboard-interactive/pam for invalid user supervisor from 191.5.98.238 port 47846 ssh2
```

SSH server Version Details below
```
❯ sshd -V
OpenSSH_9.3, OpenSSL 3.0.9 30 May 2023
```


LaurenceJJones:
Hey 👋🏻 

Thanks for opening a PR since patterns only update when CrowdSec pushes a release we want to stay away from updating patterns and push a fix via the actual parser so people can enjoy it straight away.

I can do this unless you are comfortable pushing a fix via the hub.

In future versions of the hub we will host patterns there and it will be able to have a proper life cycle not tied to CrowdSec versions

vamega:
Would it be okay for me to wrap that into https://github.com/crowdsecurity/hub/pull/785
I worry that changing this separately from that CR will introduce a conflict.

LaurenceJJones:
> Would it be okay for me to wrap that into [crowdsecurity/hub#785](https://github.com/crowdsecurity/hub/pull/785) I worry that changing this separately from that CR will introduce a conflict.

You can bundle it together

PR #2402, opened 2023-08-02
force rfc 3339 date format in metrics push

blotus:


buixor (approved):

LaurenceJJones:
Be careful the windows test are failing for real as the strings do not match should fix them

Issue #2405, opened 2023-08-03
[Memory management] Postoverflow context


LaurenceJJones:
We load the standard patterns in postoverctx on init but if postoverflows are empty we do not remove leaving to unused memory

https://github.com/crowdsecurity/crowdsec/blob/644c767019c55244067039d1aadb8a8d6c781a33/pkg/parser/unix_parser.go#L108

```
parsers.Povfwctx = &UnixParserCtx{}
```

Should be within the if clause

https://github.com/crowdsecurity/crowdsec/blob/644c767019c55244067039d1aadb8a8d6c781a33/pkg/parser/unix_parser.go#L140

The other question is after all things are compiled shouldnt we just blank both 🤷🏻

LaurenceJJones:
Also saved some memory overwriting the grokky hosts since they may not be used???

```
parsers.Ctx.Grok = grokky.Host{}
parsers.Povfwctx.Grok = grokky.Host{}
return parsers, nil
```

PR #2406, opened 2023-08-03
[1.5.4] slight mem ops

LaurenceJJones:
After we compile all parsers / postoverflows we can nil / empty grokky as it is not reused and can save more in RE2 FF


LaurenceJJones:
Linked to #2405

LaurenceJJones:
Splitting ino smaller PR's for scope purposes

PR #2415, opened 2023-08-09
apic: minor refactoring

mmetc:

PR #2416, opened 2023-08-09
Add CAPI whitelist length check

LaurenceJJones:
If user configures the file but fails to define an actual whitelist we should check length to save allocs


LaurenceJJones:
Closing due to being merge in #2415

PR #2419, opened 2023-08-10
File init improvements

LaurenceJJones:
Current implementation has two flaws:

- If user supplies same file X time, its loaded X times. This problem is extended with RE2 feature flag
- The current map check only checks strings map, so added additional func
- If ftype is wrong bail before loading file in memory


pkg/exprhelpers/helpers.go
```diff
@@ -130,20 +130,26 @@ func UpdateRegexpCacheMetrics() {
...
 		return err
 	}
 	defer file.Close()
 
-	if fileType == "" {
-		log.Debugf("ignored file %s%s because no type specified", fileFolder, filename)
-		return nil
-	}
-	if _, ok := dataFile[filename]; !ok {
-		dataFile[filename] = []string{}
```
sabban, line 145: I may miss something, but I don't see where the global variable dataFile is allocated ?

LaurenceJJones: They were already being made here https://github.com/crowdsecurity/crowdsec/blob/0fc68d8490a075ece3e7a7d0f1980c87ad6a2596/pkg/exprhelpers/helpers.go#L35-L37

AlteredCoder (approved):

PR #2420, opened 2023-08-10
Reset grokky once all patterns are compiled

LaurenceJJones:
Since parserctx is passed to all nodes and held in memory the GC will not auto cleanup the memory footprint so resetting manually saves around 40mb RAM in my production systems

Keep in mind this RAM saving is extreme in my case since im using both RE2 FF, for other users it would still be sizable but not this extreme

### DIFF
--- a/pkg/parser/unix_parser.go
+++ b/pkg/parser/unix_parser.go
@@ -148,6 +148,12 @@ func LoadParsers(cConfig *csconfig.Config, parsers *Parsers) (*Parsers, error) {
 		parsers.Ctx.Profiling = true
 		parsers.Povfwctx.Profiling = true
 	}
-
+	/*
+		Reset CTX grok to reduce memory footprint after we compile all the patterns
+	*/
+	parsers.Ctx.Grok = grokky.Host{}
+	parsers.Povfwctx.Grok = grokky.Host{}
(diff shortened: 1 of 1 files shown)


pkg/parser/unix_parser.go
```diff
@@ -148,6 +148,12 @@ func LoadParsers(cConfig *csconfig.Config, parsers *Parsers) (*Parsers, error) {
 		parsers.Ctx.Profiling = true
 		parsers.Povfwctx.Profiling = true
 	}
-
+	/*
+		Reset CTX grok to reduce memory footprint after we compile all the patterns
+	*/
+	parsers.Ctx.Grok = grokky.Host{}
+	parsers.Povfwctx.Grok = grokky.Host{}
+	parsers.StageFiles = []Stagefile{}
```
LaurenceJJones, line 156: This not needed but we dont need to know the files after we have loaded

Might need to check the reload handler

LaurenceJJones:
image

LaurenceJJones:
Link to #2405

LaurenceJJones:
image

Most optimal performance is RE2 in grokky turned on but not RE2 in file as the regexp provided by us are not complicated and take alot more memory for little gain

AlteredCoder (approved):

PR #2418, opened 2023-08-09
use go 1.21.1

mmetc:

PR #2422, opened 2023-08-10
Runtime whitelist parsing improvement

LaurenceJJones:
So to improve this we implement a few things

- Check if whitelist has any IP base whitelist before looping the sources
- If looping the expressions and the current node is already whitelisted exit early
  - This helps with whitelists that have alot of expressions since we can exit early instead of looping all expressions
- Only check if whitelisted once no need to modify the line above when there a bigger block below the expressions


pkg/parser/whitelist.go
```diff
@@ -22,3 +25,118 @@ type ExprWhitelist struct {
...
+		expression.ExprDebugger, err = exprhelpers.NewDebugger(filter, exprhelpers.GetExprOptions(map[string]interface{}{"evt": &types.Event{}})...)
+		if err != nil {
+			n.Logger.Errorf("unable to build debug filter for '%s' : %s", filter, err)
+		}
+
+		n.Whitelist.B_Exprs = append(n.Whitelist.B_Exprs, expression)
+		n.Logger.Debugf("adding expression %s to whitelists", filter)
+	}
+	valid := false
+	if n.ContainsIPLists() || n.ContainsExprLists() {
```
LaurenceJJones, line 138: We can use the new `ContainsWLs()` instead of both

mazzma12:
[@CodiumAI-Agent](https://github.com/CodiumAI-Agent) /review

LaurenceJJones:
Depends on #2418

pkg/types/event_test.go
```diff
@@ -0,0 +1,82 @@
...
+				nil,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			ips := tt.evt.ParseIPSources()
+			if !slices.EqualFunc(ips, tt.expected_outcome, func(a, b net.IP) bool {
```
mmetc, line 75: I think with testify, `assert.Equal(t, ips, tt.expected_outcome)` should be enough

pkg/types/event_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/crowdsecurity/crowdsec/pkg/models"
+)
+
+func TestParseIPSources(t *testing.T) {
+	tests := []struct {
+		name             string
+		evt              Event
+		expected_outcome []net.IP
```
mmetc, line 15: just call this 'expected' (and expectedErr if you need it too)

pkg/types/event.go
```diff
@@ -73,6 +74,21 @@ func (e *Event) GetMeta(key string) string {
 	return ""
 }
 
+func (e *Event) ParseIPSources() []net.IP {
```
mmetc, line 77: do you need to export this, and the other methods on Node?

pkg/parser/whitelist_test.go
```diff
@@ -0,0 +1,302 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			var err error
+			node.Whitelist = tt.whitelist
+			node.CompileWLs()
+			isWhitelisted := node.CheckIPsWL(tt.event.ParseIPSources())
+			if !isWhitelisted {
+				isWhitelisted, err = node.CheckExprWL(map[string]interface{}{"evt": tt.event})
+			}
+			if err != nil {
```
mmetc, line 294: require.NoError, and no need for message - the failure is clear from test name and context

pkg/parser/whitelist_test.go
```diff
@@ -0,0 +1,302 @@
...
+			expected_err: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			node.Whitelist = tt.whitelist
+			_, err := node.CompileWLs()
+			if err == nil && tt.expected_err {
```
mmetc, line 65: you can test for the actual error message, see how we use

cstest.RequireErrorContains(t, err, tc.expected_err)

pkg/parser/whitelist.go
```diff
@@ -22,3 +25,114 @@ type ExprWhitelist struct {
...
+			}
+			n.Logger.Tracef("whitelist: %s not in [%s]", src, v)
+		}
+	}
+	return isWhitelisted
+}
+
+func (n *Node) CheckExprWL(cachedExprEnv map[string]interface{}) (bool, error) {
+	err := error(nil)
+	output := interface{}(nil)
```
mmetc, line 72: you can remove this declaration and assign output with :=

mmetc (changes requested):

pkg/types/event_test.go
LaurenceJJones:
```
panic: runtime error: comparing uncomparable type []net.IP [recovered]
        panic: runtime error: comparing uncomparable type []net.IP
```

pkg/types/event.go
LaurenceJJones: Yes as it used as a helper function.. unless it can be called privately?

pkg/parser/whitelist.go
LaurenceJJones: Sorry could you elaborate more? Assigning to `output := interface{}` causes an error

LaurenceJJones:
@buixor Can confirm the new whitelist hubtest pass with current code

```
╭─loz repo: hub on  whitelist_test_updates [$] via  v1.21.1 took 18s
[🧱] × hubtest whitelists
─────────────────────
Test         Result
─────────────────────
whitelists   ✅
─────────────────────

╭─loz repo: hub on  whitelist_test_updates [$] via  v1.21.1 took 3s
╰─λ hubtest nextcloud-whitelist
──────────────────────────────
Test                  Result
──────────────────────────────
nextcloud-whitelist   ✅
──────────────────────────────
```

mmetc (approved):

PR #2427, opened 2023-08-12
Add acquis docker start script and add inside readme

LaurenceJJones:


LaurenceJJones:
/kind enhancement
/area agent

I couldn't think fo a better prefix than `ACQUIS` dont know if they a more specific key we can use

mmetc:
If we do it.. why base64? There is no need to encode a plaintext

LaurenceJJones:
Base64 to improve some consistency between shells and docker compose, for example.

mmetc:
> Base64 to improve some consistency between shells and docker compose, for example.

I'm sorry but do you have an example? I don't see how it helps other than obfuscating, shells and yaml handle multiline strings

PR #2429, opened 2023-08-14
move plugins/notifications/* to cmd/notification-*

mmetc:
This ensures all dependencies are kept in sync, and simplifies packaging under freebsd/gentoo/etc because there is a single vendor directory.


sabban (approved):
lgtm

PR #2430, opened 2023-08-16
add basic trainer class

seemanne:


pkg/bayesiantrain/trainer.go
```diff
@@ -0,0 +1,159 @@
...
+	inputChan := make(chan evalIpResult, 1000)
+	outputChan := make(chan evalHypothesisResult)
+	defer close(inputChan)
+	defer close(outputChan)
+
+	for _, v := range s.ParsedIpEvents {
+		go evaluateProgramOnBucket(&v, compiled, inputChan)
+	}
+
+	go controllerRoutine(inputChan, outputChan, s.total)
```
sbs2001, line 127: I’m not aware of the larger context of where this fits in. Looks like you’re adding new function in scenarios/parsers. That being said  I think using go routines won't add performance but is only making the code more complex. The “go-routined” function `evaluateProgramOnBucket` is not doing any I/O work hence doesn't make sense to use go routines on it. Something like
```go
var result evalHypothesisResult
for _, v := range s.ParsedIpEvents {
 r := evaluateProgramOnBucket(&v, compiled, inputChan)
// update result
}
```    
Would have equivalent if not better performance while eliminating the overhead of controller routine etc

seemanne: Ok maybe I should explain. The whole thing is the main training loop for the bayesian buckets. The construction works the following way:
* The logs are loaded into the `LogEventStorage` into the map with the key being the Ip and the all the events for this Ip are added to the `fakeBucket`
* The user can then test different hypothesis expr to see if any of them would make for good conditions in the bucket using `TestHypothesis`
* To speed up the hypothesis testing the idea way to run it in parallel threads for each IP (as its basically counting some stuff per IP) 
* The goal of the channel/goroutine design in `TestHypothesis` is to enable this parallelism by spawing an independent routine for each IP (fake bucket) and then collecting all the results using the controller.  

Does this make more sense now?

sbs2001: Understood, the goroutines would definitely increase throughput. If the training is CPU intensive task than goroutine would make sense.

seemanne: Nice, thank you

PR #2432, opened 2023-08-17
Remove duplicate line

encg:
Remove duplicate line (60) which is outputting `Configuration Folder` twice in `cscli config show`.


buixor (approved):
thx!

PR #2435, opened 2023-08-18
fix the behavior of `-a` option of `cscli alerts list` : do not hide alerts with no associated decisions

buixor:
When the option to include capi alerts (`-a`) is not present in `cscli alerts list`, it was setting up a filter to exclude alerts that are not associated with decisions of scope (crowdsec, cscli, console, cscli-import), which is problematic for alerts with no decisions (ie. auditd, where you needed to do `cscli alerts list --scope pid -a`). This makes it so that option doesn't rely on decisions but on alert.SourceScope directly.

It as well means that by default, alerts with no decisions appear in alerts list.


mazzma12:
[@CodiumAI-Agent](https://github.com/CodiumAI-Agent) /review

mmetc (approved):

PR #2453, opened 2023-09-05
refactor cscli capi

mmetc:


mmetc:
this PR is obsolete but see if we can salvage the capi mock

PR #2454, opened 2023-09-05
add missing scenarios in first login when authenticating with TLS

mmetc:
the method extraction is to avoid adding complexity and code duplication from the fix


pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -46,142 +46,176 @@ func IdentityHandler(c *gin.Context) interface{} {
...
+	ret := authInput{}
 
-		if !clientMachine.IsValidated {
-			return nil, fmt.Errorf("machine %s not validated", machineID)
-		}
+	if err = c.ShouldBindJSON(&loginInput); err != nil {
+		return nil, fmt.Errorf("missing: %w", err)
+	}
+	if err = loginInput.Validate(strfmt.Default); err != nil {
+		return nil, errors.New("input format error")
```
buixor, line 138: do we want to log the actual error here ?

buixor (approved):
slight suggestion, otherwise ok

Issue #2463, opened 2023-09-11
[feat] Email plugin : add send_timeout parameter


Bigouden:
### What would you like to be added?

/kind feature

### Why is this needed?

The current mail plugin configuration allows a maximum of 10s SMTP send delay which cannot be configured.

https://github.com/xhit/go-simple-mail/blob/master/email.go#L215

Some SMTP servers, like my email provider, take longer to accept a message : 

![image](https://github.com/crowdsecurity/crowdsec/assets/9884807/a2f33357-7e68-4b1f-97cf-060e46152560)

Crowdsec ;

![image](https://github.com/crowdsecurity/crowdsec/assets/9884807/a3d92290-d7bb-459f-9fa2-af1b19b7e677)

It would therefore be interesting to have a configuration parameter allowing to control the SMTP sending timeout.

LaurenceJJones:
The new email binary and configuration options will be available next release. Closing due to completion in main.

PR #2464, opened 2023-09-11
[Feat] email timeout options

LaurenceJJones:
#2463

Expose the timeout options to the user but allow send as well as connect incase user wants both


mmetc (changes requested):
I think this could be way simpler. Timeout duration for an email is always on the range of a few seconds, can't be 10 minutes for example, nor it can be 50ms. Why don't we keep the option as "integer, number of seconds" ?

PR #2465, opened 2023-09-11
notification-email: configurable timeouts

mmetc:
https://github.com/crowdsecurity/crowdsec/issues/2463

Expose the timeout options to the user but allow send as well as connect incase user wants both


LaurenceJJones:
@mmetc just move ahead with this impl

/kind enhancement
/area agent

LaurenceJJones:
impl #2463

LaurenceJJones (commented):
add configuration example to email.yaml so they know what keys look like even if commented

buixor (approved):

mmetc:
> add configuration example to email.yaml so they know what keys look like even if commented

Done. However, you were right in using a duration string because we already have a "timeout" option in the generic plugin settings, which we parse as duration. So it's better to stay consistent :/

LaurenceJJones:
> > add configuration example to email.yaml so they know what keys look like even if commented
> 
> Done. However, you were right in using a duration string because we already have a "timeout" option in the generic plugin settings, which we parse as duration. So it's better to stay consistent :/

Yeah, and I thought creating a nested yaml statement allows a user to easily see this is for email, not the plugin timeouts

mmetc:
I see your point but all the other mail-specific options are at the same level. I think the comment in the config example is enough. I added "helo_host" which was missing.

mmetc:
I'll merge after updating docs in https://docs.crowdsec.net/docs/next/notification_plugins/email

PR #2479, opened 2023-09-20
cscli hubtest whitelist

LaurenceJJones:
Add whitelist properties to the gen parser assert function so when crowdsec pulls in the file it can actually check those properties.

ONLY TESTED ON s02-enrich parsers for now as those are the one we have in the hub


LaurenceJJones:
Feedback from @buixor 

We should always print if whitelist is false as we can check for the opposite (Should not be whitelisted and is ? have we messed up the function?)

AlteredCoder (approved):

PR #2489, opened 2023-09-25
Replace encoding/json with goccy/go-json for performance

mmetc:
It occurs to me that we can swap json handling with a popular drop-in replacement https://github.com/goccy/go-json
It should be compatible with the stdlib API. Here if we want to test/benchmark.


mmetc:
closing because the Go team promised us json v2 and they keep their word

PR #2492, opened 2023-09-27
fix the float comparison for hubtests

### LINKED ISSUE #2434: [Hubtest] Float64 and Int dont match crowdsec parser dump
### What happened?

Linked to https://github.com/crowdsecurity/hub/pull/806

When hubtests are run float64 and ints have the potential to be mismatched by the parser_dump.yaml

Reason for this yaml.Unmarshal always unmarshals to 16 bit precision if int or float is too big

Example

CrowdSec Parser dumps shows TS as
```yaml
ts: 1.649314776603171e+09
```
However, when creating the assert file we dump float and int as value
```yaml
results["s01-parse"]["crowdsecurity/caddy-logs"][0].Evt.Unmarshaled["caddy"]["ts"] == 1629470295.021685
```

So when a user pushes the test to the hub they will never match we can update the assert dump to print floats as 16 bit for my use case above. However, we may run into the issue again if a user has a massive int

```go
case float64:
	ret = append(ret, fmt.Sprintf(`%s == %s`+"\n", ekey, strconv.FormatFloat(val, 'e', 16, 64)))
```

Parsers.assert and yaml unmarshal should print values that will match 

Fork my hub branch https://github.com/crowdsecurity/hub/pull/806 and run

```
sudo TZ=UTC cscli hubtest run caddy-logs
```

_No response_

### Crowdsec version

<details>

```console
$ cscli version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
# paste output here

# On Windows:
# paste output here
</details>


### Config show

<details>

```console
$ cscli config show
# paste output here
```

</details>

### Prometheus metrics

<details>

```console
$ cscli metrics
# paste output here
```

</details>

<details>

</details>

buixor:
- fix #2434 : Use `Abs(a,b) < 1e-6` approach (IEEE 754) to allow approx float comparison
  - Move the initialization of expr helpers

### DIFF
--- a/pkg/exprhelpers/expr_lib.go
+++ b/pkg/exprhelpers/expr_lib.go
@@ -419,6 +419,13 @@ var exprFuncs = []exprCustomFunc{
 			new(func() (string, error)),
 		},
 	},
+	{
+		name:     "FloatApproxEqual",
+		function: FloatApproxEqual,
+		signature: []interface{}{
+			new(func(float64, float64) bool),
+		},
+	},
 }
 
 //go 1.20 "CutPrefix":              strings.CutPrefix,
--- a/pkg/exprhelpers/helpers.go
+++ b/pkg/exprhelpers/helpers.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/base64"
 	"fmt"
+	"math"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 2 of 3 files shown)


LaurenceJJones (approved):
LGTM

PR #2495, opened 2023-09-28
Improved expr debugger

buixor:
- New expr debugger approach
 - [x] Tests
 - [x] Kill old patch/visit approach


sabban (approved):
lgtm, seems like a great improvement for debugging

AlteredCoder (approved):

Issue #2499, opened 2023-09-29
Crowdsec 1.5.4 invalid memory address or nil pointer dereference (ProxMox LXC)


Mirabis:
### What happened?

Crowdsec crashes whenever I try to start it since moving to 1.5.4.
<details>

```console
time="29-09-2023 15:47:00" level=error msg="crowdsec - goroutine lapi/pullFromAPIC crashed : runtime error: invalid memory address or nil pointer dereference"
time="29-09-2023 15:47:00" level=error msg="please report this error to https://github.com/crowdsecurity/crowdsec/"
time="29-09-2023 15:47:00" level=error msg="stacktrace/report is written to /tmp/crowdsec-crash.1951062260.txt : please join it to your issue"
time="29-09-2023 15:47:00" level=fatal msg="crowdsec stopped"
```

</details>

**Stacktrace/report:**
<details>

```console
root@crowdsec:/var/log# cat /tmp/crowdsec-crash.1951062260.txt
error : runtime error: invalid memory address or nil pointer dereference
version: v1.5.4-debian-pragmatic-amd64-e4dcdd25728b914823525f1efabf18d5c454902b
BuildDate: 2023-09-20_12:15:26
GoVersion: 1.20.5
goroutine 118 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
github.com/crowdsecurity/go-cs-lib/trace.WriteStackTrace({0x1c819a0, 0x3072fc0})
        github.com/crowdsecurity/go-cs-lib@v0.0.4/trace/trace.go:26 +0x24d
github.com/crowdsecurity/go-cs-lib/trace.CatchPanic({0x1f069bd, 0x11})
        github.com/crowdsecurity/go-cs-lib@v0.0.4/trace/trace.go:41 +0xfa
panic({0x1c819a0, 0x3072fc0})
        runtime/panic.go:884 +0x213
net.networkNumberAndMask(0xc000376272?)
        net/ip.go:498
net.(*IPNet).Contains(0xc000376272?, {0xc000e94000, 0x10, 0xc000a07958?})
        net/ip.go:522 +0x27
github.com/crowdsecurity/crowdsec/pkg/apiserver.(*apic).whitelistedBy(0xc000002000, 0xc000a07a08?)
        github.com/crowdsecurity/crowdsec/pkg/apiserver/apic.go:629 +0xed
[...]

PR #2501, opened 2023-09-29
fix issue #2499 - nil dereference while using capi whitelists

mmetc:


buixor (approved):

mmetc:
This should fix the issue but I'm not sure how to reproduce it to verify that it's gone.
I should reproduce and add a test before merging.

Issue #1337, opened 2022-03-10
Bug/ Postoverflow assigned metadata not visible to notifications


RaphMad:
**Describe the bug**
When using a postoverflow parser performing a _statics: - meta_ assignment, the assigned value is not present in the object passed for notification rendering.

When using the same parser in the s02-enrich stage, the value is present for notifications.

**To Reproduce**
1. Install a postoverflow parser performing a meta-assignment, e.g.: https://github.com/crowdsecurity/hub/blob/master/postoverflows/s00-enrich/crowdsecurity/rdns.yaml
2. Trigger any scenario
3. Verify in the logfile that the value is assigned:
`time="10-03-2022 12:28:38" level=debug msg=".Meta[reverse_dns] = '089144197220.atnat0006.highway.a1.net.'" id=shy-night name=crowdsecurity/rdns stage=s00-enrich`

**Expected behavior**
When rendering the generated notificaiton via `{{ . | toJson }}`, a `Meta` entry with key `reverse_dns` and value `089144197220.atnat0006.highway.a1.net.` should be present.

**Technical Information (please complete the following information):**
 - OS: Debian/dockerized
 - Versionv 1.3.1

**Additional context**
When moving the postoverflow parser into s02-enrich, the meta value _is_ visible within the output of `{{ . | toJson }}`.

LaurenceJJones:
Hey sorry, about literally no response. I am going to try to replicate this now.

LaurenceJJones:
i can confirm this behaviour still exists in latest version. We are tracking down the issue.

LaurenceJJones:
Issue exists in how postoverflow alert works.

When overflown the bucket created the overflow state and when we pass to postoverflow it only keep the meta/enricher/parsed within the event created for that state only its not passed to the Overflown alert.

Now this gets tricky as where do we inject the meta data? as events are the metadata within the poured event and postoverflow does not really class as an event. There is a Metadata property on the base alert which has now been used for alert context, however, we cannot inject the data here as it will be forwarded to CAPI and has to be a string in the forma of a slice.

@buixor thoughts?

PR #2502, opened 2023-09-30
[postoverflow] Reinject meta into overflown state

LaurenceJJones:
#1337

This one is difficult to find an appropriate fix..... since Meta is converted to events on "overflow" anything set in meta in postoverflow stage is never carried beyond. This PR re injects all meta objects across all events, to make sure the user has it.

I am primarily looking for feedback as I couldnt really see any other property without creating a new one but we already have too many "meta" properties.


sabban (commented):
lgtm

sabban (commented):
lgtm

sabban (approved):
lgtm

LaurenceJJones:
Closing for now due to concerns that re injecting the meta without any consideration from user perspective could cause confusion as we should control which alert in the post overflown is injected.

PR #2517, opened 2023-10-04
[whitelist] metrics improvement

### LINKED ISSUE #2462: [enhancement] Whitelist metrics

Whilst using crowdsec we provide various metrics about parsing/acquisition/buckets , however, whitelist are very ambiguous to the user as currently there just a `parsed` and `non-parsed` metric. 

To which the user will always see 100% parsed since expr/ip/cidr are not taken into account within this metric.

So my first thought was to include this metric expr/ip/cidr to drive it, however, still `parsed` and `non-parsed` isnt intuative to the user. So I purpose we create a dedicated whitelist section and the current acquisition metric should exclude whitelist files since like I said `parsed` and `non-parsed` are just not useful.

Here is an example I thought, however, when it comes to expression it would be quite messy.

name| whitelisted|NW|
--------|-----------------|-----|
crowdsecurity/whitelist:192.168.0.1/24 | 2 | 1598|
my/personal-whitelist:1.1.1.1 | 1400 | 200 |
my/personal-expression:1 | 0 | 1600 |
my/personal-expression:2 | 0 | 1600 |
my/personal-expression:3 | 1 | 1599 |

So the top two example are self explanatory , however, expressions are difficult since we dont want to show the full expression line so I purpose we just show the index of the expression that passed. So in my example expression index 3 has passed.

If anyone else has any other ideas I am open to it I just dont know the best way to present the data without having a singe metric for the whitelist file

EDIT: OR we only increase the whitelist metrics and not whitelist should just be ignored since if the metric is not there I dont think it will show the user any value

name| count |
--------|-----------------|
crowdsecurity/whitelist:192.168.0.1/24 | 2 |
my/personal-whitelist:1.1.1.1 | 1400 |
my/personal-expression:3 | 1 |

/kind enhancement

### Why is this needed?

Provide better whitelist metrics for support as well as users as current parser metrics do not give the level of information needed to debug

LaurenceJJones:
Fix #2462
Based on #2422

We split out whitelist metrics from parsers, and provide detail breakdown of each expression.

Mostly looking for feedback as I know on the metrics side, we still need to merge the above before moving forward with this one.

Example of metrics;

```
Whitelist Metrics:
╭───────────────────────────────────────────┬────┬──────╮
│                 Whitelist                 │ WL │ NoWL │
├───────────────────────────────────────────┼────┼──────┤
│ crowdsecurity/jellyfin-whitelist:[0]      │ -  │ 4    │
│ crowdsecurity/nextcloud-whitelist:[0]     │ 4  │ -    │
│ crowdsecurity/whitelists:[10.0.0.0/8]     │ -  │ 4    │
│ crowdsecurity/whitelists:[127.0.0.1]      │ -  │ 4    │
│ crowdsecurity/whitelists:[172.16.0.0/12]  │ -  │ 4    │
│ crowdsecurity/whitelists:[192.168.0.0/16] │ -  │ 4    │
│ crowdsecurity/whitelists:[::1]            │ -  │ 4    │
╰───────────────────────────────────────────┴────┴──────╯
```

Currently I am using the actual index but for users we may just want to increase the digit +1 so they know if was the first expression that passed for example.


mmetc:
Hi @LaurenceJJones what's the state on this?
Still aimed at 1.5.6? Can I help address the conflicts?
I ask because I'd like to refact and address https://github.com/crowdsecurity/crowdsec/issues/2618

LaurenceJJones:
Closed in favor of #2813

PR #2531, opened 2023-10-10
"cscli bouncers add": increase key size, deprecate and ignore --length option 

mmetc:
the switch to base64 made the keys shorter (24 characters), this PR increases their size to 32 bytes, 42 chars once encoded

Also deprecate the --length option, users can already provide a key


mmetc:
we'll just deprecate / ignore the -l flag

buixor (approved):

PR #2540, opened 2023-10-12
fix null deref in cti calls if key is empty

### LINKED ISSUE #2537: CrowdsecCTI Nil defer
https://github.com/crowdsecurity/crowdsec/blob/3b1563a5386b423faba4a5fe7650881d151c829b/pkg/exprhelpers/crowdsec_cti.go#L78-L91

We must reorder these checks, and not use the logger on cti client if cti client is nil as it just crashes

buixor:
fix #2537


LaurenceJJones:
@buixor Seems a bats test is also broken

LaurenceJJones (approved):
LGTM

PR #2597, opened 2023-11-16
Refact cwhub: simplify enable/disable/download

mmetc:
- the result of filepath.Join is already Cleaned
 - no need to log the creation of parentDir
 - filepath.Abs() only returns error if the current working directory has been removed


sabban (approved):
lgtm

PR #2599, opened 2023-11-16
add new env var to enable console_management

he2ss:


docker/docker_start.sh
```diff
@@ -353,6 +353,11 @@ for BOUNCER in $(compgen -A variable | grep -i BOUNCER_KEY); do
     fi
 done
 
+if [ "$ENABLE_CONSOLE_MANAGEMENT" ] != "" ]; then
```
mmetc, line 356: there is an extra ]

blotus (approved):
Merge after #2601 is merged

PR #2633, opened 2023-12-01
cscli config show: pretty print struct output

mmetc:
This PR allows a user to discover field names in the run-time configuration without searching the source code.

The only issue I see is that some custom types like logrus.Level are not marshaled, i.e. will print "4" instead of "info" like the version with "%v". But it handles pointers and even recursive structures, so it can be a more readable (if less detailed) alternative to spew.

An alternative to litter would be kylelemons/godebug, which is less verbose (does not print struct names which are useless in this case) but has the same issue with LogLevel.


sabban (approved):
lgtm

Issue #2142, opened 2023-03-25
Support custom CA in HTTP notification


Neraud:
### What would you like to be added?

/kind enhancement

It would be great to be able to provide a custom CA (like `curl --cacert`) for the HTTP notification.

### Why is this needed?

In situations where the target HTTPS service uses a certificate that isn't signed by a known CA, the only workaround is to completly disable TLS verification with `skip_tls_verification: true`

It's a better practice to allow the user to configure a custom CA and validate the certificate.

LaurenceJJones:
Will be included in next release, however, there is a caveat since the plugins run as a low privilege user you must allow readable access to the configured user/group.

PR #2634, opened 2023-12-04
[http plugin] Add capath, certpath, keypath to load custom certs

LaurenceJJones:
Implement  #2142

I can replicate these changes across to other plugins that use http base stuff, However, I wanted to draft a PR to ask should we include the helpers function in `go-cs-lib` as then we can have a central place tag @mmetc


mmetc:
Mmmh.. I think the function is simple and idiomatic enough that it doesn't justify the added dependency.

mmetc (approved):

PR #2657, opened 2023-12-13
cscli machines: lint + write output to stdout instead of log

mmetc:
this allows us to run it with --error in the postinst scripts and avoid duplicate warnings about wal, capi etc.


sabban (approved):
lgtm

PR #2677, opened 2023-12-21
CI: bump golangci-lint run to 1.55, update defaults

mmetc:
The plan is to enable linters first, then fix issues types one by one.


sabban (approved):
lgtm

PR #2691, opened 2024-01-02
enable context sending when you enroll into the console

buixor:


AlteredCoder:
Can you please add a proper message after the enrollment to say what console options are enabled ?
Also, if we can log (if it is not already the case) what console options are enabled when crowdsec starts?

PR #2693, opened 2024-01-03
minor waf fixes

buixor:
- do not bail out on missing host header
- add `eq` operator
- support negative match


LaurenceJJones (approved):
LGTM but let me know if you plan to use this PR for the others items

PR #2698, opened 2024-01-05
[appsec] implement count transformation

buixor:
the `count` transformation transform a selection into the number of items


blotus (changes requested):
LGTM, but please add some tests :)

blotus (approved):

PR #2721, opened 2024-01-11
log death reason of file reader if available

### LINKED ISSUE #2718: [File] Log the underlying reason for dead reader
Dead reader errors are impossible to debug because we do not log the actual underlying reason that comes from Tail package.

We could do the following:

```go
		case <-tail.Dying(): //our tailer is dying
			logger.Warningf("File reader of %s died. Reason: %s", tail.Filename, tail.Err())
			t.Kill(fmt.Errorf("dead reader for %s, reason: %s", tail.Filename, tail.Err()))
			return fmt.Errorf("reader for %s is dead", tail.Filename)
```

Current code just states dead reader:

https://github.com/crowdsecurity/crowdsec/blob/437a97510ae8b60e1f7cb4ce6cc23ad82ea01f98/pkg/acquisition/modules/file/file.go#L455-L458

blotus:
Closes #2718

### DIFF
--- a/pkg/acquisition/modules/file/file.go
+++ b/pkg/acquisition/modules/file/file.go
@@ -453,9 +453,14 @@ func (f *FileSource) tailFile(out chan types.Event, t *tomb.Tomb, tail *tail.Tai
 			}
 			return nil
 		case <-tail.Dying(): //our tailer is dying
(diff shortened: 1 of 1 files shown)


LaurenceJJones (approved):
LGTM

PR #2726, opened 2024-01-11
Ignore missing console/context.yaml if not explicitly required by config.yaml

mmetc:
before: always emit a warning if console/context.yaml does not exist

after: use the file if there (it's the default location) but only warn that it's missing if it's explicitly set


sabban (approved):
lgtm

PR #2740, opened 2024-01-15
Add original http request to hooks

AlteredCoder:
Add original http request to pre_eval, onmatch and post_eval.


blotus (approved):

LaurenceJJones:
The original request needs to be modifiy to match the incoming `x-crowdsec` headers are else `r.Host` will be the `AppSec` server hostname instead of the _actual_ requested hostname of the server.

LaurenceJJones (changes requested):
The original request needs to be modifiy to match the incoming `x-crowdsec` headers are else r.Host will be the `AppSec` server hostname instead of the actual requested `hostname`of the server.

blotus (approved):

LaurenceJJones (approved):
LGTM

PR #2751, opened 2024-01-17
post-install: reduce verbosity

mmetc:
this removes the initial WAL warning too, is that ok or should I put it back?

Otherwise, we don't see anymore:

```
find: '/usr/local/openresty/nginx/logs': No such file or directory
...
WARN[2024-01-17T11:24:37Z] can't load CAPI credentials from '/etc/crowdsec/online_api_credentials.yaml' (missing login field)
INFO[2024-01-17T11:24:37Z] push and pull to Central API disabled       
```


sabban (approved):
lgtm

PR #2756, opened 2024-01-18
pkg/hubtest: extract methods + consistent error handling

mmetc:
- extract methods from loops to increase maintainability
 - treat errors in the same way for all item types


sabban (approved):
lgtm

PR #2760, opened 2024-01-19
Support `console options` in `console enroll`

buixor:
w/ a few simplifications

### DIFF
--- a/pkg/csconfig/console.go
+++ b/pkg/csconfig/console.go
@@ -19,6 +19,13 @@ const (
 )
 
 var CONSOLE_CONFIGS = []string{SEND_CUSTOM_SCENARIOS, SEND_MANUAL_SCENARIOS, SEND_TAINTED_SCENARIOS, SEND_CONTEXT, CONSOLE_MANAGEMENT}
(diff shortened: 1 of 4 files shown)


pkg/csconfig/console.go
```diff
@@ -19,6 +19,13 @@ const (
 )
 
 var CONSOLE_CONFIGS = []string{SEND_CUSTOM_SCENARIOS, SEND_MANUAL_SCENARIOS, SEND_TAINTED_SCENARIOS, SEND_CONTEXT, CONSOLE_MANAGEMENT}
+var CONSOLE_CONFIGS_HELP = map[string]string{
```
LaurenceJJones, line 22: Maybe we dont need each help option to say `to the console`? as when the option is used the context is already around the console?

buixor: some are "from console", I prefered verbosity to lack of clarity, but please suggest alternate phrasing, I just copied the existing ones :<

AlteredCoder (approved):

AlteredCoder (approved):

Issue #2757, opened 2024-01-19
[http] Unix socket support


LaurenceJJones:
### What would you like to be added?

With my recent fascination in Unix sockets I have been working on adding support to some projects I use, however, our http notification plugin does not support `unix:` URL's

/kind enhancement

### Why is this needed?

If users have a service that support unix sockets we should allow user to push notification across them. We should just add this support to http notification as that what it is another plugin is not needed.

PR #2764, opened 2024-01-20
[http] Add unix socket option

LaurenceJJones:
Implement #2757 

Add unix socket option to http plugin, we have to use this in conjunction with URL parameter as we dont know which path the user wants so if they would like to communicate over unix socket they need to use both, however, the hostname can be whatever they want. (All communication will be over the unix socket but we dont know the path or scheme and rather not add specific arguments just for that when they can define within URL param)

Example of pushing to Gotify over unix socket (Dont worry about token as it is just a local version)

<details>

```
type: http
name: http_default
log_level: info

format: |
  {{ range . -}}
  {{ $alert := . -}}
  {
    "extras": {
      "client::display": {
      "contentType": "text/markdown"
    }
  },
  "priority": 3,
  "title": "test from crowdsec!",
  "message": "message!"
  }
  {{ end -}}

url: "http://_/message?token=AxHH7ZX1ttm-ZWc"
unix_socket: /opt/gotify.sock

# Any of the http verbs: "POST", "GET", "PUT"...
method: POST

headers:
  Content-Type: application/json
#   Authorization: token 0x64312313
```

</details>

```bash
$ sudo cscli notifications test http_default
WARN[2024-01-20T14:32:49Z] You are using sqlite without WAL, this can have a performance impact. If you do not store the database in a network share, set db_config.use_wal to true. Set explicitly to false to disable this warning.
INFO[2024-01-20T14:32:49Z] loaded capi whitelist from /etc/crowdsec/capi_whitelists.yaml: 1 IPs, 0 CIDRs
DEBU[0000] starting plugin                               args="[/usr/local/lib/crowdsec/plugins/notification-http]" path=/usr/local/lib/crowdsec/plugins/notification-http
DEBU[0000] plugin started                                path=/usr/local/lib/crowdsec/plugins/notification-http pid=61824
DEBU[0000] waiting for RPC address                       path=/usr/local/lib/crowdsec/plugins/notification-http
DEBU[0000] using plugin                                  version=1
TRAC[0000] waiting for stdio data
INFO[0000] Using socket '/opt/gotify.sock'               @module=http-plugin
INFO[2024-01-20T14:32:49Z] registered plugin http_default
INFO[2024-01-20T14:32:49Z] pluginTomb dying
INFO[0000] received signal for http_default config       @module=http-plugin
INFO[2024-01-20T14:32:49Z] killing all plugins
DEBU[0000] received EOF, stopping recv loop              err="rpc error: code = Unavailable desc = error reading from server: EOF"
INFO[0000] plugin process exited                         path=/usr/local/lib/crowdsec/plugins/notification-http pid=61824
DEBU[0000] plugin exited
```

image


LaurenceJJones:
Did you manage to test and review @mmetc

mmetc (approved):

PR #2765, opened 2024-01-20
[cscli] Add tab completion to notifications [test, inspect]

LaurenceJJones:
Sometimes you might not know of the top of your head what notification you can test, this PR adds tab completion via `cscli completion`

Examples:
No filter
image

Filter
image

Update: Split function to reusable and pass to inspect command as that also needs a valid plugin config name


buixor (approved):
lgtm and tested, seems ok

PR #2767, opened 2024-01-22
Various small bats fixes

sabban:
* fix the nobody/nogroup being nobody/nobody on centos
* fix the '//' not matching '/' in 04_capi

### DIFF
--- a/test/bats/01_cscli.bats
+++ b/test/bats/01_cscli.bats
@@ -116,8 +116,8 @@ teardown() {
 
     # complex type
     rune -0 cscli config show -o json --key Config.PluginConfig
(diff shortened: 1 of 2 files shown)


test/bats/01_cscli.bats
```diff
@@ -115,13 +115,9 @@ teardown() {
     assert_output "&false"
 
     # complex type
-    rune -0 cscli config show --key Config.PluginConfig
-    assert_output - <<-EOT
-	&csconfig.PluginCfg{
```
mmetc, line 120: Actually, the point of this was to test how it pretty-printed a structure, it can be something other than pluginconfig

PR #2213, opened 2023-05-20
local api unix socket support

cyberb:
Initial unix socket support, tested on my running instance.

config.yaml
```
api:
  server:
    listen_uri: /var/snap/crowdsec/current/crowdsec.sock
```

Also
local_api_credentials.yaml
```
url: /var/snap/crowdsec/current/crowdsec.sock
```

Will check if there are http test server tests so I can add a unix socket unit test.

Some comments:

pkg/csconfig/api.go contains too many structs in one file, it is better to split into multiple files.

PR #2770, opened 2024-01-22
LAPI: local api unix socket support

mmetc:
This started as https://github.com/crowdsecurity/crowdsec/pull/2213 by @cyberb but I refactored some parts first, as a consequence I could not keep a clean commit history of his contribution, of which I am very sorry.

On the plus side, the option is now implemented as a separate configuration field: `.api.server.listen_socket = /path/to/unix.sock` and does not interfere with TCP sockets, with or without TLS.


LaurenceJJones:
When using Nginx as an upstream proxy via `http://unix:/` `proxy_pass` directive we need to allow the user to get the remote IP from the headers. Now Gin does have a `trusted_proxies` function that we do use, however, when nginx passes t the unix socket the `RemoteAddr` is set to `@` for `Nil`.

I don't know the best way to move forward on this 🤷🏻

mmetc:
@LaurenceJJones does this mean this PR is not useful, or that it will need changes that are not backward compatible if we fix the proxy issue later?
If it's not the case, we can review/merge, test until 1.6.1 and look for a fix

LaurenceJJones:
> @LaurenceJJones does this mean this PR is not useful, or that it will need changes that are not backward compatible if we fix the proxy issue later? If it's not the case, we can review/merge, test until 1.6.1 and look for a fix

We can merge, the only issue is that machines/bouncers will not have the correct IP address if the request was from an upstream proxy

pkg/apiserver/controllers/v1/utils.go
```diff
@@ -25,10 +27,24 @@ func getBouncerFromContext(ctx *gin.Context) (*ent.Bouncer, error) {
...
+		return strings.HasPrefix(localAddr.Network(), "unix")
+	}
+
+	return false
+}
+
 func (c *Controller) AbortRemoteIf(option bool) gin.HandlerFunc {
 	return func(gctx *gin.Context) {
+		if !option {
+			return
```
LaurenceJJones, line 41: Dont we need to call gctx.Next() before the return?

pkg/apiserver/controllers/v1/utils.go
```diff
@@ -25,10 +27,24 @@ func getBouncerFromContext(ctx *gin.Context) (*ent.Bouncer, error) {
...
+		}
+		if isUnixSocket(gctx) {
+			return
```
LaurenceJJones, line 44: same

```
Dont we need to call gctx.Next() before the return?
```

pkg/apiserver/controllers/v1/utils.go
mmetc: No, calling Next() explicitly is only required if you need to do some processing _after_ the request is served. Otherwise if not stopped, the next middleware is called. And we can remove it from the ApiKey middleware too.

cmd/crowdsec-cli/machines.go
```diff
@@ -318,7 +318,7 @@ func (cli *cliMachines) add(args []string, machinePassword string, dumpFile stri
 		if clientCfg != nil && clientCfg.Credentials != nil && clientCfg.Credentials.URL != "" {
 			apiURL = clientCfg.Credentials.URL
 		} else if serverCfg != nil && serverCfg.ListenURI != "" {
```
LaurenceJJones, line 320: `serverCfg.ListenURI != "" || serverCfg.ListenSocket != ""` ?

pkg/apiserver/middlewares/v1/api_key.go
```diff
@@ -139,8 +139,13 @@ func (a *APIKey) MiddlewareFunc() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		var bouncer *ent.Bouncer
 
+		clientIP := c.ClientIP()
+		if clientIP == "" && isUnixSocket(c) {
```
LaurenceJJones, line 143: But what if the request is proxied to unix socket from a web server? we need to set the remote addr on the request before c.ClientIP() is called.

https://github.com/gotify/server/blob/46281d6a51dc0ca6a2ac13f81db89224b990d876/router/router.go#L34-L39

pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -213,18 +222,23 @@ func (j *JWT) Authenticator(c *gin.Context) (interface{}, error) {
 		}
 	}
 
+	clientIP := c.ClientIP()
+	if clientIP == "" && isUnixSocket(c) {
```
LaurenceJJones, line 226: But what if the request is proxied to unix socket from a web server? we need to set the remote addr on the request before c.ClientIP() is called.

https://github.com/gotify/server/blob/46281d6a51dc0ca6a2ac13f81db89224b990d876/router/router.go#L34-L39

cmd/crowdsec-cli/machines.go
mmetc: good catch! see last commit

pkg/apiserver/middlewares/v1/jwt.go
mmetc: so it's the proxy that sets "@" as remoteaddr? and I have to copypasta the 3-line middleware from gotify, right?

mmetc: Guess it only works with g.ForwardedByClientIP = true

buixor (approved):
lgtm

PR #2776, opened 2024-01-24
refact "cscli bouncers"

mmetc:
There are a few patterns there that, if applied consistently to cmd/crowdsec-cli, would enhance maintainability.

 - encapsulate reference to global state (even csConfig) via command constructors
 - don't export functions if not needed
 - it should be possible to split crowdsec-cli in sub-packages (cmd/crowdsec-cli/bouncers) I didn't do it here because we ought to do it with the table helpers first
 - no need to parse each flag, or to store them together, or to share them as global vars
 - don't ask for more configuration than required (db vs lapi server)
 - don't over-indent or overuse closures
 - parse flags directly to their type (for time.Duration)
 - don't repeat information in user facing messages (i.e. error doing thing with X: tried to blahblah X: failed because X is not Y)


mmetc:
Actually - we do need to load LAPI configuration because that's how we check the command is running on the right node. But I'll have to nerf the logs in this case

sabban (approved):
lgtm

PR #2789, opened 2024-01-26
Progress bar for one shot acquisition (time machine -- cold logs)

mmetc:
This also disables prometheus metrics in one shot mode because in most cases the port is already taken by LAPI.


mmetc:
I had to disable the message

log.Infof("%d existing buckets", leaky.LeakyRoutineCount)

when the count is 0. Is that ok?

mmetc:
The progress bar is not disabled in case of redirection. We might check isatty?

PR #2809, opened 2024-02-02
appsec: get the original UA from headers

### LINKED ISSUE #2787: [AppSec] User user-agent is not forwarded to AppSec
### What happened?

The AppSec component does not set the _actual_ useragent of the remote user, so any of the sec rules that try to match on useragent may currently produce false negatives.

Send remote useragent as a header and set correctly within the parsed request stage

Dump request sent to inband rules and you will see lua resty useragent as the remote users

_No response_

### Crowdsec version

<details>

```console
$ cscli version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
# paste output here

# On Windows:
# paste output here
</details>


### Config show

<details>

```console
$ cscli config show
# paste output here
```

</details>

### Prometheus metrics

<details>

```console
$ cscli metrics
# paste output here
```

</details>

<details>

</details>

blotus:
Closes #2787


LaurenceJJones (approved):
LGTM

PR #2846, opened 2024-02-14
refact pkg/apiclient

mmetc:


pkg/apiclient/auth_jwt.go
```diff
@@ -130,27 +130,43 @@ func (t *JWTTransport) refreshJwtToken() error {
...
-	// In a few occasions several goroutines will execute refreshJwtToken concurrently which is useless and will cause overload on CAPI
-	// we use a mutex to avoid this
-	// We also bypass the refresh if we are requesting the login endpoint, as it does not require a token, and it leads to do 2 requests instead of one (refresh + actual login request)
+func (t *JWTTransport) needsTokenRefresh() bool {
+	return t.Token == "" || t.Expiration.Add(-time.Minute).Before(time.Now().UTC())
+}
+
+// prepareRequest returns a copy of the  request with the necessary authentication headers.
+func (t *JWTTransport) prepareRequest(req *http.Request) (*http.Request, error) {
+	req = cloneRequest(req)
```
LaurenceJJones, line 139: Do we _need_ to clone the request?

mmetc: I thought, we already do it for api key, and we have a retry-backoff loop so we reuse it. Sure @blotus knows more

pkg/apiclient/resperr.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(e.Errors) > 0 {
+		err += fmt.Sprintf(" (%s)", e.Errors)
+	}
+
+	return err
+}
+
+// CheckResponse verifies the API response and builds an appropriate Go error if necessary.
+func CheckResponse(r *http.Response) error {
+	if c := r.StatusCode; 200 <= c && c <= 299 || c == 304 {
```
LaurenceJJones, line 29: Shouldnt this be `200 >= c && c <= 299 || c == 304`

So c is between 200's?

LaurenceJJones (approved):
LGTM

PR #2856, opened 2024-02-22
refact pkg/apiserver (auth helpers)

mmetc:


mmetc:
Required for the metrics. It would probably make sense to move the two helpers to middlewares/v1, since that's where they are appended to context.

blotus (approved):

PR #2868, opened 2024-03-04
Add patterns_dir configuration option

kampka:
Currently the patterns directory is implicitly set to `/etc/crowdsec/patterns`. This PR make that configurable while keeping the default if not set.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -14,6 +14,7 @@ config_paths:
   index_path: /etc/crowdsec/hub/.index.json
   notification_dir: /etc/crowdsec/notifications/
   plugin_dir: /usr/local/lib/crowdsec/plugins/
+  #pattern_dir: /etc/crowdsec/patterns/
 crowdsec_service:
   #console_context_path: /etc/crowdsec/console/context.yaml
   acquisition_path: /etc/crowdsec/acquis.yaml
@@ -27,7 +28,7 @@ db_config:
   type: sqlite
   db_path: /var/lib/crowdsec/data/crowdsec.db
   #max_open_conns: 100
-  #user: 
+  #user:
   #password:
   #db_name:
   #host:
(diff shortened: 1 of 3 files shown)


LaurenceJJones:
I was just wondering why this would be needed? Other than the obvious that patterns need to be in a different directory.

kampka:
I'm building a Crowdsec module for NixOS, an operating system built around the concepts of immutability and reproducibility. 
To facilitate that, static configuration does not reside in the "mutable" space of the file system, but in the immutable nix store.
This would ideally also be true for the patterns used by crowdsec.
In that case, the patterns would be living under `/nix/store/<reproducible-hash-value>-crowdsec/config/patterns` for example.
Given that NixOS does have mechanisms to work around these challenges, eg. symlinking the patterns dir from the nix store to `/etc`, but it would be cleaner if that would not be necessary. This change is mostly for cosmetic purposes.

LaurenceJJones:
```
time="2024-03-05T09:48:48Z" level=info msg="Loading grok library /etc/crowdsec/patterns"
time="2024-03-05T09:48:48Z" level=fatal msg="crowdsec init: while loading parsers: failed to load parser patterns : open /etc/crowdsec/patterns: no such file or directory"
```

kampka:
I've added the trailing `/`to the path.
Sadly, the test suite is not really trivial to get running, so this is a best guess.
I could also imagine there is now an `mkdir` required in one of the test setup scripts, but that is hard to pinpoint.

mmetc:
Functional tests are run in a local environment, and don't use /etc/crowdsec, so they won't find the directory. This must be fixed in test/lib/config, where the two fixtures are prepared: one for testing from sources, one for testing a .deb, or .rpm package.

mmetc:
see for example https://github.com/crowdsecurity/crowdsec/blob/d8877a71fc1faac18539cd29fa736deccaf4ea92/test/lib/config/config-local#L91

the config-global fixture should be ok. No need to append '/' to the path IMO

mmetc:
another option would be to just remove patterns_dir from config.yaml, and by using the default there is no need to change the fixture script

kampka:
Sure thing, fixed. Just FYI, the term `PatternsDir` is already present in the Windows installer, where I picked it from. Just in case you'd want to make that uniform at some point :)

kampka:
@mmetc needs another approval for the tests to run because of rebase, otherwise this should be good to merge? :pray:

PR #2876, opened 2024-03-07
bin/crowdsec: avoid writing errors twice when log_media=stdout

mmetc:
fix https://github.com/crowdsecurity/crowdsec/pull/2729 for level=fatal


sabban (approved):
lgtm

PR #2881, opened 2024-03-08
warn if user is using inotify to tail a symlink

### LINKED ISSUE #2604: Detect when file acquisition uses inotify on a symlink
If you configure crowdsec to watch a symlink, and the file behind the symlink gets rotated, crowdsec will stop tracking new logs from this file (this can easily happen when running in k8s).

We should at least log a warning if we detect inotify polling on a symlink, or force falling back to the stat polling method for this file.

blotus:
Closes #2604


sabban (approved):
seems to be the right logic.
Lgtm

PR #2885, opened 2024-03-12
acquisition : take prometheus level into account

buixor:
properly take into account the aggregation level of prometheus metrics in the acquisition

### DIFF
--- a/pkg/acquisition/acquisition.go
+++ b/pkg/acquisition/acquisition.go
@@ -185,7 +185,7 @@ func GetMetricsLevelFromPromCfg(prom *csconfig.PrometheusCfg) int {
 		return configuration.METRICS_FULL
 
 	}
-	if prom.Enabled == false {
+	if !prom.Enabled {
 		return configuration.METRICS_NONE
 	}
 	if prom.Level == "aggregated" {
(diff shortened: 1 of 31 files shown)


pkg/acquisition/acquisition.go
```diff
@@ -180,10 +180,30 @@ func LoadAcquisitionFromDSN(dsn string, labels map[string]string, transformExpr
...
 
+func GetMetricsLevelFromPromCfg(prom *csconfig.PrometheusCfg) int {
+	if prom == nil {
+		return configuration.METRICS_FULL
+
+	}
+	if !prom.Enabled {
+		return configuration.METRICS_NONE
+	}
+	if prom.Level == "aggregated" {
```
blotus, line 191: Maybe use a const string here ?

pkg/acquisition/acquisition.go
```diff
@@ -180,10 +180,30 @@ func LoadAcquisitionFromDSN(dsn string, labels map[string]string, transformExpr
...
+		return configuration.METRICS_AGGREGATE
+	}
+
+	if prom.Level == "full" {
```
blotus, line 195: Maybe use a const string here ?

PR #2917, opened 2024-03-27
Fix REQUEST_URI behavior + fix #2891

### LINKED ISSUE #2891: [AppSec] ParsedRequest.URL is the request sent to appsec endpoint
When we parse the request to a struct we provide `url` the parameter of `r.URL` which is the original request sent to appsec endpoint we should in fact use the parsed url

buixor:
- [x] Checked that it doesn't break existing vpatch rules (`cscli hubtest run --appsec --all`)
 - [x] Add the new helpers in the doc (`normalizepath`, `trim`, `urldecode`, `htmlentitydecode`)
 - [x] Fix some documentation (`contain` -> `contains`. Remove `hexdecode` that doesn't exist)

### DIFF
--- a/pkg/appsec/request.go
+++ b/pkg/appsec/request.go
@@ -30,6 +30,7 @@ type ParsedRequest struct {
 	Host                 string                  `json:"host,omitempty"`
 	ClientIP             string                  `json:"client_ip,omitempty"`
 	URI                  string                  `json:"uri,omitempty"`
+	Path                 string                  `json:"path,omitempty"`
 	Args                 url.Values              `json:"args,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/appsec/request.go
```diff
@@ -30,6 +30,7 @@ type ParsedRequest struct {
 	Host                 string                  `json:"host,omitempty"`
 	ClientIP             string                  `json:"client_ip,omitempty"`
 	URI                  string                  `json:"uri,omitempty"`
+	Path                 string                  `json:"path,omitempty"`
```
buixor, line 33: drop this

LaurenceJJones (approved):
LGTM tested with a basic remote http includes and it seems to be working now

PR #2919, opened 2024-03-29
Downgrade couldnt find in hub to infof instead of errorf

### LINKED ISSUE #2918: [crowdsec] lower log level to warn hub
We now have this log line when a user supplies a custom scenario or parser, however, the level of error is quite high can we just warn instead?

```
time="2024-03-29T06:30:13Z" level=error msg="scenario laurencejjones/ufw-block (/etc/crowdsec/scenarios/ufw-block-trigger.yaml) couldn't be find in hub (ignore if in unit tests)"
```

LaurenceJJones:
Fix #2918


mmetc:
Relaced by https://github.com/crowdsecurity/crowdsec/pull/3010

PR #2932, opened 2024-04-07
feat: File notification plugin

LaurenceJJones:
An official file notification plugin which is on par with the (not so third party) https://github.com/zbalkan/notification-file it accepts the same `file.yaml`, however, since I didnt want to use a workaround with logrus I implemented most of the features with standard go packages

There are many pitfalls that can arise with this method as in I cannot gaurantee the order of print, meaning a log that happens at one time may be appended to the log file before another. However, in the example to wazuh we add the timestamp to the output so wazuh can handle the actual timestamp log which is when it is formatted.

I did initially try to handle the queuing system within the plugin itself then write to a buffer on ticker the downside to that is the plugin broker does not have any context that there are items within the queue so I decided to not handle a queue and just print to the file as soon as the mutex is free.

The log rotation functionally is mostly written by copilot, I have tested it locally a bunch of times and on my system its fine. However, smaller system with limit compute may hit a wall when trying to compress so I made it an option.

LMK if we should improve it in any way ✨


cmd/notification-file/main.go
```diff
@@ -0,0 +1,253 @@
...
+	err = os.Remove(src)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func WriteToFileWithCtx(ctx context.Context, cfg PluginConfig, log string) error {
+	FileWriteMutex.Lock()
```
LaurenceJJones, line 168: Need to check with team as I dont even know if mutexs are even needed tested without and it was "fine", however, I dont have a high load test env and when reading the code that sends to plugin channel it inside a go routine so could be written multiple times 🤷🏻

cmd/notification-file/main.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return err
+	}
+	defer dstFile.Close()
+
+	// Create a gzip writer
+	gw := gzip.NewWriter(dstFile)
+	defer gw.Close()
+	// Read the source file and write its contents to the gzip writer
+	b, err := io.ReadAll(srcFile)
```
mmetc, line 149: guess we can avoid reading the whole file in memory with io.Copy

buixor (approved):

PR #2935, opened 2024-04-10
cscli support: include stack traces

mmetc:


pkg/trace/trace.go
```diff
@@ -0,0 +1,168 @@
...
+	log.Errorf("crowdsec - goroutine %s crashed: %s", component, r)
+	log.Error("please report this error to https://github.com/crowdsecurity/crowdsec/issues")
+
+	filename, err := tk.writeStackTrace(r)
+	if err != nil {
+		log.Errorf("unable to write stacktrace: %s", err)
+	}
+
+	log.Errorf("stacktrace/report is written to %s: please join it to your issue", filename)
+	log.Fatal("crowdsec stopped")
```
LaurenceJJones, line 167: Did we use to fatal at the end of a trace?

mmetc: yes if it's not in apiserver. other goroutines panics are not recoverable and we don't have a return path

LaurenceJJones (approved):

PR #2939, opened 2024-04-12
feat: support stdout in cscli support dump

### LINKED ISSUE #2938: [enhancement] cscli support dump support stdout
In many functions we have within `cscli` we support the flag `-f-` which mean print to stdout. It can be useful to support this functionality for `cscli support dump` since within a container we dont know if the user has access to the place where they dumped the support file.

The user can obviously use `docker cp` to move the file from the container to the host, however, if the user does not know this they can simply run

```
docker exec crowdssec cscli support dump -f- > /tmp/crowdsec.zip
```

LaurenceJJones:
Fix #2938 

Add the ability to print to stdout when the `-f-` is presented


cmd/crowdsec-cli/support.go
```diff
@@ -476,7 +476,13 @@ cscli support dump -f /tmp/crowdsec-support.zip
 				log.Fatalf("could not finalize zip file: %s", err)
 			}
 
-			err = os.WriteFile(outFile, w.Bytes(), 0o600)
+			if outFile == "-" {
+				_, err = os.Stdout.Write(w.Bytes())
+				outFile = "stdout"
```
mmetc, line 481: I suggest instead if we are on stdout, skip the line "Written zip file to %s"

mmetc (approved):

PR #2940, opened 2024-04-12
feat: collect pprofs on cscli support dump

### LINKED ISSUE #2937: [enhancement] cscli support dump collect pprofs
In many cases pprofs can be useful to see cpu/memory/goroutines which is an endpoint which is enabled on the LAPI side. When a user runs `cscli support dump` we should collect this information and store it within the zipped output.

/kind enhancement

LaurenceJJones:
Fix #2937 

Opening as draft since its not really working, I dont know why 🤷🏻 maybe @mmetc can help


cmd/crowdsec-cli/support.go
```diff
@@ -264,6 +268,50 @@ func collectAcquisitionConfig() map[string][]byte {
 	return ret
 }
 
+func collectPprofs() ([]byte, []byte, []byte) {
+	log.Info("Collecting pprof data")
+	client := &http.Client{
+		Timeout: 30 * time.Second,
```
LaurenceJJones, line 274: Most likely this can be removed

mmetc:
Works for me, and it takes a long time to download. I'll add a prompt to ask users if they want to include it (will do the same when I do logs). I won't set a timeout.

mmetc:
done in https://github.com/crowdsecurity/crowdsec/pull/2987

Issue #799, opened 2021-05-14
Improvement/Support JSON log output 


blotus:
We use [logrus](https://github.com/sirupsen/logrus) to handle all of crowdsec logs.
As logrus natively supports a [JSON formatter](https://github.com/sirupsen/logrus/blob/master/json_formatter.go), we should allow users to configure which output format they want (either plaintext or JSON).

nanikjava:
@blotus is the log configuration only for `crowdsec` or including the `cli` ?

LaurenceJJones:
@nanikjava To answer your question it would be crowdsec agent itself than the CLI since the cli already support `-o json` flag

sharkymcdongles:
This would be a nice feature to have because structured logging is quite important for a security tool.

ScreNN:
I support the idea

LaurenceJJones:
Close as completed as #2941 implemented this.

PR #2941, opened 2024-04-14
Add possibility to configure log format #799

victoredvardsson:
https://github.com/crowdsecurity/crowdsec/issues/799

logrus which is used as logging backend supports json formatter.

This change makes it possible to configure either text or json output.


pkg/types/utils.go
```diff
@@ -33,6 +33,14 @@ func SetDefaultLoggerConfig(cfgMode string, cfgFolder string, cfgLevel log.Level
 		if compress != nil {
 			_compress = *compress
 		}
+		
+		if LogFormat == "json" {
+			logFormatter = &log.JSONFormatter{TimestampFormat: time.RFC3339}
+		} else if LogFormat == "text" {
+			logFormatter = &log.TextFormatter{TimestampFormat: time.RFC3339, FullTimestamp: true, ForceColors: forceColors}
+		} else {
+			return fmt.Errorf("log format '%s' unknown", LogFormat)
```
LaurenceJJones, line 42: imo to not break backwards compat could we just default to `text` and issue a warning saying there was no value presented so default to `text`

```go
if LogFormat == "json" {
  logFormatter = &log.JSONFormatter{TimestampFormat: time.RFC3339}
} else {
   if LogFormat == "" {
       log.Warn("no log_format configured default to 'text' format")
   }
   logFormatter = &log.TextFormatter{TimestampFormat: time.RFC3339, FullTimestamp: true, ForceColors: forceColors}
}
```

LaurenceJJones: so if the user configures it will suppress the `warn`

LaurenceJJones: Also I think the above changes will fix the failing tests since it doesnt set to a value so you get a nil pointer

victoredvardsson: Sounds reasonable, i will update the code accordingly. Thanks for the feedback!

victoredvardsson:
Looks like all checks went through now 👍

mmetc:
I still need to add a formatter for stderr. We don't want fatal errors to be formatted in json there. Will do asap and merge for 1.6.2

LaurenceJJones:
> merge is postponed to 1.6.3 since I'd like to correctly bubble up errors from goroutines and 1.6.2 is too soon

We keeping `1.6.3` or moving to `1.6.4`?

mmetc:
Next one, I'd rather remove the remaining os.Exit and log.Fatal before implementing this

luis-guimaraes-exoawk:
Any news on this feature? It's very important for us

LaurenceJJones:
> Any news on this feature? It's very important for us

aiming for `1.6.5` which is the next release

LaurenceJJones:
Tested and works ✅ 

The only comments is that some log lines contain special characters or simply alot of information that we should use parameter instead of containing it all within the `msg` objects examples:

```
{"level":"debug","msg":"Response: HTTP/2.0 200 OK\r\nContent-Length: 876\r\nContent-Type: application/json\r\nDate: Sun, 22 Dec 2024 11:38:52 GMT\r\nVia: 1.1 e4faf88ed08954d3c5034fc535379dc6.cloudfront.net (CloudFront), 1.1 936c7ee6d0620cb8a766a50c04b3fa30.cloudfront.net (CloudFront)\r\nX-Amz-Apigw-Id: DMQL1EXWjoEEAHQ=\r\nX-Amz-Cf-Id: u0EUeH2MMqcUnhnI21rrAGYrOYyZnlWiZtBb9gPGVFRH_LD1qCGnew==\r\nX-Amz-Cf-Pop: LHR5-P7\r\nX-Amz-Cf-Pop: LHR50-P3\r\nX-Amzn-Requestid: 10f480e2-6468-404a-b365-5cc7ac8253a0\r\nX-Amzn-Trace-Id: Root=1-6767fa4b-0230272b064ebb435b2cd65e;Parent=1eaae8e82d2f732c;Sampled=0;Lineage=1:80359e24:0\r\nX-Cache: Miss from cloudfront\r\n\r\n{\"code\": 200, \"token\": \"<redacted>\", \"expire\": \"2024-12-22T19:38:52Z\"}","time":"2024-12-22T11:38:52Z"}
```
We should add a type parameter `type=request` or `type=response` and the response should be message 🤷🏻 

Also some of the special characters get turned into the html safe equivalent (`\u003e\u003e`):

```
{"level":"debug","msg":"starting one source 0/2 -\u003e\u003e *fileacquisition.FileSource","time":"2024-12-22T11:38:52Z"}
{"level":"debug","msg":"starting one source 1/2 -\u003e\u003e *fileacquisition.FileSource","time":"2024-12-22T11:38:52Z"}
```

We should just not use special characters in message objects as they add no value but this can be done in another PR as these are nothing to do with the actual PR just a side effect of legacy things.

mmetc:
agree on both points

 - unnecessary use of < and >
 - more structured logging

can we merge this and work on the above for 1.6.6 / 1.7.0 ?

LaurenceJJones (approved):
LGTM and tested 🦾

Issue #1435, opened 2022-04-11
Improvement/ Omnibus Ticket Docker Deployment Observations


Leopere:
# Starting from a Discord Thread: https://discord.com/channels/921520481163673640/961638950345244753 

## Initial Thread Message
Okay so now that I'm dunked into deploying for Docker I have some requests and complaints.

- [x] Registering bouncers is far too manual of a process and it blows.

- [x] acquis.yml should have a docker label or command invocation argument alternative so that acquisitions can be defined all in the same place because not all docker deployments support having configs.

- [x] it should if it doesn't currently recognize the difference between docker network sources. 
 

- [x] There seemed to be nearly no mention in the official docs that I could find for registering bouncers.

- [x] Not all containers log out to a file and it should be considered somehow to hook into Docker's logging mechanism.

**Is your feature request related to a problem? Please describe.**
My issue is just a series of observations I've come to with my attempts to deploy CrowdSec into a Docker Swarm Stack utilizing Traefik as a reverse proxy ingress route.

**Describe the solution you'd like**
In some cases Documentation improvements and in other cases maybe bouncer design guidelines/best practices to encourage a relatively uniform behavior when it comes to bouncers.

**Describe alternatives you've considered** (Optional) 
Discussed in Discord and scoured documentation.

**Example of what you imagine**
Starting a conversation as per requested by people in Discord to determine what direction to go with some of these features.

LaurenceJJones:
We have improved `Registering bouncers is far too manual of a process and it blows.` by providing extra ENV variables which allows you to provide multiple bouncers with keys.

Need some more information on why this is a point `it should if it doesn't currently recognize the difference between docker network sources.` ?

If top point is completed is this still a valid point `There seemed to be nearly no mention in the official docs that I could find for registering bouncers.` ? we have improved documentation over the past year.

Might be best if we ask you to revisit these topics to give a fresh insight into your points.

Leopere:
> We have improved `Registering bouncers is far too manual of a process and it blows.` by providing extra ENV variables which allows you to provide multiple bouncers with keys.

yay!



> We have improved `Registering bouncers is far too manual of a process and it blows.` by providing extra ENV variables which allows you to provide multiple bouncers with keys.
> 
> Need some more information on why this is a point `it should if it doesn't currently recognize the difference between docker network sources.` ?
> 
[...]

PR #2959, opened 2024-04-23
enhance: container discovery via labels

### LINKED ISSUE #2957: [docker] auto-discovery

I have been dwelling on this feature for a while and I think it would improve the ease of deploying containers without thinking about updating the acquisition for every deployment.

So this goes on the same methodology that Traefik implements is that we already have a background ticker that is discovering containers and then checking if they match a live acquisition. However, we could inspect the labels tied to the container and if they have something along the lines of this:

```
labels:
  - "crowdsec.enable=true"
  - "crowdsec.labels.program=nginx"
```

Then we could just adopt the container as we know its enabled and that program for the stdout is nginx for example. However, this could be tricky as we only enable a module if its defined within the acquisition. So my first thoughts are:

- Should we detect if there is a local docker socket then enable a ghost acquisition?
- Instead should we inform people to define a acquisition like
```yaml
source: docker
autodiscover: true
```
This means only this acquisition is in charge of auto discovering the containers.

/kind enhancement

### Why is this needed?

Enhances the use of the docker module within crowdsec and users can define containers using labels instead of manually creating an acquisition per container.

LaurenceJJones:
Fix #2957

Implement container auto discovery via labels set on the container

How to setup:

Add acquisition example:

```yaml
##/etc/crowdsec/acquis.d/docker.yaml
source: docker
use_container_labels: true
```

Then you can launch containers using the following labels:

```
docker run -d -p 80:80 -l crowdsec.enable=true -l crowdsec.labels.type=nginx nginx:alpine
```

Under the hood it will check if `crowdsec.enable=true` is set before trying to parse the labels. If no labels are provided it errors as we need to type variables for the parsers.


LaurenceJJones:
Updated config option to be `use_container_labels` since we already do auto discovery this is more explicit of what the option does

LaurenceJJones:
Linked to #1435

pkg/acquisition/modules/docker/docker.go
```diff
@@ -87,10 +87,14 @@ func (d *DockerSource) UnmarshalConfig(yamlConfig []byte) error {
 		d.logger.Tracef("DockerAcquisition configuration: %+v", d.Config)
 	}
 
-	if len(d.Config.ContainerName) == 0 && len(d.Config.ContainerID) == 0 && len(d.Config.ContainerIDRegexp) == 0 && len(d.Config.ContainerNameRegexp) == 0 {
+	if len(d.Config.ContainerName) == 0 && len(d.Config.ContainerID) == 0 && len(d.Config.ContainerIDRegexp) == 0 && len(d.Config.ContainerNameRegexp) == 0 && !d.Config.UseContainerLabels {
 		return fmt.Errorf("no containers names or containers ID configuration provided")
 	}
 
+	if d.Config.UseContainerLabels && len(d.Config.ContainerName) > 0 && len(d.Config.ContainerID) > 0 && len(d.Config.ContainerIDRegexp) > 0 && len(d.Config.ContainerNameRegexp) > 0 {
+		return fmt.Errorf("auto_discover and container_name, container_id, container_id_regexp, container_name_regexp are mutually exclusive")
```
LaurenceJJones, line 95: update wording

pkg/acquisition/modules/docker/docker.go
```diff
@@ -409,6 +446,29 @@ func (d *DockerSource) EvalContainer(container dockerTypes.Container) (*Containe
 
 	}
 
+	if d.Config.UseContainerLabels {
+		parsedLabels := d.getContainerLabels(container.ID)
+		if len(parsedLabels) != 0 {
+			if v, ok := parsedLabels["enable"]; ok && strings.ToLower(v.(string)) == "true" {
```
blotus, line 452: I think this can panic if the user creates a label like `crowdsec.enable.foo` as the type assertion will fail.

pkg/acquisition/modules/docker/docker.go
```diff
@@ -409,6 +446,29 @@ func (d *DockerSource) EvalContainer(container dockerTypes.Container) (*Containe
...
+				if _, ok = parsedLabels["labels"]; !ok {
+					d.logger.Error("container has 'crowdsec.enable' label set to true but no 'labels' keys found")
+					return &ContainerConfig{}, false
+				}
+				d.logger.Debugf("container labels +%v", parsedLabels["labels"])
+				labels := make(map[string]string)
+				for k, v := range parsedLabels["labels"].(map[string]interface{}) {
```
blotus, line 459: Same thing here if the user set `crowdsec.labels=foo` (which is a likely misconfiguration ?)

pkg/acquisition/modules/docker/docker.go
```diff
@@ -375,6 +379,39 @@ func (d *DockerSource) getContainerTTY(containerId string) bool {
...
+	return result
+}
+
+func parseKeyToMap(m map[string]interface{}, key string, value string) {
+	if !strings.HasPrefix(key, "crowdsec") {
+		return
+	}
+	parts := strings.Split(key, ".")
+	for i := 1; i < len(parts)-1; i++ {
+		if i == len(parts)-1 {
```
blotus, line 404: This can never be true as the loop condition is `i < len(parts)-1`

blotus (changes requested):

pkg/acquisition/modules/docker/docker.go
LaurenceJJones: I think its best for the program to panic 🤔 

Jokes aside, I will fix this should be as easy as checking the type cast before each use 🤷🏻

blotus (approved):

PR #2979, opened 2024-04-30
building cross-platform AMD64 and ARM64

vncloudsco:
I came across this tutorial to run on arm and I found it unnecessary it can be handled if we use command dpkg --print-architecture in linux.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ BUILD_REQUIRE_GO_MINOR ?= 21
 
 GO = go
 GOTEST = $(GO) test
-
+GOARCH = $(dpkg --print-architecture)
 BUILD_CODENAME ?= alphaga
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ BUILD_REQUIRE_GO_MINOR ?= 21
 
 GO = go
 GOTEST = $(GO) test
-
+GOARCH = $(dpkg --print-architecture)
```
LaurenceJJones, line 36: but what about non debian systems?

LaurenceJJones:
Consulted the development team we decided to reject the PR due to Makefile being used in all platforms builds meaning that it will cause exception errors on windows, freebsd. If you are compiling directly on the machine, then goarch will already be set correctly so then there no need to manually set it

PR #2986, opened 2024-05-06
cscli: reusable QueryLAPIStatus(), QueryCAPIStatus()

mmetc:
these will be used by "support dump"


mmetc:
done in https://github.com/crowdsecurity/crowdsec/pull/2987

PR #2989, opened 2024-05-07
CI: disable log formatting for consistency

mmetc:
log files won't be formatted/colored during CI tests because they are not in production or during package testing


sabban (approved):
lgtm

PR #3006, opened 2024-05-13
jwt transport: fix retry on unauthorized from CAPI

blotus:
Fix unauthorized from CAPI that might happen in the following scenario:
 - enroll instance
 - restart crowdsec
 - accept instance (on the console)
 - generate alert


buixor (approved):
LGTM

PR #3009, opened 2024-05-14
fix: add sslmode check and append flag to conn string

### LINKED ISSUE #2993: [mariadb/msql] Add support for SSL flag

Currently mariadb and mysql database connection does not support the SSL flag, we should add support

/kind enhancement

### Why is this needed?

Allows users to configure mariadb/mysql to use ssl and CrowdSec can still connect to them

LaurenceJJones:
Fix  #2993

### DIFF
--- a/pkg/csconfig/database.go
+++ b/pkg/csconfig/database.go
@@ -48,7 +48,7 @@ type AuthGCCfg struct {
 }
 
 type FlushDBCfg struct {
-	MaxItems   *int       `yaml:"max_items,omitempty"`
+	MaxItems *int `yaml:"max_items,omitempty"`
 	// We could unmarshal as time.Duration, but alert filters right now are a map of strings
 	MaxAge     *string    `yaml:"max_age,omitempty"`
 	BouncersGC *AuthGCCfg `yaml:"bouncers_autodelete,omitempty"`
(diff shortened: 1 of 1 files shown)


LaurenceJJones:
Tested via configuring mariadb using a self signed certificate via their docs:

```
root@bookworm:/opt/crowdsec# sudo mariadb
MariaDB [(none)]> alter user 'crowdsec'@'%' require ssl;
Query OK, 0 rows affected (0.008 sec)
```

Force the crowdsec user to require ssl

Connecting without ssl causes

```
root@bookworm:/opt/crowdsec# cscli machines list
FATA unable to create new database client: failed creating schema resources: mysql: querying mysql version Error 1045: Access denied for user 'crowdsec'@'localhost' (using password: YES)
```

Using config.yaml to `sslmode: skip-verify` because of self sign certificate allows a connection

```
root@bookworm:/opt/crowdsec# cscli machines list
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Name                                               IP Address   Last Update            Status   Version           Auth Type   Last Heartbeat
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
c90cca144e044e6297dded898a0e76daE3TXLO4bOj87E9E2   127.0.0.1    2024-05-14T11:02:51Z   ✔        v1.6.2-cd84df25   password    49s
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

checking the ssl tabe after connection shows crowdsec is currently connected over TLS

```
MariaDB [(none)]> SHOW SESSION STATUS LIKE 'Ssl_cipher';
+---------------+------------------------+
| Variable_name | Value                  |
+---------------+------------------------+
| Ssl_cipher    | TLS_AES_256_GCM_SHA384 |
+---------------+------------------------+
1 row in set (0.001 sec)
```

pkg/csconfig/database.go
```diff
@@ -131,6 +131,9 @@ func (d *DatabaseCfg) ConnectionString() string {
 		} else {
 			connString = fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?parseTime=True", d.User, d.Password, d.Host, d.Port, d.DbName)
 		}
+		if d.Sslmode != "" {
```
mmetc, line 134: actually, this can be "require" or "disable", right?

LaurenceJJones: for mysql its, true, false, "skip-verify", "required" but its just a string, we detect if its not empty

mmetc (approved):

PR #3019, opened 2024-05-17
db: don't set machine heartbeat until first connection

mmetc:
here I'm not 100% sure of the behavior we want from prune


LaurenceJJones:
Yop 👋🏻 same on the prune, for me I would expect prune to delete anything that I set the time for so even if the last heartbeat is nil it should go from creation time instead which is what you already done I believe. I will test this PR shortly and give comments.

mmetc:
@LaurenceJJones are you ok with this?

PR #3027, opened 2024-05-23
cscli: faster table rendering

mmetc:
this should keep exactly the same look as the previous version, except for a missing hyphen when wrapping path names (which should have never been there)


buixor (approved):
lgtm

PR #3034, opened 2024-05-29
improved tls middleware revocation checks

mmetc:


pkg/apiserver/middlewares/v1/jwt.go
```diff
@@ -59,27 +58,19 @@ func (j *JWT) authTLS(c *gin.Context) (*authInput, error) {
...
 
-		return nil, errors.New("TLS auth is not configured")
+		return nil, err
 	}
 
-	validCert, extractedCN, err := j.TlsAuth.ValidateCert(c)
+	extractedCN, err := j.TlsAuth.ValidateCert(c)
 	if err != nil {
-		log.Error(err)
-		c.JSON(http.StatusForbidden, gin.H{"message": "access forbidden"})
```
buixor, line 71: to check

mmetc: The forbidden error is raised here https://github.com/crowdsecurity/crowdsec/blob/crl/pkg/apiserver/middlewares/v1/api_key.go#L145

and tested here https://github.com/crowdsecurity/crowdsec/blob/crl/test/bats/11_bouncers_tls.bats#L196

As the comment says, since we could not validate the identity I think Unauthorized is more appropriate, but it's splitting hair.

buixor (approved):
lgtm

PR #3074, opened 2024-06-10
CI: update golangci-lint to 1.59; enforce canonical HTTP headers

mmetc:


LaurenceJJones (approved):
lgtm but was there a reason for canonical http headers are just to be RFC compliant?

PR #3076, opened 2024-06-10
cscli: fixed some inconsistency in returning errors

mmetc:
note: cmd.Help() never fails if it's not customized with HelpFunc()

### DIFF
--- a/cmd/crowdsec-cli/alerts.go
+++ b/cmd/crowdsec-cli/alerts.go
@@ -540,7 +540,7 @@ func (cli *cliAlerts) NewInspectCmd() *cobra.Command {
 		DisableAutoGenTag: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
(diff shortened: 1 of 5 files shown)


cmd/crowdsec-cli/alerts.go
```diff
@@ -533,7 +533,7 @@ func (cli *cliAlerts) NewInspectCmd() *cobra.Command {
 		DisableAutoGenTag: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
-				printHelp(cmd)
+				_ = cmd.Help()
```
LaurenceJJones, line 536: is there a reason to underscore the return value when you can just use the function directly?

mmetc: yes, there are linters that complain about unchecked errors. we don't use them for now but it's a nice to have

blotus (approved):

PR #3111, opened 2024-07-05
feat(cscli): add a way to disable port binding in dashboard setup

quantumsheep:
Hey!

I don't necessary want to bind the port.

`-1` seems fine (`0` means random) but if you have another way to disable the port let me know. Another specialized flag maybe?

Thanks,


quantumsheep:
Just saw https://docs.crowdsec.net/blog/cscli_dashboard_deprecation/ 😫

LaurenceJJones:
Closing due to not having enough time to implement before dashboard was removed from `cscli`.

thank you for putting in the time to contribute.

PR #3112, opened 2024-07-05
feat(cscli): add network flag for dashboard setup

quantumsheep:
Hey!

Another one for the nginx-proxy use-case. I want to connect the created container to a specific network.

Thanks,


LaurenceJJones:
Closing due to not having enough time to implement before dashboard was removed from `cscli`.

thank you for putting in the time to contribute.

PR #3115, opened 2024-07-08
Store alert remediations status in DB

### LINKED ISSUE #2869: [Alert db model] Remediation type on struct is not held in DB

So since I refactored `cscli notifications reinject` it seems it not longer "works" as intended, because the remediation status is not held within the alert at the database level. So when an alert comes into the API it will trigger the notification, however, if a user does the reinject command it wont.

We should update the alert schema to hold the remediation at the database level.

/kind enhancement

### Why is this needed?

the reason this is handy if the user checks the previous alert then at least they can get this information, as the user might not know if there was a remediation or not

Also this will patch `cscli notficiations reinject` to work as intended without users having to do

`cscli notifications reinject <id> -a '{ "remediation": true}'`

blotus:
Fixes #2869


LaurenceJJones:
Just note if we can also put this in the output of `cscli alerts inspect` in the template

buixor (approved):

PR #3129, opened 2024-07-12
remove warning "maxopenconns is 0, default to 100"

mmetc:
also don't store as pointer since value 0 is not useful


sabban (approved):
lgtm

PR #3144, opened 2024-07-19
update expr

blotus:


sabban (commented):
minor comments

pkg/exprhelpers/debugger.go
```diff
@@ -106,62 +107,30 @@ func (o *OpOutput) String() string {
...
-		if i == endLine {
-			code_snippet += lines[i][:endCol]
-			break
 		}
-		code_snippet += lines[i]
 	}
 
-	log.Tracef("#code extract for ip %d [%s] -> '%s'", ip, parts[1], code_snippet)
-	return cleanTextForDebug(code_snippet)
+	var end int
```
LaurenceJJones, line 126: Can we set end to len(src) and `closest != nil` then set closest.From?

LaurenceJJones: remove if else to a single if

pkg/exprhelpers/debugger.go
```diff
@@ -106,62 +107,30 @@ func (o *OpOutput) String() string {
...
-	for i := startLine; i <= endLine; i++ {
-		if i == startLine {
-			if startLine != endLine {
-				code_snippet += lines[i][startColumn:]
-				continue
+	var closest *file.Location
+
+	for i := ip + 1; i < len(locations); i++ {
+		if locations[i].From > currentInstruction.From {
+			if closest == nil || locations[i].From < closest.From {
```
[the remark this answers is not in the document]
blotus: `closest` will only be nil until we find a first value, afterwards we want to replace what we have with any value that is closer (the location slice is not sorted)

pkg/exprhelpers/debugger.go
blotus: The case where `closest` is nil is the less common one (it will only happen for the last instruction of the expression), so we would be running a "useless" len each time.

pkg/exprhelpers/debugger.go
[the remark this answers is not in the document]
sabban: ok, I missed it is in a loop

sabban (approved):
lgtm

PR #3159, opened 2024-07-31
fix #1742 : Allow QueryCAPIStatus to return as well enrollment status

### LINKED ISSUE #1742: Allow `cscli` to tell if an instance is enrolled within the console

I would like a `cscli` command, such as `cscli capi status` or `cscli console status` to let me know if :
 - my instance is enrolled in the console
 - my enrollment request has been accepted/rejected

### Why is this needed?

Currently, I need to check into the console to know my instance status

buixor:
`cscli capi status` now as well tells if we're enrolled

### DIFF
--- a/cmd/crowdsec-cli/capi.go
+++ b/cmd/crowdsec-cli/capi.go
@@ -210,10 +210,6 @@ func QueryCAPIStatus(hub *cwhub.Hub, credURL string, login string, password stri
 		return false, false, err
 	}
 
-	if err = client.GetClient().Transport.(*apiclient.JWTTransport).Expiration.UnmarshalText([]byte(authResp.Expire)); err != nil {
-		return true, false, err
-	}
-
 	client.GetClient().Transport.(*apiclient.JWTTransport).Token = authResp.Token
 
 	if client.IsEnrolled() {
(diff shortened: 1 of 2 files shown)


cmd/crowdsec-cli/capi.go
```diff
@@ -180,12 +205,22 @@ func QueryCAPIStatus(hub *cwhub.Hub, credURL string, login string, password stri
...
-	return nil
+	if err = client.GetClient().Transport.(*apiclient.JWTTransport).Expiration.UnmarshalText([]byte(authResp.Expire)); err != nil {
+		return true, false, err
+	}
+
+	client.GetClient().Transport.(*apiclient.JWTTransport).Token = authResp.Token
+
+	if client.IsEnrolled() {
+		return true, true, nil
+	} else {
```
mmetc, line 221: you can drop the else branch around the return, it will eventually be enforced by a linter, if it's not already, because it has led to unnecessary complexity

mmetc (approved):

mmetc:
if you configure golangci-lint correctly for your IDE you should see the warnings/errors while editing instead of finding out in the CI.

PR #3191, opened 2024-08-26
implement GetFSType on openbsd with the correct statfs struct member

rnagy:


mmetc (approved):

mmetc:
Thanks! Backported for 1.6.3 in https://github.com/crowdsecurity/crowdsec/pull/3196

PR #3217, opened 2024-09-06
refact acquisition: build profiles (optionally exclude datasources from final binary)

mmetc:


LaurenceJJones:
Just to ask a question instead of determining `"minimal"` shouldnt we create it so users have to define the build flags by default? yes it will break third partys but they can patch it no? as I see users maybe wanting to create a "docker" acquisition only build or something which they cannot do from this.

mmetc:
> Just to ask a question instead of determining `"minimal"` shouldnt we create it so users have to define the build flags by default? yes it will break third partys but they can patch it no? as I see users maybe wanting to create a "docker" acquisition only build or something which they cannot do from this.
> 
> Yes, its an edge case which they can patch there own Makefile but 🤷🏻

I'm rewriting that part.

mmetc:
@LaurenceJJones better now?

sabban (approved):
lgtm

mmetc:
If we merge this, I'll remove "EXCLUDE_MEDIUM", it was just an example.

blotus (approved):

PR #3242, opened 2024-09-17
remove dependency from pkg/cwversion to pkg/acquisition

mmetc:
This allows to build cscli with no datasource code if csli_setup is excluded from the build.


sabban (approved):
lgtm

Issue #3188, opened 2024-08-23
[CAPI] Configuration to enable / disable sending signals


LaurenceJJones:
Since low touch plans are on the rise currently CrowdSec does not allows users to disable the sharing part without revoking all access to CAPI. We should add configuration to disable sharing signals but this does not matter for receiving the community blocklists as if they are enterprise they will still receive the intended list.

PR #3277, opened 2024-10-03
Add explicit configuration for signals sharing and blocklists pull

blotus:
Fixes https://github.com/crowdsecurity/crowdsec/issues/3188.

This PR adds 3 new config options:
 - `server.online_client.sharing`: Should crowdsec send alerts to CAPI? Defaults to true.
 - `server.online_client.pull.community`: When querying the decision stream in CAPI, should we get the content of the community blocklist ? Defaults to true.
 - `server.online_client.pull.blocklists`: When querying the decision stream in CAPI. should we get the content of additional blocklists the instance is subscribed to ? Defaults to true.
 
 This goes a bit beyond the scope of the initial issue, but the added granularity is probably worth it (for example, only push signals with no pull when running a honeypot, only pull additional blocklist if you use the console to distribute custom lists,...)


pkg/apiserver/apic.go
```diff
@@ -650,23 +663,22 @@ func (a *apic) PullTop(ctx context.Context, forcePull bool) error {
...
 
-	err = a.SaveAlerts(ctx, alertsFromCapi, addCounters, deleteCounters)
-	if err != nil {
-		return fmt.Errorf("while saving alerts: %w", err)
+		err = a.SaveAlerts(ctx, alertsFromCapi, addCounters, deleteCounters)
+		if err != nil {
+			return fmt.Errorf("while saving alerts: %w", err)
+		}
+	} else {
+		log.Infof("capi/community-blocklist : received 0 new entries (expected if you just installed crowdsec)")
```
LaurenceJJones, line 681: change log depending if they explicit didnt want the blocklist?

LaurenceJJones (approved):

Issue #3289, opened 2024-10-18
Generic HTTP datasource 


blotus:
Add a new datasource that will expose a generic HTTP server to receive JSON formatted logs.

The datasource will be agnostic regarding the exact content of the logs, and if "complex" objects are received (for example an array of items), it will rely on the [transform](https://docs.crowdsec.net/docs/next/data_sources/intro/#transform) features to split the items.

Requirements:
 -  Authentication: Support basic auth and custom headers name/values (in the case of header, it should probably allow an `AND` between multiples headers as some services seem to send multiple auth-related headers). In P2, mTLS support.
 - TLS Support

Optional:
 - Source IP filtering

Questions:
 - Input format: At the beginning, only support JSON data. Should we only support "one-shot" requests (ie, every items in an array) or NDJSON (one item per line).
 - Maximum body size: should the user be able to configure a max body size to avoid potential memory issues when dealing with a lot of logs ?
 - Chunk size configuration ?(mostly for streaming purpose)
 - Compression support ? Rely on the Content-type provided by the client ?
 - Stream support for decoding ? (ie, client open a persistent connection and stream data to crowdsec).
 - Configuration of HTTP response code on success and success headers.
 - Timeout configuration (should we kill the connection if no data is transmitted after some time)
 - How to support client that perform a health-check ? Have a configuration with a custom route + return code ?

LaurenceJJones:
Mark as solved due to #3294 ??

PR #3294, opened 2024-10-23
add HTTP datasource

he2ss:
feature : https://github.com/crowdsecurity/crowdsec/issues/3289


pkg/acquisition/modules/http/http.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+		tlsConfig.Certificates = []tls.Certificate{cert}
+	}
+
+	if hc.AuthType == "mtls" && hc.TLS.CaCert != "" {
+		caCert, err := os.ReadFile(hc.TLS.CaCert)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read ca cert: %w", err)
+		}
+		caCertPool := x509.NewCertPool()
```
LaurenceJJones, line 222: Some users may want to load from system pool?

ref: https://github.com/crowdsecurity/go-cs-bouncer/blob/9e16e182103d8f9ac30352bff0c1eca27b0619ba/tls.go#L11-L35

pkg/acquisition/modules/http/http.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	if hc.Path == "" {
+		hc.Path = "/"
+	}
+	if hc.Path[0] != '/' {
+		return errors.New("path must start with /")
+	}
+
+	switch hc.AuthType {
```
LaurenceJJones, line 105: The errors returned here do not provide context as to why `X is required`

I would expect it to say "Basic auth is selected but no X is provided"

This is small nit pick but could provide enough error context to user to self resolve

mmetc (commented):
All data sources should be optionally excluded with build tags: grep for 'datasource_kafka' for an example. They also register with an init() method so you can see with "cscli version" which ones were included

blotus (approved):

Issue #3307, opened 2024-10-29
New datasource: VictoriaLogs


zekker6:
### What would you like to be added?

I would like to have a support of VictoriaLogs (API docs available [here](https://docs.victoriametrics.com/victorialogs/querying/)) in a way where crowdsec would pull/watch for new log entries similar to an existing integration with [Loki](https://docs.crowdsec.net/docs/next/data_sources/loki/).

P.S. I'm one of maintainers of VictoriaLogs and would be happy to create a PR to add the support of this datasource and maintain it later of if needed. Just wanted to check in if you will find this useful before sending a PR.

/kind feature

### Why is this needed?

I'm currently pushing all logs coming from different sources to VictoriaLogs and the only way to use crowdsec is to set up a cron task to pull the logs and append these to log lines to a file watched by crowdsec.
This works, but having a datasource support would be make it much cleaner.

buixor:
Hello,

It sounds like a great idea! Please go ahead, and we'll do our best to review at the appropriate time. Don't hesitate to reach out to us on Discord if needed.

There is no public documentation on how to implement a datasource, however you can look at this interface : https://github.com/crowdsecurity/crowdsec/blob/master/pkg/acquisition/acquisition.go

The Loki datasource should be a good source of inspiration for you too :
https://github.com/crowdsecurity/crowdsec/tree/master/pkg/acquisition/modules/loki

(note we didn't use the tail endpoint of Loki as it was missing logs when there was a burst, but you might not have this issue w/ Victoria).

tiny-pangolin:
Currently, my solution to this is using telegraf to tail the logs on each sensor, but I would love to see a native API integration, so I don't have to store logs on disk

zekker6:
Cloding this as it was added in [v1.6.5](https://github.com/crowdsecurity/crowdsec/releases/tag/v1.6.5)

PR #3310, opened 2024-10-30
acquisition/victorialogs: add new datasource

zekker6:
Related issue: https://github.com/crowdsecurity/crowdsec/issues/3307

Data source supports:
- cat mode with automatic adjustment of poll interval (same as one at Loki datasource)
- tail mode by using tailing API

Most of the logic is exactly the same as Loki datasource.
The only difference is related to adding support of live dailing and different response formats.
VictoriaLogs uses newline delimeted JSON for both reqular query and tail modes.

Tested locally with the following config:
```
source: victorialogs
log_level: debug
url: https://vlogs-endpoint
mode: tail
limit: 1000
query: |
  app_name:nginx
labels:
  type: nginx
---
source: victorialogs
log_level: debug
url: https://vlogs-endpoint
mode: cat
since: 1m
limit: 1000
query: |
  app_name:nginx
labels:
  type: nginx
```


pkg/cwversion/component/component.go
```diff
@@ -7,20 +7,21 @@ package component
...
+	"datasource_file":         false,
+	"datasource_journalctl":   false,
+	"datasource_k8s-audit":    false,
+	"datasource_kafka":        false,
+	"datasource_kinesis":      false,
+	"datasource_loki":         false,
+	"datasource_s3":           false,
+	"datasource_syslog":       false,
+	"datasource_wineventlog":  false,
+	"datasource_victorialogs": false,
```
mmetc, line 23: this should be listed in the Makefile too -- the only reason to "register" datasources is to optionally exclude from the builds with a tag, and still list them in "cscsli version"

zekker6: Added this to Makefile in https://github.com/crowdsecurity/crowdsec/pull/3310/commits/f4603e1b45bbad65d07cb03b288637f5093db92c, thanks for noticing

denisgolius (commented):
AFAIK it requires to install curl every time when mostly Linux Distributions have installed `wget` out-from-box. Do you think curl will be better?

blotus:
> AFAIK it requires to install curl every time when mostly Linux Distributions have installed `wget` out-from-box. Do you think curl will be better decision?

This change only impacts the CI, and AFAIK curl is installed by default on the images provided by github.

I made the change just to test/debug why the victoria logs container seems to fail to start (or at least, github/docker thinks it has failed, I cannot reproduce the issue locally).

zekker6:
Hello @blotus, I've updated healthcheck so it should work now. This can be tested locally with:
```
docker run --rm -it --name=victorialogs1 \
                            --health-cmd "wget -q -O - http://0.0.0.0:9428" \
                            --health-interval 30s \
                            --health-timeout 10s \
                            --health-retries 5 \
                            --health-start-period 30s \
                  victoriametrics/victoria-logs:v1.5.0-victorialogs
```
Previously, it was using `localhost` which does not work in alpine based images.

zekker6:
@blotus Hopefully I've fixed the last reason for the tests to fail at https://github.com/crowdsecurity/crowdsec/pull/3310/commits/2b42a4374dd6347ea5c6c392f588e3debf3d8c26, could you please re-run tests?

mmetc:
@blotus should this method be called or removed

  Error: pkg/acquisition/modules/victorialogs/internal/vlclient/vl_client.go:69:21: func `(*VLClient).resetFailStart` is unused (unused)
  func (lc *VLClient) resetFailStart() {

zekker6:
@mmetc I've removed unused method as it was not needed.

mmetc (approved):

blotus (approved):

Issue #2999, opened 2024-05-13
appsec: allow to load multiple appsec-config per appsec datasource


blotus:
Currently, only one appsec config is allowed per appsec datasource.
This is quite limiting, as if a user wants to have multiple set of rules (for example, vpatch rules + CRS), they will need to create a custom appsec config, which might or might not get new appsec rules depending on how it is written.

We should allow to load multiple appsec config, each config would be stand-alone (ie, hooks from one config would not affect the other one)

LaurenceJJones:
keep for `1.6.3` or _not_ ?

buixor:
Currently, several approaches are possible:

# Solution 1 : We merge/concat everything

 - We append `out of band/Inband` Rules and `on_load/pre_eval/post_eval/on_match` Hooks.

 - For all possibly conflicting options ( `default_pass_action`, `default_remediation` , `blocked_http_code`, `passed_http_code`, `user_blocked_http_code` and `user_passed_http_code`) and (`disable_body_inspection` / `request_body_in_memory_limit`)  the last one take the precedence.
 - For Hooks, they are all evaluated in-line, which means that in case of conflicting instructions (ie. `SetRemediation`), the last one will take the precedence.





~~# Solution 2 : We deal with them sequentially~~
~~We keep each `AppsecRuntimeConfig` on its own, and we process each of them sequentially~~ 
 
Loking at the code, `solution #2` will imply an explosion of memory costs and doesn't look like a viable solution.

buixor:
fixed in https://github.com/crowdsecurity/crowdsec/pull/3314

PR #3314, opened 2024-11-04
Support multiple appsec configs

buixor:
cf. #2999

 - Allow AppSec to load multip appsec config at once (rules & hooks are appended, for mutually exclusive options, last value is taken into account)


pkg/acquisition/modules/appsec/appsec.go
```diff
@@ -120,7 +121,7 @@ func (w *AppsecSource) UnmarshalConfig(yamlConfig []byte) error {
 		w.config.Routines = 1
 	}
 
-	if w.config.AppsecConfig == "" && w.config.AppsecConfigPath == "" {
+	if w.config.AppsecConfig == "" && w.config.AppsecConfigPath == "" && len(w.config.AppsecConfigs) == 0 {
```
blotus, line 124: We also need a check to make those 3 options mutually exclusive

blotus (approved):

PR #3317, opened 2024-11-07
loki: add no_ready_check option

blotus:
When using Loki hosted in Grafana cloud, the `/ready` endpoint does not exist, which prevents crowdsec from starting.

Add an option to bypass this check.
Defaults to `false` to keep the same behavior by default.

### DIFF
--- a/pkg/acquisition/modules/loki/loki.go
+++ b/pkg/acquisition/modules/loki/loki.go
@@ -283,19 +283,18 @@ func (l *LokiSource) OneShotAcquisition(ctx context.Context, out chan types.Even
 		}
 	}
 
-	ctx, cancel := context.WithCancel(ctx)
-	c := l.Client.QueryRange(ctx, false)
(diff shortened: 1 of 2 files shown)


pkg/acquisition/modules/loki/loki.go
```diff
@@ -313,12 +324,17 @@ func (l *LokiSource) readOneEntry(entry lokiclient.Entry, labels map[string]stri
 }
 
 func (l *LokiSource) StreamingAcquisition(ctx context.Context, out chan types.Event, t *tomb.Tomb) error {
+	var err error
```
LaurenceJJones, line 327: unused var?

LaurenceJJones (approved):

zoispag:
This is not documented in [Loki]([](https://docs.crowdsec.net/docs/next/log_processor/data_sources/loki/)) Data source, is it?

PR #3347, opened 2024-11-27
appsec: do not attempt to deduplicate native modsec rules

### LINKED ISSUE #3343: Appsec rule deduplication breaks multilines seclang rules
When deduplication is performed on the loaded appsec rules, we do not handle properly multilines seclang rules, for example (from the CRS):
```
SecAction \
    "id:901200,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
...
```

The parser will consider `SecAction \` to be its own rule and will remove all duplicated instances, which breaks CRS support.

buixor:
Do not attempt to dedup modsec native rules as they might have no IDs. Doing this can mess up stuff (see #3343 and #3350 )


pkg/acquisition/modules/appsec/appsec_runner.go
```diff
@@ -45,6 +46,8 @@ func (r *AppsecRunner) MergeDedupRules(collections []appsec.AppsecCollection, lo
 				logger.Debugf("Discarding duplicate rule : %s", rule)
 			}
 		}
+		// Don't mess up with native modsec rules
+		rulesArr = append(rulesArr, collection.NativeRules...)
```
sabban, line 50: If I understand correctly, you should make the append after the logger.Warningf() to avoid telling misleading information to users (ie count the NativeRules in excess for the duplicates)

sabban (approved):
lgtm, except for the logger.Warning() that could be a bit misleading.

pkg/acquisition/modules/appsec/appsec_runner.go
```diff
@@ -35,19 +35,24 @@ type AppsecRunner struct {
...
+				discarded++
 				logger.Debugf("Discarding duplicate rule : %s", rule)
 			}
 		}
+		// Don't mess up with native modsec rules
+		rulesArr = append(rulesArr, collection.NativeRules...)
 	}
-	if len(rulesArr) != len(dedupRules) {
-		logger.Warningf("%d rules were discarded as they were duplicates", len(rulesArr)-len(dedupRules))
+	if discarded > 0 {
```
LaurenceJJones, line 54: not an issue but personally I would have done

```go
if discarded := len(rulesArr) - len(dedupRules); discarded > 0 {
  logger.Warningf("%d rules were discarded as they were duplicates", discarded)
}
```

pkg/acquisition/modules/appsec/appsec_runner.go
```diff
@@ -35,19 +35,24 @@ type AppsecRunner struct {
 func (r *AppsecRunner) MergeDedupRules(collections []appsec.AppsecCollection, logger *log.Entry) string {
 	var rulesArr []string
 	dedupRules := make(map[string]struct{})
+	discarded := 0
 
 	for _, collection := range collections {
+		// Dedup *our* rules
 		for _, rule := range collection.Rules {
 			if _, ok := dedupRules[rule]; !ok {
```
LaurenceJJones, line 43: Can we improve the readability of this range

```go
if _, ok := dedupRules[rule]; ok {
  // discarded++
  logger.Debugf("Discarding duplicate rule : %s", rule)
  continue
}
rulesArr = append(rulesArr, rule)
dedupRules[rule] = struct{}{}
```

pkg/acquisition/modules/appsec/appsec_runner.go
LaurenceJJones: actually I see what manu was saying beforehand so yeah we need to track an extra variable, classing as resolved.

LaurenceJJones (approved):

PR #3355, opened 2024-12-08
Add support for centralized allowlists

blotus:
This PR adds a new type of allowlist that is managed by LAPI and applies to alerts, blocklists content and appsec (but not to manual decisions with `cscli`):
 - alerts are dropped (with a log message)
 - blocklist content matching an allowlist is ignored
 - appsec requests matching an allowlist are not processed

Compared to existing types of allowlists in crowdsec (in parsers, postoverflows, custom profiles or appsec hooks), they only support IPs and ranges (ie, no arbitrary expression), but they can have an optional expiration.

An alert is considered allowlisted in the following situations:
 - The source of the alert is an exact match with a non-expired allowlist item
 - The source of the alert belongs to a non-expired (range) allowlist item
 -  The source of the alert contains a non-expired allowlisted item (eg, `1.2.3.4` is allowlisted, and an alert on `1.2.3.0/24` is generated): while this seems counter-intuitive, range alerts are rare and it would be add a lot of complexity to carve out the specific IPs that are allowlisted from the alert source (and generate multiple sub-alerts).

They can be managed with:
 - `cscli`
 - The console (allowlists created from the console are not editable with `cscli` to avoid conflict), in which case they are pulled from CAPI or PAPI.
 
If a local allowlist is created, and another one with the same name is created in the console, the one from the console will replace the local one.


cmd/crowdsec-cli/cliallowlists/allowlists.go
```diff
@@ -0,0 +1,562 @@
...
+		enc := json.NewEncoder(out)
+		enc.SetIndent("", "  ")
+
+		if err := enc.Encode(allowlists); err != nil {
+			return errors.New("failed to serialize")
+		}
+
+		return nil
+	case "raw":
+		//return cli.listCSV(out, allowlists)
```
buixor, line 257: missing output ?

cmd/crowdsec-cli/cliallowlists/allowlists.go
```diff
@@ -0,0 +1,562 @@
...
+	for _, allowlist := range allowlists {
+		if strings.Contains(allowlist.Name, toComplete) && !slices.Contains(args, allowlist.Name) && !allowlist.FromConsole {
+			ret = append(ret, allowlist.Name)
+		}
+	}
+
+	return ret, cobra.ShellCompDirectiveNoFileComp
+}
+
+func (cli *cliAllowLists) validAllowlistsWithConsole(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
buixor, line 78: comment on this function purpose and the why (ie. allowlists from console and cscli are strictly separated).

cmd/crowdsec-cli/cliallowlists/allowlists.go
```diff
@@ -0,0 +1,562 @@
...
+	flags := cmd.Flags()
+
+	flags.StringP("description", "d", "", "description of the allowlist")
+
+	cmd.MarkFlagRequired("description")
+
+	return cmd
+}
+
+func (cli *cliAllowLists) create(cmd *cobra.Command, args []string) error {
```
buixor, line 187: I think we want console-based and cscli-based allowlists to have different prefixes ?

cmd/crowdsec-cli/cliallowlists/allowlists.go
```diff
@@ -0,0 +1,562 @@
...
+	t.AppendHeader(table.Row{"Name", "Description", "Creation Date", "Updated at", "Managed by Console", "Size"})
+
+	for _, allowlist := range *allowlists {
+		t.AppendRow(table.Row{allowlist.Name, allowlist.Description, allowlist.CreatedAt, allowlist.UpdatedAt, allowlist.ConsoleManaged, len(allowlist.Items)})
+	}
+
+	io.WriteString(out, t.Render()+"\n")
+}
+
+func (cli *cliAllowLists) listContentHuman(out io.Writer, allowlist *models.GetAllowlistResponse) {
```
buixor, line 125: I think we want the origin (cscli | console) to appear here

cmd/crowdsec-cli/cliallowlists/allowlists.go
```diff
@@ -0,0 +1,562 @@
...
+		enc := json.NewEncoder(out)
+		enc.SetIndent("", "  ")
+
+		if err := enc.Encode(allowlist); err != nil {
+			return errors.New("failed to serialize")
+		}
+
+		return nil
+	case "raw":
+		//return cli.listCSV(out, allowlists)
```
buixor, line 475: missing output

cmd/crowdsec-cli/main.go
```diff
@@ -166,6 +168,8 @@ func (cli *cliRoot) initialize() error {
 		}
 	}
 
+	csConfig.DbConfig.LogLevel = ptr.Of(cli.wantedLogLevel())
```
buixor, line 171: unrelated change ?

pkg/apiserver/papi_cmd.go
```diff
@@ -224,27 +219,79 @@ func ManagementCmd(message *Message, p *Papi, sync bool) error {
 
 		ctx := context.TODO()
 
-		if forcePullMsg.Blocklist == nil {
+		if forcePullMsg.Blocklist == nil && forcePullMsg.Allowlist == nil {
 			p.Logger.Infof("Received force_pull command from PAPI, pulling community and 3rd-party blocklists")
```
buixor, line 223: update message to mention allowlist

pkg/acquisition/modules/appsec/appsec.go
```diff
@@ -31,6 +32,8 @@ const (
 )
 
 var DefaultAuthCacheDuration = (1 * time.Minute)
+var negativeAllowlistCacheDuration = (5 * time.Minute)
+var positiveAllowlistCacheDuration = (5 * time.Minute)
```
buixor, line 36: looks like a lot to me, but it depends if SE gets a `force_pull` via PAPI when the allowlist is updated ?

pkg/apiclient/auth_jwt.go
```diff
@@ -186,17 +186,18 @@ func (t *JWTTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
-			log.Tracef("resp-jwt: %s (err:%v)", string(dump), err)
-		}
 
 		if err != nil {
 			// we had an error (network error for example), reset the token?
 			t.ResetToken()
 			return resp, fmt.Errorf("performing jwt auth: %w", err)
 		}
 
+		if log.GetLevel() >= log.TraceLevel {
```
buixor, line 196: unrelated change ?

pkg/apiclient/auth_jwt.go
```diff
@@ -186,10 +186,6 @@ func (t *JWTTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 		}
 
 		resp, err = t.transport().RoundTrip(clonedReq)
-		if log.GetLevel() >= log.TraceLevel {
```
buixor, line 189: unrelated change ?

pkg/acquisition/modules/loki/internal/lokiclient/loki_client.go
```diff
@@ -205,6 +205,7 @@ func (lc *LokiClient) getURLFor(endpoint string, params map[string]string) strin
 func (lc *LokiClient) Ready(ctx context.Context) error {
 	tick := time.NewTicker(500 * time.Millisecond)
 	url := lc.getURLFor("ready", nil)
+	lc.Logger.Debugf("Using url: %s for ready check", url)
```
buixor, line 208: unrelated change?

pkg/apiserver/controllers/v1/alerts.go
```diff
@@ -154,6 +175,11 @@ func (c *Controller) CreateAlert(gctx *gin.Context) {
 			}
 		}
 
+		if allowlisted, reason := c.isAllowListed(ctx, alert); allowlisted {
+			log.Infof("alert source %s is allowlisted by %s, skipping", *alert.Source.Value, reason)
```
buixor, line 179: Let's add scenario, and if we can the allowlist name

PR #3357, opened 2024-12-10
cscli: display expired decisions' expiration time in red

mmetc:
had a customer testing the bouncer without noticing the decisions were expired.


sabban (approved):
lgtm

Issue #3183, opened 2024-08-20
[cscli] Test environment improvements


LaurenceJJones:
### What would you like to be added?

Currently if a user wants to download a test environment its pretty hard to get it configured correctly here are the issue:

- [ ] Cscli binary `hubtest` and `explain` evoke CrowdSec from path and this may not be the case for test environments
- [ ] Previous #3182 means patterns dir is not respected
- [ ] Invoking ./test_env.sh should replace configurations with full paths instead of relative directories

/kind enhancement

### Why is this needed?

Easier to setup a dev / test environment

Natsui31:
<details><summary>Example of bug repro from fresh debian 12</summary><p>

```
apt update \
  && apt install -y gettext curl wget git libsqlite3-0
cd $(mktemp -d)
wget https://github.com/crowdsecurity/crowdsec/releases/download/v1.6.2/crowdsec-release.tgz
tar xvzf crowdsec-release.tgz
cd crowdsec-v1.6.2
./test_env.sh
export PATH=$PATH:"$PWD"
alias csdev="$PWD/cmd/crowdsec-cli/cscli -c $PWD/config/dev.yaml"
git clone https://github.com/crowdsecurity/hub
export PATH=$PATH:"$PWD/cmd/crowdsec/":"$PWD/cmd/crowdsec-cli/"
cd hub
csdev hubtest run --all
```
</p></details>

With this environment, we got the same error as described in [[cscli] We added patterns_dir option to config.yaml but isnt used in hubtest](https://github.com/crowdsecurity/crowdsec/issues/3182#issue-2475619361)

`config_paths.pattern_dir` is not set on `./config/dev.yaml`

```
root@05391e978e44:/tmp/tmp.MqISOY3GSv/crowdsec-v1.6.2/hub# csdev hubtest run --all
+ /tmp/tmp.MqISOY3GSv/crowdsec-v1.6.2/cmd/crowdsec-cli/cscli -c /tmp/tmp.MqISOY3GSv/crowdsec-v1.6.2/config/dev.yaml hubtest run --all
INFO Running test 'CVE-2017-9841'
ERRO running test 'CVE-2017-9841' failed: unable to copy 'patterns' from '/etc/crowdsec/patterns' to '/tmp/tmp.MqISOY3GSv/crowdsec-v1.6.2/hub/.tests/CVE-2017-9841/runtime/patterns': open /etc/crowdsec/patterns: no such file or directory
[...]

PR #3386, opened 2024-12-29
enhancement: Hubtest respect patterndir option set via config.yaml

LaurenceJJones:
Another one for #3183 , this passes through the configuration option for patterndir set via config.yaml

However, I have to unset the "noNeedConfig" option for hubtest, but I see if @mmetc has a better way 🤷🏻


buixor:
Might want to fallback to `csconfig.DefaultConfigPath("patterns")` if it's not set. @mmetc ?

LaurenceJJones:
> Might want to fallback to `csconfig.DefaultConfigPath("patterns")` if it's not set. @mmetc ?

`cscli` on startup calls the configuration init command and it automatically sets `pattern_dir` to `config_dir + /patterns`

mmetc (approved):

PR #3416, opened 2025-01-20
use the actual bucket name when checking for simulation mode

blotus:


LaurenceJJones:
@blotus did the CI just die or is it an actual issue with code change?

mmetc (approved):

Issue #3424, opened 2025-01-23
Flush.agents_autodelete deletes activate agents


srkoster:
### What happened?

I'm running Crowdsec with the crowdsec helm-chart on Kubernetes and am using Postgresql as my database. My agents use an auto registration token to register themselves with the LAPI, so after a while, the list of agents grows and I would like to clean the unused agent entries. 

I can do it manually with the cli `cscli machines prune` which works fine. Things got strange when I activated the `agents_autodelete` in my `db_config`:
```
db_config:
  type:     postgresql
  ...
  flush:
    agents_autodelete:
      login_password: 30m
```

After about 1 hour all my agent pods where in a crashloop with the message "unable to start crowdsec routines: authenticate watcher (name of pod): API error: ent: machine not found"

I found out that all my machines where deleted from the LAPI (both through `cscli machines list` and looking in the database).

### What did you expect to happen?

Only agents that had a latest heartbeat <= now - 30 minutes should be deleted.
Active agents should be kept in the machines list of the LAPI 

### How can we reproduce it (as minimally and precisely as possible)?

On version 1.6.4, set the `db_config` as indicated and wait for 1 hour.

### Anything else we need to know?

When analyzing my Postgresql database logs, I found the problem. Every now and then, a query similar to the one below was being issued by crowdsec. It updates the `machines` table for an agent and also updated the `last_heartbeat` column. However, it sets the `last_heartbeat` column with the first timestamp that the agent was seen (retrieved from the metrics). If the flush operation is run after this query, and before the next heartbeat update from the agent, the agent gets deleted falsely.

![Image](https://github.com/user-attachments/assets/8ecd4f34-68a0-4db1-9619-162750014e9e)

The origin of the logic for the `last_heartbeat` update resides in the [MachineUpdateBaseMetrics](https://github.com/crowdsecurity/crowdsec/blob/4935dc536e9be4da2cb49f05ab7ca68d57c8ffae/pkg/database/machines.go#L39C2-L43C3) which I believe shouldn't update the `last_heartbeat` column at all.

### Crowdsec version

<details>

```console
$ cscli version
version: v1.6.4-523164f6
Codename: alphaga
BuildDate: 2024-11-26_13:16:01
GoVersion: 1.23.3
Platform: docker
libre2: C++
User-Agent: crowdsec/v1.6.4-523164f6-docker
Constraint_parser: >= 1.0, <= 3.0
Constraint_scenario: >= 1.0, <= 3.0
[...]

PR #3425, opened 2025-01-23
Removed updating of machine last_heartbeat based on baseMetrics in MachineUpdateBaseMetrics

srkoster:
This fix removes the update of machine last_heartbeat in `MachineUpdateBaseMetrics` for 3 reasons
- The implemented logic doesn't make sense. It updates the fields with the first timestamp of found for that machine in the baseMetrics. When using flush.agents_autodelete, agents can be deleted wrongly because of this logic. (See #3424 
- It shouldn't update the last_heartbeat column at all. Machines will call back to the LAPI and `UpdateMachineLastHeartBeat` is being called to update it.
- `BouncerUpdateBaseMetrics` in the bouncer package doesn't update the similar column last pull neither.

### DIFF
--- a/pkg/database/machines.go
+++ b/pkg/database/machines.go
@@ -34,14 +34,6 @@ func (c *Client) MachineUpdateBaseMetrics(ctx context.Context, machineID string,
 	os := baseMetrics.Os
 	features := strings.Join(baseMetrics.FeatureFlags, ",")
 
-	var heartbeat time.Time
-
-	if len(baseMetrics.Metrics) == 0 {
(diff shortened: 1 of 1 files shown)


buixor:
The issue is that the [payload generated by the LP is static right now](https://github.com/crowdsecurity/crowdsec/blob/1b5514c9e47e8398956d57ca21a5a472afe706ac/cmd/crowdsec/lpmetrics.go#L140) so the timestamp is the startup timestamp. I agree this fixes the issue while we think further about what we want to do with LP metrics.

buixor (approved):

PR #3434, opened 2025-01-29
don't ask user to reload systemd service when running in docker

mmetc:


mmetc:
With the last change, restart message and user confirmation prompts are only displayed

 - if run in a shell (docker exec... then cscli)
 - if the container is started with run -ti

blotus (approved):

LaurenceJJones (approved):

PR #3451, opened 2025-02-07
cscli hub/items: always show action plan; fix --interactive in pipes

mmetc:


mmetc:
TODO: make hub update and upgrade honor the --error flag, even if it's not logging.

blotus (approved):

PR #3457, opened 2025-02-09
Flags for console features

Simbiat:
Add support for console features' flags besides management

### DIFF
--- a/docker/docker_start.sh
+++ b/docker/docker_start.sh
@@ -447,9 +447,31 @@ for BOUNCER in $(compgen -A variable | grep -i BOUNCER_KEY); do
     fi
 done
 
-if [ "$ENABLE_CONSOLE_MANAGEMENT" != "" ]; then
+## Enable console features
+if [ "$ENABLE_CONSOLE_ALL" != "" ]; then
     # shellcheck disable=SC2086
-    cscli console enable console_management
+    cscli console enable -a
+else
+    if [ "$ENABLE_CONSOLE_MANAGEMENT" != "" ]; then
+        # shellcheck disable=SC2086
+        cscli console enable console_management
+    fi
+    if [ "$ENABLE_CONSOLE_CONTEXT" != "" ]; then
+        # shellcheck disable=SC2086
+        cscli console enable context
+    fi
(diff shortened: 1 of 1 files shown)


Simbiat:
Hm... I can't add labels, though 🤔 It should be `/kind enhancement` and `/area configuration` from what I understand

Simbiat:
Ah, so I just need to type those as commands. got it 👍

docker/docker_start.sh
LaurenceJJones, line 1: imo this is fine, but multiple calls to `cscli` can cause a slowdown if we could do

```
else
    CONSOLE_FLAGS=""
    if [ "$ENABLE_CONSOLE_MANAGEMENT" != "" ]; then
        # shellcheck disable=SC2086
        CONSOLE_FLAGS="$CONSOLE_FLAGS console_management"
    fi
    if [ "$ENABLE_CONSOLE_CONTEXT" != "" ]; then
        # shellcheck disable=SC2086
        CONSOLE_FLAGS="$CONSOLE_FLAGS context"
    fi
    if [ "$ENABLE_CONSOLE_TAINTED" != "" ]; then
        # shellcheck disable=SC2086
        CONSOLE_FLAGS="$CONSOLE_FLAGS tainted"
    fi
    if [ "$ENABLE_CONSOLE_MANUAL" != "" ]; then
        # shellcheck disable=SC2086
        CONSOLE_FLAGS="$CONSOLE_FLAGS manual"
    fi
    if [ "$ENABLE_CONSOLE_CUSTOM" != "" ]; then
        # shellcheck disable=SC2086
        CONSOLE_FLAGS="$CONSOLE_FLAGS custom"
    fi
   if [ "$CONSOLE_FLAGS" != "" ]; then
      # shellcheck disable=SC2086
      cscli console enable$CONSOLE_FLAGS
   fi
fi
```

Simbiat: Agreed. I did not know you can do that, though, since this is not mentioned in [knowledgebase](https://docs.crowdsec.net/docs/cscli/cscli_console_enable/) (as any of the console features, actually). Updated

Simbiat:
Not push or anything, but what does one need to do to get a merge around these parts? 🤠

PR #3466, opened 2025-02-13
refact: avoid use of defer calls in loops

mmetc:


pkg/apiserver/apic.go
```diff
@@ -363,6 +363,18 @@ func shouldShareAlert(alert *models.Alert, consoleConfig *csconfig.ConsoleConfig
 	return true
 }
 
+func (a *apic) sendBatch(ctx context.Context, signals []*models.AddSignalsRequestItem) error {
+	ctxBatch, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	_, _, err := a.apiClient.Signal.Add(ctxBatch, (*models.AddSignalsRequest)(&signals))
+	if err != nil {
```
LaurenceJJones, line 371: This is redundant if check you can just return err even if it's nil

blotus (approved):

PR #3483, opened 2025-02-24
Fix cp -n

michacassola:
Running the wizard script for installation gives me the following output a few times:

`cp: warning: behavior of -n is non-portable and may change in future; use --update=none instead`

So I updated the commands and comments regarding that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ ENTRYPOINT ["/bin/bash", "/docker_start.sh"]
 
 FROM slim AS full
 
-# Due to the wizard using cp -n, we have to copy the config files directly from the source as -n does not exist in busybox cp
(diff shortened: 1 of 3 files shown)


LaurenceJJones:
@mmetc  should we use `rsync` instead of `cp` or since it outside of the container we should stick to `cp`?

mmetc:
> @mmetc should we use `rsync` instead of `cp` or since it outside of the container we should stick to `cp`?

-1, you can't have a dependency on rsync just to install something

blotus:
Hello,

I don't think we can merge as is: depending on the OS, `cp` will accept different arguments.
For example, with a recent-ish `cp` running on ubuntu 24.04, the man page does mention that `-n` is deprecated and `--update=node` should be used.

But on Ubuntu 22.04 (which is still supported), the man page does not mention this and when attempting to use `cp --update=node`, `cp` returns an error: `cp: option '--update' doesn't allow an argument`

Issue #3492, opened 2025-03-03
enhance: [http acquis] add unix socket support


LaurenceJJones:
Since you may be using an upstream reverse proxy you may want to use unix sockets for this instead of proxying to a port.

PR #3499, opened 2025-03-05
enhance: add listen_socket to http acquisition

LaurenceJJones:
fix: #3492 

Allow users to configure a unix socket instead of using `listen_addr` only for http acquisitions.


mmetc:
@LaurenceJJones there is no check for empty ListeAddr or ListenSocket before running the respective goroutines?

mmetc (approved):

Issue #3316, opened 2024-11-07
Allow reader retry if acquisition file read fails


rr404:
### What would you like to be added?

For feature request please pick a kind label by removing `<!-- -->` that wrap the example lines below

/kind feature
Completely new feature not currently available

**Context**: Runnin crowdsec as non-root on Cloudways
**Issue**: "randomly" when log rotate, reader doesn't seem to have permission to read newly created file (maybe some weird race condition). It can work 3 days and crash the 4th or crash on first rotation, it seems random.  
I assume it's because we're not root that this happens, but maybe there are other cases where log reader fails this way

Could we have a config to say that a reader retries like N times every X seconds if fails to read a rotating log ?

My crowdsec logs show this (I obfuscated the path): 
`
time="2024-09-18T00:00:50Z" level=info msg="Re-opening moved/deleted file /<absolute_path>/nginx_wordpress-11111-11111.cloudwaysapps.com.access.log ..."

time="2024-09-18T00:00:50Z" level=warning msg="file reader of //<absolute_path>/nginx_wordpress-11111-11111.cloudwaysapps.com.access.log died : Unable to open file /<absolute_path>/nginx_wordpress-11111-11111.cloudwaysapps.com.access.log: open /<absolute_path>/nginx_wordpress-11111-11111.cloudwaysapps.com.access.log: permission denied" tail=/<absolute_path>/nginx_wordpress-11111-11111.cloudwaysapps.com.access.log type=file
`

### Why is this needed?

Be able to run CrowdSec on CloudWays managed servers (where we don't have root privileges)

david-garcia-garcia:
Same here, if a reader fails on log rotation, it will not retry and simply die.

LaurenceJJones:
> Same here, if a reader fails on log rotation, it will not retry and simply die.

this issue should only happen when running CrowdSec as a non root user, so is this the same in your case?

david-garcia-garcia:
@LaurenceJJones this is totally making our kubernetes deployment loose agent data ingest periodically and requires manual intervention to resucitate ingest again.

In our specific use case:

- Log files are made visible to crowdsec through a volume mount where the underlying storage is NFS. This volumes a Read/Write many.
- Both log generation and log rotation are taken care of from other containers. While this is happening (rotation), the log file might "glitch" (truncate/rotate, copy/rotate, whatever) and might no be accesible for an instant.
- An access denied error occurs, bit it is transient
- The agent does not recover

[...]

PR #3508, opened 2025-03-13
Fix monitorNewFiles for NFS + Remove dead tails from tail map

david-garcia-garcia:
An attempt to fix https://github.com/crowdsecurity/crowdsec/issues/3316

* Adds two new configuration options for file acquisition: DiscoveryPollInterval and DiscoveryPollEnable
* Current implementation of monitorNewFiles will not work with NFS because it does not emit events, now it relies on DiscoveryPollEnable and DiscoveryPollInterval if configured to periodically scan for files
* When a tail for a file dies, it is removed from the tailed files map, and will be picked up again on the next file discovery

### DIFF
--- a/pkg/acquisition/modules/file/file.go
+++ b/pkg/acquisition/modules/file/file.go
@@ -40,9 +40,11 @@ type FileConfiguration struct {
 	Filenames                         []string
 	ExcludeRegexps                    []string `yaml:"exclude_regexps"`
 	Filename                          string
-	ForceInotify                      bool  `yaml:"force_inotify"`
-	MaxBufferSize                     int   `yaml:"max_buffer_size"`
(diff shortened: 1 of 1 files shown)


pkg/acquisition/modules/file/file.go
david-garcia-garcia, line 1: I'm zero happy on how the refactor affects readability or the changes but there was already some duplicated logic that would have been further duplicated without a little refactoring...

david-garcia-garcia:
Was 100% I had all the tests passing before submiting the PR....

I fixed a race condition in the implementation of TestLiveAcquisition, but i now still have issues with the tests regarding permission denied, but it seems to be affecting the master branch also so it might not be specific to this PR.

blotus (approved):

PR #3522, opened 2025-03-19
feat(apic): add ApicAuth client and token re-authentication logic

sabban:
+ Introduced ApicAuth client for managing API authentication tokens.
+ Implemented token re-authentication logic in NewAPIC function.
+ Updated database schema to include ApicAuth table and related fields.
+ Added necessary hooks and interceptors for ApicAuth client.


buixor:
Let's do some further testing with HA setups before merging @sabban

mmetc (approved):

PR #3550, opened 2025-03-28
do not return an error if we cannot fetch allowlists when starting the appsec

blotus:


mmetc (approved):

LaurenceJJones:
Class it as enhancement rather than bugfix due to fact that LAPI/LP's should be on same version if this is not the case, then this is enhancement that this case can handle it.

PR #3558, opened 2025-04-01
CI: golangci-lint v2

mmetc:
The new .golangci.yml is not compatible with v1, but we have to move on at some point when the language server and IDE plugins are updated to support v2.

https://github.com/nametake/golangci-lint-langserver/issues/51
https://github.com/golang/vscode-go/issues/3732

https://golangci-lint.run/product/migration-guide/#command-migrate


sabban (approved):
lgtm

PR #3578, opened 2025-04-16
refact parser Init: argument types

mmetc:


mmetc:
@buixor was there a reason for the opaque argument to Init() here? Can I rename it NewUnixParserCtx() ?

blotus (approved):

PR #3593, opened 2025-04-29
lapi: return specific error if a unix socket path is too long for the OS

mmetc:
give the user a hint on what's happening (nothing more we can do) and make sure the tests don't fail for the same reason.


LaurenceJJones (approved):
lgtm

PR #3598, opened 2025-04-30
enhance: Remove docker acquis internal timer use docker events

LaurenceJJones:
Instead of polling the api every one second, we can instead use the events hook in docker this is a continuous event stream that notify us of a container event. We filter for `start` and `die` which is crate and remove, to which we then trigger a function call to get all containers and update the state.

This means this should  be less intensive since we dont have to constantly poll for all containers and only do so when we actually get a notification to do so.


mmetc:
@LaurenceJJones let me know what you think of my change, I didn't test in docker but the only difference should be that it's applied to the initial connection.

sabban (approved):
lgtm

PR #3635, opened 2025-05-16
refact: pkg/database decisions extract, rename

mmetc:


mmetc:
@blotus too many conflicts with the ip changes, I'll redo this one for 1.7

PR #3650, opened 2025-05-27
refact slicetools.Chunks() -> slices.Chunk()

mmetc:


mmetc:
the standard library panics if the chunk size is 0, even with an empty slice. so using this PR adds at least an in/else to each call because we can't use cmp.Or(size, len(slice))

PR #3654, opened 2025-05-30
Add retry to VictoriaLogs data source

### LINKED ISSUE #3653: CrowdSec exits if the VictoriaLogs data source goes down temporarily
### What happened?

I have a setup using fluent-bit to collect logs, ship them to a central VictoriaLogs instance, and then CrowdSec is ingesting logs from VictoriaLogs. Everything works great when started in the right order. CrowdSec will only start if VictoriaLogs is already running.

However, if VictoriaLogs goes down temporarily, CrowdSec dies.

I expected that CrowdSec would notice the lost connection, and at least retry a few times before dying. I hoped it would just keep retrying with some kind of back-off strategy.

Requires a working docker-compose setup.

1. Create a working folder.
2. Create a file vlogs.yaml with these contents
```yaml
source: victorialogs
mode: tail
log_level: info
url: http://vlogs:9428/
limit: 10
query: '*'
labels:
  type: other
```
3. Create a file compose.yaml with these contents
```yaml
services:
  vlogs:
    image: victoriametrics/victoria-logs:v1.23.2-victorialogs
  fluentbit:
    image: fluent/fluent-bit:4.0.2
    depends_on:
      - "vlogs"
    command:
      - "-i dummy"
      - "-p dummy='{_msg=\"This is a test message\"}'"
      - "-o http"
      - "-p host=vlogs"
      - "-p port=9428"
      - "-p uri=/insert/jsonline?_stream_fields=host,service&_msg_field=log&_time_field=date"
      - "-p format=json_lines"
      - "-p json_date_format=iso8601"
  crowdsec:
    image: crowdsecurity/crowdsec:v1.6.8
    depends_on:
      - "vlogs"
    volumes:
      - ./acquis.yaml:/etc/crowdsec/acquis.d/vlogs.yaml
```
4. Run docker-compose up

This should start three containers, with fluent-bit generating test log messages, VictoriaLogs collecting them, and CrowdSec ingesting them. The CrowdSec init takes a short delay, but eventually, you should just see lines from the fluent-bit container for each record.

5. At that point, in another terminal run docker stop <vlogs container> to stop the VictoriaLogs container.

You should see the VictoriaLogs container stop after a short delay (it tries to quit gracefully, but the CrowdSec request is hanging on, so it waits for a time out). Just after VictoriaLogs exits, the CrowdSec container will also exit with an error. You will see fluent-bit chugging along trying to send data, but failing.

6. Cleanup by running docker-compose down

I have forked crowdsecurity/crowdsec and create a branch victorialogs-retry-bug with a fix that is working for me.

### Crowdsec version

<details>

```console
$ cscli version
version: v1.6.8-f209766e
Codename: alphaga
BuildDate: 2025-03-25_15:56:53
GoVersion: 1.24.1
Platform: docker
libre2: C++
User-Agent: crowdsec/v1.6.8-f209766e-docker
Constraint_parser: >= 1.0, <= 3.0
Constraint_scenario: >= 1.0, <= 3.0
Constraint_api: v1
Constraint_acquis: >= 1.0, < 2.0
Built-in optional components: cscli_setup, datasource_appsec, datasource_cloudwatch, datasource_docker, datasource_file, datasource_http, datasource_journalctl, datasource_k8s-audit, datasource_kafka, datasource_kinesis, datasource_loki, datasource_s3, datasource_syslog, datasource_victorialogs, datasource_wineventlog
```

</details>

### OS version

<details>
These are on the host running Docker. My actual setup does not use Docker, and all the programs are running natively. The same thing happens there as well.

```console
# On Linux:
$ cat /etc/os-release
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
$ uname -a
Linux holmes 6.12.10-76061203-generic #202412060638~1743109366~22.04~1fce33b SMP PREEMPT_DYNAMIC Thu M x86_64 x86_64 x86_64 GNU/Linux
```

</details>

### Acquisition config

_No response_

### Config show

_No response_

### Prometheus metrics

_No response_

_No response_

thebondo:
The current VictoriaLogs data source will exit if the data source goes away once a tail based connection has been established. This causes the whole crowdsec service to exit as well.

This update fixes the Tail method on the VictoriaLogs client so that it tries to reconnect with a backoff if the connection is lost.

Resolves #3653


zekker6:
Hello @thebondo, do you think you will have some time to continue work on this PR?
If not, I can take over and create a PR with your commit and get it ready for review (attributing you as an original author).

thebondo:
Hello @zekker6. Yes, I can still contribute on this. I have a question about the approach for the final fix.

As I understand it, the tail endpoint is expected to stay open and report log entries when received (after some possible delay, which helps keep entries sorted by time). However, the connection is sometimes lost. Ideally, the tail consumer would like to see all log entries exactly once. However, there does not seem to be a good way to start a new tail query that will both

1. return all log entries received after that last query disconnected
2. not return any log entries received before the last query disconnected

The solution I have so far is to keep track of the latest timestamp seen while the tail connection is open. If the connection is lost, then compute `start_offset` based on the time of the new query, the latest timestamp, and possible a small constant to account for potential delays in the request process.

This has the drawback that it can potentially provide a log entry more than once. From looking at the VictoriaLogs source, the start time for the tail interval appears to be calculated as `time.Now() - offset - start_offset`. That means that the consumer cannot specify the start time exactly because it involves the time the request is actually handled by VictoriaLogs (which is the reason for adding a small constant to `start_offset`).

Does this all sound correct to you? If so, I will update the code to use `start_offset`. I will also remove the use of the `limit` argument, which is not documented for the tail endpoint and does not appear to do anything either.

zekker6:
@thebondo Thank you for an update!

> Does this all sound correct to you? If so, I will update the code to use start_offset. I will also remove the use of the limit argument, which is not documented for the tail endpoint and does not appear to do anything either.

Yes, this sounds correct to me.

zekker6:
Hello @thebondo, I've recently noticed that my CrowdSec instance was affected by this and wanted to check if you're planning to finalize this PR any time soon?
I would like to offer some help with updates discussed above if that will help.

mmetc:
For crowdsec 1.7.4+ we are in the process of converting datasources to a blocking Stream() API -- this means any error that is returned by the Stream function will automatically trigger a retry loop upstream. I'll see if anything needs to be handled differently for victorialogs, will keep this PR open but it's normally not the responsibility of the datasource module anymore.

PR #3658, opened 2025-06-02
modernize: replace legacy slice/map/range idioms with stdlib

mmetc:


pkg/types/appsec_event.go
```diff
@@ -187,11 +184,8 @@ func (w MatchedRules) ByTags(match []string) MatchedRules {
 func (w MatchedRules) ByTag(match string) MatchedRules {
 	ret := MatchedRules{}
 	for _, rule := range w {
-		for _, tag := range rule["tags"].([]string) {
-			if tag == match {
-				ret = append(ret, rule)
-				break
-			}
+		if slices.Contains(rule["tags"].([]string), match) {
+			ret = append(ret, rule)
```
LaurenceJJones, line 188: Missing the break inside the if statement

mmetc: yes, there's no loop anymore to break

PR #3666, opened 2025-06-04
refact pkg/cwhub

mmetc:
- pkg/cwhub: simplify url handling
 - pkg/cwhub: extract method collectSpecs()
 - pkg/cwhub: refact sync.go
 - pkg/cwhub: refact relativePathComponents
 - pkg/cwhub: don't call panic() in tests
 - pkg/cwhub: named return values
 - pkg/cwhub: compute hash on the fly
 - cwhub: stricter SafePath()


sabban (approved):
lgtm

PR #3671, opened 2025-06-05
refact csplugin: move ProfileAlert to models

mmetc:
this is a step to make csplugin support optional. next is tomb management, then add the build tag


sabban (approved):
lgtm

Issue #3704, opened 2025-06-26
Crash when querying for new decisions


david-garcia-garcia:
### What happened?

Bouncer is unable to query new decisions.

I have captured the following logs from the LAPI pod (there are many of these dumps):

```
crowdsec-lapi-bddc95c64-sqsq4:/var/lib/crowdsec/data/trace# cat crowdsec-crash.196145974.txt
error: runtime error: invalid memory address or nil pointer dereference
version: v1.6.5-72b4354b
BuildDate: 2025-02-10_14:52:20
GoVersion: 1.23.6
Platform: docker
goroutine 814489169 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:26 +0x5e
github.com/crowdsecurity/go-cs-lib/trace.(*traceKeeper).writeStackTrace(0x37c4000, {0x1f6b540, 0x37a13f0})
        github.com/crowdsecurity/go-cs-lib@v0.0.16/trace/trace.go:152 +0x16e
github.com/crowdsecurity/go-cs-lib/trace.WriteStackTrace(...)
        github.com/crowdsecurity/go-cs-lib@v0.0.16/trace/trace.go:42
github.com/crowdsecurity/crowdsec/pkg/apiserver.recoverFromPanic(0xc000baa100)
        github.com/crowdsecurity/crowdsec/pkg/apiserver/apiserver.go:100 +0x17c
panic({0x1f6b540?, 0x37a13f0?})
        runtime/panic.go:785 +0x132
github.com/crowdsecurity/crowdsec/pkg/database.(*Client).QueryNewDecisionsSinceWithFilters(0xc0000418c0, {0x26ea008, 0xc0008fc500}, 0x0, 0xc00538a840)
        github.com/crowdsecurity/crowdsec/pkg/database/decisions.go:314 +0x897
github.com/crowdsecurity/crowdsec/pkg/apiserver/controllers/v1.(*Controller).StreamDecisionNonChunked(0xc000c9de00, 0xc000baa100, 0xc004c7efc0, {0x6?, 0x3?, 0x227749f?}, 0xc00538a840)
        github.com/crowdsecurity/crowdsec/pkg/apiserver/controllers/v1/decisions.go:364 +0x612
github.com/crowdsecurity/crowdsec/pkg/apiserver/controllers/v1.(*Controller).StreamDecision(0xc000c9de00, 0xc000baa100)
[...]

PR #3706, opened 2025-06-27
fix: catch if since is nil

LaurenceJJones:
catch panic in #3704 if since is nil


mmetc:
Done in https://github.com/crowdsecurity/crowdsec/pull/3635 too

blotus:
Superseded by https://github.com/crowdsecurity/crowdsec/pull/3635

PR #3730, opened 2025-07-15
cscli setup: new service detection and configuration

mmetc:
- [x]  trim down and review detect.yaml


blotus (changes requested):
- [ ] Split the warning messages for custom/altered acquis files:
    -  For the altered warning, explicitly tell the user the content of the file will be replaced if they continue
 - [ ] Allow to pass the location of `detect.yaml` through an env var
    - If the env var exists but is empty, do not run the setup
 - [x] `detect-config` flag is not supported when calling `cscli setup` directly
 - [ ] Validate the content of the `detect.yaml` file in the CI (at least for syntax errors and basic things like no acquis filename duplication)

PR #3756, opened 2025-07-28
appsec: Add a `DropRequest` helper

blotus:
Add a new `DropRequest` helper to allow users to easily drop a request directly from a `pre_eval` hook.

This can be used for example to implement geoblocking:
```yaml
name: crowdsecurity/geoip-block
default_remediation: ban
pre_eval:
 - filter: IsInBand == true && GeoIPEnrich(req.RemoteAddr).Country.IsoCode not in ["FR", "US"]
   apply:
    - |
      let reason = Sprintf("Forbidden source country %s", GeoIPEnrich(req.RemoteAddr).Country.IsoCode );
      DropRequest(reason)
```

The helpers used to customized the response in case of a block are also exposed in the `pre_eval` hook to allow for customization of the response, like with any other rule.


blotus:
Superseded by #4016

PR #3757, opened 2025-07-29
docker: enforce volume use for /var/lib/crowdsec/data/

blotus:
Enforce using a volume for `/var/lib/crowdsec/data/`.

The check can be bypassed by setting the env var `CROWDSEC_BYPASS_DB_VOLUME_CHECK`.

Note that the check is also performed if the container is running without LAPI enabled, as datafiles are also stored in this location.

Also add a new `CROWDSEC_CONTAINER_ENV` for crowdsec to know in which context the container is running (docker, k8s, ...)


mmetc:
db_type is set but not used.

mmetc (approved):

PR #3784, opened 2025-08-05
remove the cscli_setup feature flag

mmetc:
This was never used outside testing, should not impact anyone. If someone really used the feature, it's not compatible anyway with the previous version and they'll get an error at runtime.


sabban (approved):
lgtm

PR #3787, opened 2025-08-05
fix apache2 detection on rpm based distribution

sabban:
fix apache2 detection on rpm based distribution


LaurenceJJones:
Think you got your changes mixed up with #3777

PR #3789, opened 2025-08-06
cscli setup improvements

mmetc:
- remove redundant detect options, keep only --force and --ignore
- use gopsutil instead of osinfo - it can recognize derived distributions
- detect systemd units even if masked or disabled

The goal is to detect what's installed and potentially configurable, regardless if it's running, temporary disabled or not yet configured


sabban (approved):
lgtm

PR #3801, opened 2025-08-08
fix: Appsec dont switch logger level at runtime

LaurenceJJones:
**What changed**
  - `pkg/appsec/coraza_logger.go`:
    - Stop toggling logger level in `crzLogEvent.Msg`.
    - `NewCrzLogger` now clones a per-runner `logrus.Logger` (copies output, formatter, hooks) and fixes it at Trace; emission is gated by our own threshold.
    - `WithLevel` updates only the logical threshold (does not mutate the underlying logger).
    - Preserves original entry fields as default context for all events.

- **Why**
  - Avoid race conditions and cross-runner log noise from runtime level changes.
  - Keep per-rule debug behavior via `rule_id` + `SetRuleDebug/GetRuleDebug` without mutating shared state.

- **Impact**
  - Safer, deterministic logging with the same output/format.
  - Per-rule debug continues to work as before, minus shared-level side effects.

- **Files**
  - `pkg/appsec/coraza_logger.go`


mmetc:
Here, can we use pkg/logging.SubLogger() ?

PR #3846, opened 2025-09-04
update kinesis datasource to aws sdk v2

mmetc:


mmetc:
After moving anon creds to the test build tag, CI reported an error with s.ctx being null, so I made a second attempt to remove ctx from the struct.

To do: review the usage of ctx vs cctx

we may eventually need to pass ctx to Configure* methods too

Issue #3336, opened 2024-11-20
Add SASL authentication for Kafka acquisition


acm-073:
### What would you like to be added?

/kind enhancement
Hello,

crowdsec already supports log acquisition from Kafka streams. I tried to hook it up to Azure Event Hub, which has a Kafka endpoint. There are two things that need to be done to make this work:

- Azure Event Hub requires SASL authentication, which is supported by the underlying kafka-go client but not available in the Kafka acquisition config
- Azure Event Hub requires a TLS connection, but no client certificate. The way the current kafka TLS config works is: either TLS config is present, then a client certificate is expected/required, or TLS config is not present, then a plain socket connection is established

It would be great if crowdsec would support log acquisition from Azure Event Hubs because it is a common scenario in Azure to stream logs through an event hub.

I have created a working prototype and would be happy to create a PR if you're interested.

Regards
Albrecht

### Why is this needed?

Streaming logs to Azure Event Hub is a common scenario in Azure cloud architectures. Being able to acquire logs directly from Event Hubs would greatly simplify the integration of crowdsec with Azure cloud applications.

LaurenceJJones:
With this

> Azure Event Hub requires a TLS connection, but no client certificate. The way the current kafka TLS config works is: either TLS config is present, then a client certificate is expected/required, or TLS config is not present, then a plain socket connection is established

Do you want to just just load the CA certificate then just so I can understand? as if the ca cert is within the system, then you shouldnt need to define it here unless you can tell us if it self signed as we dont know how azure does its TLS.

acm-073:
> With this
> 
> > Azure Event Hub requires a TLS connection, but no client certificate. The way the current kafka TLS config works is: either TLS config is present, then a client certificate is expected/required, or TLS config is not present, then a plain socket connection is established
> 
> Do you want to just just load the CA certificate then just so I can understand? as if the ca cert is within the system, then you shouldnt need to define it here unless you can tell us if it self signed as we dont know how azure does its TLS.
> 
> Edit: ahh I think I see the issue.

Let me detail some more:
- if the TLS config in kafka acquisition is present, then the acquisition tries to load a client cert which it would use for client authentication. If none is configured, then the client configuration fails. A client cert is NOT required for Azure Event Hub - only server authentication (where the server presents a SSL certificate) is used. For this, of course the client needs to have the "public" CA certificates, but they are usually available in the environment
[...]

PR #3867, opened 2025-09-08
Implement "PLAIN" SASL mechanism

acm-073:
This improvement implements issue https://github.com/crowdsecurity/crowdsec/issues/3336
However, with a limited scope: only the "PLAIN" SASL mechanism is supported so far.

The code has been tested with Azure EventHub Kafka endpoint and works fine there.


acm-073:
need to re-create PR from dedicated branch so that some other crap does not get pulled in

PR #3873, opened 2025-09-09
feat: add timeout resilience to Docker acquisition (socket proxy errors)

LaurenceJJones:
- Add configurable retry mechanism for Docker log readers
- Implement per-container log options to prevent race conditions
- Add intelligent consecutive failure tracking with reset logic
- Use typed errors for better failure classification
- Preserve EOF-resilient behavior with proper Since timestamp management

Fixes socket proxy timeout issues that previously caused containers to be permanently removed from monitoring after temporary failures.

### DIFF
--- a/pkg/acquisition/modules/docker/docker.go
+++ b/pkg/acquisition/modules/docker/docker.go
@@ -30,25 +30,33 @@ import (
 	"github.com/crowdsecurity/crowdsec/pkg/types"
 )
 
+// Error types for distinguishing different failure modes
+var (
+	ErrConnectionSuccessfulThenLost = errors.New("connection was successful but then lost")
+	ErrConnectionLostImmediately    = errors.New("connection lost immediately")
+)
+
 type DockerConfiguration struct {
 	configuration.DataSourceCommonCfg `yaml:",inline"`
 
-	CheckInterval       string   `yaml:"check_interval"`
-	FollowStdout        bool     `yaml:"follow_stdout"`
-	FollowStdErr        bool     `yaml:"follow_stderr"`
-	Until               string   `yaml:"until"`
-	Since               string   `yaml:"since"`
-	DockerHost          string   `yaml:"docker_host"`
-	ContainerName       []string `yaml:"container_name"`
-	ContainerID         []string `yaml:"container_id"`
-	ContainerNameRegexp []string `yaml:"container_name_regexp"`
-	ContainerIDRegexp   []string `yaml:"container_id_regexp"`
-	ServiceName         []string `yaml:"service_name"`
-	ServiceID           []string `yaml:"service_id"`
-	ServiceNameRegexp   []string `yaml:"service_name_regexp"`
-	ServiceIDRegexp     []string `yaml:"service_id_regexp"`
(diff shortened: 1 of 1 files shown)


mmetc:
@blotus can we talk about this before approving. I think we could use cenkalti/backoff

pkg/acquisition/modules/docker/docker.go
```diff
@@ -924,34 +1035,62 @@ func (d *DockerSource) TailContainer(ctx context.Context, container *ContainerCo
...
+			return nil
+		}
+
+		// Check service health to determine if we should retry or give up
+		serviceHealthy := d.isServiceStillRunning(ctx, service)
+
+		if !serviceHealthy {
+			// Service was removed - don't retry, remove from monitoring
+			service.logger.Infof("service %s no longer exists, removing from monitoring: %v", service.Name, err)
+			deleteChan <- service
```
LaurenceJJones, line 1069: im un decided if the container should self delete, since we get docker events now we already get this.

my fear is that it may cause a race condition like:

docker event die -> docker event alive -> container self delete

pkg/acquisition/modules/docker/docker.go
```diff
@@ -872,15 +909,90 @@ func ReadTailScanner(scanner *bufio.Scanner, out chan string, t *tomb.Tomb) erro
...
+			return nil
+		}
+
+		// Check container health to determine if we should retry or give up
+		containerHealthy := d.isContainerStillRunning(ctx, container)
+
+		if !containerHealthy {
+			// Container is dead/stopped - don't retry, remove from monitoring
+			container.logger.Infof("container %s is no longer running, removing from monitoring: %v", container.Name, err)
+			deleteChan <- container
```
LaurenceJJones, line 973: im un decided if the container should self delete, since we get docker events now we already get this.

my fear is that it may cause a race condition like:

docker event die -> docker event alive -> container self delete

pkg/acquisition/modules/docker/docker.go
```diff
@@ -924,34 +1035,62 @@ func (d *DockerSource) TailContainer(ctx context.Context, container *ContainerCo
...
 			// The only known case currently is when using docker-socket-proxy (and maybe a docker daemon restart)
-			d.logger.Debugf("readerTomb dying for container %s, removing it from runningContainerState", container.Name)
-			deleteChan <- container
-			// Also reset the Since to avoid re-reading logs
-			d.Config.Since = time.Now().UTC().Format(time.RFC3339)
-			d.containerLogsOptions.Since = d.Config.Since
-
-			return nil
+			container.logger.Debugf("readerTomb dying for container %s, connection lost", container.Name)
+			readerTomb.Kill(nil)
```
LaurenceJJones, line 1039: we already dying we dont need to call `kill`?

pkg/acquisition/modules/docker/docker.go
mmetc: "we get docker events" -- if the proxy was down during the container death, we are not notified of that and remove it when we can confirm it's missing, right?

PR #3884, opened 2025-09-11
[WAF] do not iterate over all transaction variables for nothing // swap alert generation + event send order

blotus:


buixor:
- swap alert generation + event send order : is meant to prevent a race condition under heavy load with OOB rules that would trigger a concurrent map read/write when the generated (oob) event is being parsed while simultaneously being accessed for alert generation

PR #3901, opened 2025-09-17
fix #3896: drop cached token if username has changed

### LINKED ISSUE #3896: [CAPI] JWT auth cache issue
### What happened?

# Bug: Enrollment breaks after restoring `online_api_credentials.yaml` due to cached JWT

## Summary

When migrating or reinstalling CrowdSec, restoring a backed-up `online_api_credentials.yaml` after a fresh install can leave the agent using a **stale, cached JWT** stored in the local database. On startup, CrowdSec does **not** verify that the JWT’s username matches the credentials in `online_api_credentials.yaml`, so it reuses the cached token. This causes enrollment issues and mis-attributed alerts until the JWT expires.

## Impact / Side effects

* **Alerts are attributed to the wrong username** (from the cached JWT) for the period between token creation and token expiry.
* **Enrollment remains offline** until the cached JWT expires or is cleared.
* Running **console enrollment appears as a new machine**, because the JWT corresponds to the old username.

## Steps to Reproduce

1. Install CrowdSec and enroll with the Console, generating credentials and a JWT (token is cached in the local DB).
2. Uninstall or migrate CrowdSec.
3. Reinstall CrowdSec (this registers new credentials and authenticates, caching a new JWT).
4. After reinstall, restore your backed-up `online_api_credentials.yaml` (containing the *old* username/password).
5. Restart CrowdSec.

**Observed:** On restart, the agent reuses the cached JWT from the DB without checking whether the JWT’s username matches the credentials file on disk.

## Workarounds

* **Clear the cached JWT** from the local database, then restart CrowdSec so it re-authenticates.
* **Wait a minute after installation** since the cache JWT is compared agaisnt creation time of > 1 minute, if you wait at least a minute before restarting the CrowdSec service it will not reuse the cached JWT.

## Expected Behavior

* On startup (or when credentials change), the agent should **validate the cached JWT** against the username in `online_api_credentials.yaml`.
* If there’s a mismatch or the token cannot be validated, the agent should **discard the cached JWT** and **re-authenticate** using the credentials file so enrollment comes back under the correct account/machine.

## Actual Behavior

* The agent **reuses the cached JWT** from the DB even when the credentials file has been replaced with different (restored) credentials.
* Enrollment stays broken and alerts—if any—are sent under the wrong username until the JWT expires.

Follow the steps:
1. Install CrowdSec and Enroll into the console.
2. Move `mv /etc/crowdsec/online_api.credentials.yaml{,.bak}`
3. `cscli capi register` and wait at least a minute.
4. To simulate a "fresh install" just simply restart CrowdSec to ensure the JWT is populated in database
5. `mv /etc/crowdsec/online_api.credentials.yaml{.bak,}`
6. `systemctl restart crowdsec`

Now the JWT token in the database is reused even though the file on disk has changed, in half an hour the console will report the engine has not contacted and classified as offline.

_No response_

### Crowdsec version

<details>

```console
$ cscli version
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
# paste output here

# On Windows:
# paste output here
</details>


### Config show

<details>

```console
$ cscli config show
# paste output here
```

</details>

### Prometheus metrics

<details>

```console
$ cscli metrics
# paste output here
```

</details>

<details>

</details>

mmetc:


mmetc:
closing, need a better solution

PR #4000, opened 2025-10-24
move hubupdate.sh to libexec

mmetc:


mmetc:
> Just one quick question will the old file be deleted alongside the upgrade on debian-like distributions ?

it's not on etc so yes

PR #4004, opened 2025-10-24
Docker config.yaml - set SQLite use_wal to true

gonsalvesc:
This pull request sets the the default value of SQLite use_wal to true. This removes the below warning at startup.
/kind enhancement
/area configuration

```
level=warning msg="sqlite is not using WAL mode, LAPI might become unresponsive when inserting the community blocklist"
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -206,7 +206,7 @@ linters:
 
     maintidx:
       # raise this after refactoring
-      under: 18
+      under: 20
 
     misspell:
       locale: US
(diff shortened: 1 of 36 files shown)


LaurenceJJones:
> This pull request sets the the default value of SQLite use_wal to true. This removes the below warning at startup.
> 
> ```
> level=warning msg="sqlite is not using WAL mode, LAPI might become unresponsive when inserting the community blocklist"
> ```

The problem is we cant set it to `true` by default, as somebody could be storing the sqlite from a NFS mount and WAL should not be used with NFS shares.

PR #4007, opened 2025-10-27
feat: Add ParseKVLax for Flexible Key-Value Parsing

LaurenceJJones:
This PR is mainly aimed at supporting a looser key value parsing for Unifi CEF lines (even though the tests have full log line in practice we will parse out the message so it more aimed). Our current implementation for key value only parses:

- `foo=bar`
- `bar="foo bar"`

## Overview
This PR introduces `ParseKVLax`, a new key-value parser that complements the existing `ParseKV` function. While `ParseKV` maintains strict regex-based parsing for backward compatibility, `ParseKVLax` uses a scanner-based approach to support **unquoted values with spaces** and other complex log formats.

### New Function: `ParseKVLax`
- **Scanner-based approach**: Finds valid `key=` patterns and determines value boundaries intelligently
- **Supports multi-word values**: `UNIFIhost=Express 7 port=443` now correctly parses as `{"UNIFIhost": "Express 7", "port": "443"}`
- **Robust quote handling**: Properly filters out `key=` patterns inside quoted values
- **Escape sequence support**: Handles escaped quotes (`\"`) and backslashes (`\\`)

### Key Features
- ✅ **Backward Compatible**: Original `ParseKV` unchanged - all existing code continues to work
- ✅ **Opt-in**: Use `ParseKVLax` only where needed for loose parsing
- ✅ **Well-tested**: 13 comprehensive test cases covering edge cases
- ✅ **Production-ready**: Handles CEF logs, iptables output, keycloak JSON values, etc.

### Use Cases
- **CEF logs**: `msg=User login successful` → captures full phrase
- **iptables**: `RES=0x00 SYN URGP=0` → captures flags as part of values
- **UniFi**: `UNIFIhost=Express 7` → captures version as part of value


LaurenceJJones:
linked too: https://github.com/crowdsecurity/hub/issues/940

PintjesB:
Is this still being tested?

mmetc (approved):

PR #4020, opened 2025-11-03
Update go-re2 to 1.10.0

blotus:
Update https://github.com/wasilibs/go-re2 to the latest version.

This should fix a crash on Windows and also speed up `cscli` startup, as re2 is no longer initialized with an `init` function.


mmetc:
cscli was already fixed in the current version -- in fact, the pfsense binary now builds with wasm to avoid build conflicts with a different version of abseil

PR #4021, opened 2025-11-03
option api.server.disable_usage_metrics_export

mmetc:


mmetc:
todo in a future PR: store a reference to the lapi config in Controller / V1Controller as well and remove the redundant fields

PR #4075, opened 2025-11-20
New tail_mode and stat_poll_interval settings to support not keeping file handles permanently open

### LINKED ISSUE #4074: Support not keeping file handles permanently open when using file acquisition

/kind feature

Add configuration options to allow file acquisition to work without permanently keeping file handles open.

### Why is this needed?

The motivation is to support tailing files that have a SMB/SAMBA underlying storage. SMB/SAMBA implementation will **prevent** tailed files from being deleted if a file handle is open.

A special flag can be used to avoid this FILE_SHARE_DELETE but it is only available on Windows, plus it won't be available if SAMBA/SMB is used underneath an abstraction layer like Kubernetes CSI.

https://learn.microsoft.com/en-us/rest/api/storageservices/managing-file-locks

> If this flag [FILE_SHARE_DELETE] isn't specified, any request to delete the file will fail, until the file is closed.

https://www.samba.org/samba/docs/4.5/man-html/smb.conf.5.html

> When set to yes (default): Samba checks file system permissions directly and denies deletion if permissions don't allow it This aligns with Windows semantics where deletion permissions are verified at the time of the delete request

david-garcia-garcia:
Fixes https://github.com/crowdsecurity/crowdsec/issues/4074

This PR adds two new configuration options:

* tail_mode: default or stat
* stat_poll_interval: how frequently should files be checked for changes when tail_mode=stat

To avoid introducing regresions, I built an abstraction to wrap the actual tail implementation depending on tail_mode. When using tail_mode the previous implementation used. When using tail_mode=stat, a new tail implementation that does not permamently keep the file handle open is used.


mmetc:
Hi @david-garcia-garcia,

Thanks for the PR!

We're close to releasing 1.7.4, so this won't be merged before 1.7.5.

As you can see from the last few commits in the master branch, we are
in the middle of a global refactor of the acquisition:
the current non-blocking StreamingAcquisition() method is being replaced
by a synchronous Stream() method. This lets us drop tombs in favor of
context-based cancellation (more errgroup, waitgroup and fewer channels),
and bubble all retryable errors up to a centralized backoff loop.

For reference, the journalctl datasource is already migrated and merged; syslog
is migrated but not merged yet. Docker and file are likely to be next, shortly
after the 1.7.4 release.

Since your PR is a wrapper around the tail package there shouldn't be a lot of conflicts
if we postpone the review and adapt it to the new interface and concurrency model once they
are in place.

Thanks again for your work and your patience!

david-garcia-garcia:
@mmetc I see that file acquisition has not been upgraded yet. I did a quick attempt here https://github.com/crowdsecurity/crowdsec/pull/4280 but the underlying library used for tail does not support context. I am thiking it could be more straightforward to replace that tail library with a custom made one that optionally supports the behaviour that this PR attempts to support.

mmetc:
A cross platform tail rewrite would be tricky. I think it's easier to wrap tomb behavior around context. What do you think @blotus

david-garcia-garcia:
I took a look at the library used and the implementation was not that big, plus it had several features that we are not using so it could be further simplified.

Attempted to fully replace it here:

https://github.com/crowdsecurity/crowdsec/pull/4280

PR #4080, opened 2025-11-23
Fix avoidable prometheus metrics cardinality

g00g1:
I am running decently sized crowdsec setup (more than 100k active decisions simultaneously) with significant proportion of bans added using cscli. Also, my setup requires collection of Prometheus full metrics to collect historical data regarding popularity of some decisions.

All of the above contributes to extremely large size of resulting metrics endpoint response (more than tens of megabytes), as well as RAM overhead (see screenshot 1). After manual inspection I have noticed that the problematic metrics include `cs_lapi_machine_requests_total` like these:

```
cs_lapi_machine_requests_total{machine="[redacted]",method="GET",route="/v1/allowlists/check/192.0.2.1"} 1
cs_lapi_machine_requests_total{machine="[redacted]",method="GET",route="/v1/allowlists/check/192.0.2.2"} 1
```

I would like to argue usefulness of exposing raw URL in metrics including query parameters like IP addresses specified above.

Instead, I propose to use `func (c *Context) FullPath() string` instead (available since gin-gonic/gin v1.5.0 so this change could be backported). It returns a matched route full path instead of raw URL as in original request, therefore preventing unnecessary overhead and too high labels cardinality for setups configured to offer full Prometheus metrics level without aggregation (which in fact removes some metrics instead).

I expect this to be not a breaking change.

<details><summary>Screenshot 1</summary>
<p>

image

Before / after this patch in my setup

</p>
</details>

### DIFF
--- a/pkg/apiserver/controllers/v1/metrics.go
+++ b/pkg/apiserver/controllers/v1/metrics.go
@@ -32,7 +32,7 @@ func PrometheusMachinesMiddleware() gin.HandlerFunc {
 		if machineID != "" {
 			metrics.LapiMachineHits.With(prometheus.Labels{
 				"machine": machineID,
-				"route":   c.Request.URL.Path,
+				"route":   c.FullPath(),
 				"method":  c.Request.Method,
 			}).Inc()
 		}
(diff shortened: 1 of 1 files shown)


mmetc (approved):

mmetc:
Good catch, thanks!

PR #4095, opened 2025-11-27
allowlists: apply items to existing decisions in batch

blotus:
If the user has a ton of allowlist items (eg, > 10k), one DB request was made per item, which was very slow.

Batch the checks to reduce the amount of requests and load on the DB.


Solvik:
I triggered the case with an allowlist of 150 items

PR #4100, opened 2025-11-30
add share_appsec console option and IsAppsecAlert() helper

### LINKED ISSUE #3523: Appsec events is missing some data to easily filter them in profiles
It's  a bit awkward to filter an appsec event in the profiles (to send a notif for example):
 - There's no `Type` or similar in an alert: this only exists in the events
 - Events are not populated in an appsec alert (this can make sense, as we don't really have an event)
 - Even if events were populated, `evt.GetType()` does not handle the `APPSEC` type

Currently, the workaround is to filter on the message, but this seems very fragile:
```
Alert.Message contains "AppSec block:"
```

stlalpha:
Add support for sharing AppSec WAF alerts with the console and a helper method for filtering them in profiles.

The `share_appsec` option in console.yaml controls whether AppSec alerts are pushed to CAPI. It defaults to false.

```bash
cscli console enable appsec
```

The `Alert.IsAppsecAlert()` helper can be used in profile filters:

```yaml
filters:
  - Alert.IsAppsecAlert() == true
```

AppSec alerts are identified by their message prefix ("WAF block:" or "WAF out-of-band match:") which is set in `pkg/acquisition/modules/appsec/utils.go`. This catches all AppSec alerts including custom rules.

Closes #3523


blotus:
Hello,

Thanks for the PR, but I'm not sure about the approach.

First, about sending the alerts: I don't want to add a specific settings for WAF alerts (especially one defaulting to false), it would just make things more confusing and users would be more likely to ask why their WAF alerts are not showing up in the console, and I'm not sure if it would make sense for someone to want to share WAF alerts but not log-based ones (or the other way around).

This is already controlled by the existing `custom` (note that rules written in seclang are always considered to be custom at the moment) or `tainted` settings for custom WAF rules, or by setting `api.server.online_client.sharing` to false if you don't want to send anything at all (including alerts from rules we provide).
In the very unlikely event we'd want to actually implement this, the `official`/`custom`/`tainted` granularity would likely need to be kept.

Regarding the helper: It's pretty much just "hiding" the string match on the scenario name.
This is definitely not something we want to do like this: matching on the scenario name is very fragile and could break at any time (and I realize writing this, it's something we might already have broken for some users, as the prefix used in WAF alerts has changed a few releases ago).

We already have a field in the event struct that indicates where the event is coming from (see https://github.com/crowdsecurity/crowdsec/blob/master/pkg/acquisition/modules/appsec/utils.go#L112), but it's part of the event, not the actual alert.
Adding this to the alert would require changes to both the LAPI and CAPI schema, DB models, and likely changes in our backend, so it's something that needs to be carefully considered first.

Issue #4160, opened 2025-12-18
[Docker container] 403 Forbidden HTTP error when downloading GeoLite2-City.mmdb


KyoTensu:
### What happened?

While trying to access a service through Pangolin, which is set up with a crowdsec container alongside it, I got errors and tried to restart Pangolin's docker compose. Now, the crowdsec container always start unhealthy, with the following logs:

```console
/var/lib/crowdsec/data was found in a volume
Local agent already registered
Check if lapi needs to register an additional agent
sqlite database permissions updated
/etc/crowdsec was found in a volume
Running hub update
Skipping hub update, index file is recent
/var/lib/crowdsec/data was found in a volume
Running hub upgrade
Action plan:
🔄 check & update data files

downloading https://hub-data.crowdsec.net/mmdb_update/GeoLite2-City.mmdb
level=warning msg="Failed to check last modified: bad HTTP code 403 for https://hub-data.crowdsec.net/mmdb_update/GeoLite2-City.mmdb" url="https://hub-data.crowdsec.net/mmdb_update/GeoLite2-City.mmdb"
Error: cscli hub upgrade: while downloading data for /etc/crowdsec/parsers/s02-enrich/geoip-enrich.yaml: while getting data: bad HTTP code 403 for https://hub-data.crowdsec.net/mmdb_update/GeoLite2-City.mmdb
Running: cscli  parsers install "crowdsecurity/docker-logs" 
Nothing to install or remove.
Running: cscli  parsers install "crowdsecurity/cri-logs" 
Nothing to install or remove.
Running: cscli  collections install "crowdsecurity/traefik" 
Nothing to install or remove.
Running: cscli  collections install "crowdsecurity/appsec-virtual-patching" 
Nothing to install or remove.
Running: cscli  collections install "crowdsecurity/appsec-generic-rules" 
Nothing to install or remove.
Running: cscli  parsers install "crowdsecurity/whitelists" 
Nothing to install or remove.
time="2025-12-18T16:40:17Z" level=info msg="Enabled feature flags: none"
time="2025-12-18T16:40:17Z" level=info msg="Crowdsec v1.7.4-db3efdbf"
time="2025-12-18T16:40:17Z" level=info msg="Loading CAPI manager"
time="2025-12-18T16:40:17Z" level=info msg="flushed 1/430 alerts because they were created 168h0m0s ago or more" module=db
time="2025-12-18T16:40:17Z" level=info msg="attempt 1 out of 2"
[...]

Issue #3603, opened 2025-04-30
SOLUTION - Unable to persist CrowdSec Local API key on container restart in Caddy.


aidenjanzen:
# Unable to persist CrowdSec Local API key on container restart in Caddy.

I had a lot of trouble with this and have tried multiple solutions over a couple of months. I wanted to post my solution on GitHub if others ran into the same issue.

I am running CrowdSec in a Docker container alongside a Caddy bouncer built from [hslatman’s repository](https://github.com/hslatman/crowdsec-caddy-bouncer) built with xCaddy.

Initially, everything works fine:

Generate an API key with `sudo docker exec crowdsec cscli bouncer add caddy-bouncer`.

Add the key to the Docker Compose and Caddyfile.

Restart Caddy and CrowdSec. 

And CrowdSec starts blocking malicious requests to Caddy.
## Issue: 

*However*, after performing an update or running:

`sudo docker compose down`
`sudo docker compose up`

The API key appears invalid, and the following error occurs:
```json
{"level":"error","ts":1730394008.01895,"logger":"crowdsec","msg":"auth-api: auth with api key failed return nil response, error: dial tcp [::1]:8089: connect: connection refused","instance_id":"a417f79a","address":"http://localhost:8089/","error":"auth-api: auth with api key failed return nil response, error: dial tcp [::1]:8089: connect: connection refused"}
{"level":"error","ts":1730394008.0189993,"logger":"crowdsec","msg":"failed to connect to LAPI, retrying in 10s: Get \"http://localhost:8089/v1/decisions/stream?startup=true\": dial tcp [::1]:8089: connect: connection refused","instance_id":"a417f79a","address":"http://localhost:8089/","error":"failed to connect to LAPI, retrying in 10s: Get \"http://localhost:8089/v1/decisions/stream?startup=true\": dial tcp [::1]:8089: connect: connection refused"}
```

This shows that the Caddy bouncer **cannot** connect to the CrowdSec Local API (LAPI) at container start.

As a workaround, I had to manually re-run `cscli bouncer add caddy-bouncer`, which generates a new API key, then manually update configs, and restart Caddy. This would happen **multiple** times a week.

**The goal was to avoid this manual intervention.**

## Key Findings:
- Docker Setup:

Tested fixed IP addresses inside a custom Docker network and specified a subnet for both services. Did not change the LAPI issue.

Made sure that data persistence is correctly set up through volumes:

```yaml
[...]

PR #4161, opened 2025-12-18
docker: add default healthcheck to container images

LaurenceJJones:
Adds HEALTHCHECK instruction using `cscli lapi status` to verify the local API is responding.

Related to #4160
Related to #3603


LaurenceJJones:
Just thought if LAPI only container then what? Need to test

mmetc:
what if it's a log processor only? it will test the health of the LAPI container, not the LP.

LaurenceJJones:
> what if it's a log processor only? it will test the health of the LAPI container, not the LP.

Well it will test the health between lp and lapi so technically not bad, but we don't really have a way to know I'm healthy for lp only.

mmetc:
> we don't really have a way to know I'm healthy for lp only.

maybe metrics? log processors still expose them when lapi is disabled, and they shouldn't kick in unless LP is healthy. I'm not saying we should do that, just that they are available.

PR #4219, opened 2026-01-21
stateless bucket for redis like backend

sabban:


buixor:
not the right aproach as we discussed tgth

PR #4221, opened 2026-01-22
k8s log acquisition datasource

sabban:
k8s log acquisition datasource


mmetc:
quick 2¢, I didn't follow the logic yet:

 - the package is not imported from pkg/acquisition
 - name of the package/datasource is way too verbose for comfort
 - log.Fatal should just return errors
 - vestigial presence of tomb
 - there is an empty .keep file
 - references to DockerAcquisition, DockerDatasource
 - does one shot acquisition make sense for this DS?
 - the schema file should go in pkg/acquisition/schemas/

sabban:
@copilot review

pkg/acquisition/modules/kubernetes/init.go
```diff
@@ -0,0 +1,20 @@
+package kubernetes
```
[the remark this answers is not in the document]
sabban: It's not the case, loki, victorialogs don't have acquisition in the name

sabban:
@copilot resolve the merge conflicts in this pull request

pkg/acquisition/modules/kubernetes/run.go
```diff
@@ -0,0 +1,292 @@
...
+				return nil
+			}
+			if err := onLine(sc.Text(), ns+"/"+pod+"/"+container, labels, metricsLevel, out); err != nil {
+				return err
+			}
+		}
+		if ctx.Err() != nil {
+			return nil
+		}
+		return sc.Err()
```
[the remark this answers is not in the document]
sabban: I think logline > 64kb is not current

pkg/acquisition/modules/kubernetes/run.go
```diff
@@ -0,0 +1,292 @@
...
+			})
+		}
+		cw.Wait()
+	})
+	return cancel
+}
+
+func (s *Source) tailPod(ctx context.Context, cs *kubernetes.Clientset, p *corev1.Pod, out chan pipeline.Event, wg *sync.WaitGroup, mu *sync.Mutex, cancels map[types.UID]context.CancelFunc) {
+	if p.Status.Phase != corev1.PodRunning {
+		s.logger.Debugf("SKIP tailPod(non-running) %s", podRef(p))
```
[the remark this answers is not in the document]
sabban: This is not useful, as the DeleteFunc hook already handles cancellation via stopPod, and the UpdateFunc calling tailPod on a non-Running pod is a no-op.

pkg/acquisition/modules/kubernetes/run.go
```diff
@@ -0,0 +1,325 @@
...
+	if client == nil {
+		return errors.New("kubernetes client is not initialized")
+	}
+
+	sinceTime := metav1.NewTime(time.Now().UTC())
+	req := client.CoreV1().Pods(ns).GetLogs(pod, &corev1.PodLogOptions{
+		Container:  container,
+		Follow:     true,
+		Timestamps: false,
+		SinceTime:  &sinceTime,
```
blotus: This won't actually fix potential replay if a tail stops and restarts: we are using the same sinceTime, so we will be getting the same logs again.

Maybe we should use `TailLines` instead and avoid dealing entirely with timestamps ?

pkg/acquisition/modules/kubernetes/run.go
```diff
@@ -0,0 +1,325 @@
...
+		defer func() {
+			s.mu.Lock()
+			delete(s.cancels, pod.UID)
+			s.mu.Unlock()
+		}()
+		var cw sync.WaitGroup
+		for _, cont := range pod.Spec.Containers {
+			cw.Go(func() {
+				err := s.followPodLogs(podCtx, pod.Namespace, pod.Name, cont.Name, out, s.processLine)
+				if err != nil {
```
blotus, line 288: Do we need to retry on error ?

pkg/acquisition/modules/kubernetes/config.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if c.KubeConfigFile == "" {
+		if home, err := os.UserHomeDir(); err == nil {
+			c.KubeConfigFile = filepath.Join(home, ".kube", "config")
+		}
+	}
+}
+
+func (c *Configuration) Validate() error {
+	if c.Selector == "" {
```
blotus, line 59: We could also use `labels.Parse()` to validate if the label syntax is ok.

blotus (changes requested):
This is also missing some runtime tests (there's only configuration tests, and they are pretty barebones)

pkg/acquisition/modules/kubernetes/config.go
sabban: done

pkg/acquisition/modules/kubernetes/run.go
sabban: Yes, and using the same loop as the EOF retry loop

pkg/acquisition/modules/kubernetes/run.go
sabban: done

blotus (approved):

PR #4227, opened 2026-01-23
Created endpoint GET /scenarios on LAPI

JGeek00:
This PR adds a new endpoint to LAPI. It returns a list with the scenarios that have been added to Crowdsec. It uses the same code that was already available for CAPI and returns the same data on LAPI.

**Endpoint**
GET /v1/scenarios

**Response example**
```
[
    "crowdsecurity/ssh-bf",
    "crowdsecurity/ssh-cve-2024-6387",
    "crowdsecurity/ssh-generic-test",
    "crowdsecurity/ssh-refused-conn",
    "crowdsecurity/ssh-slow-bf",
]
```

### DIFF
--- a/pkg/apiserver/apic.go
+++ b/pkg/apiserver/apic.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
-	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -89,27 +88,7 @@ func randomDuration(d time.Duration, delta time.Duration) time.Duration {
 }
 
 func (a *apic) FetchScenariosListFromDB(ctx context.Context) ([]string, error) {
-	scenarios := make([]string, 0)
-
-	machines, err := a.dbClient.ListMachines(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("while listing machines: %w", err)
-	}
-	// merge all scenarios together
-	for _, v := range machines {
-		machineScenarios := strings.Split(v.Scenarios, ",")
-		log.Debugf("%d scenarios for machine %d", len(machineScenarios), v.ID)
-
-		for _, sv := range machineScenarios {
-			if !slices.Contains(scenarios, sv) && sv != "" {
-				scenarios = append(scenarios, sv)
-			}
-		}
-	}
-
-	log.Debugf("Returning list of scenarios : %+v", scenarios)
-
-	return scenarios, nil
+	return a.dbClient.FetchScenariosListFromDB(ctx)
 }
 
 func decisionsToAPIDecisions(decisions []*models.Decision) models.AddSignalsRequestItemDecisions {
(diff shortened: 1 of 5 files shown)


LaurenceJJones:
Hey, can you add some reasons or thoughts on why this is needed? cause there some flaws, it uses the jwtauth but returns all machine scenarios not the single machine that has authed.

Plus the machine that is authing already knows it own scenarios as they are stored on disk.

JGeek00:
Hey @LaurenceJJones, that's because I'm using a software to monitor the Crowdsec instance, and it relies on LAPI. I want it to have the complete list of scenarios so that I don't have to check them manually on the console. There shouldn't be security issues because it relies on the same auth method that already exists, and with LAPI you can access more more important data like the decisions.

LaurenceJJones:
The question wasnt about security, it was more jwtauth already knows which machine authenticated, returning all scenarios for every machine doesnt let you know anything useful because it all merged into the same data, your primarily thinking of your feature as running a solo instance but LAPI can have multiple machines.

JGeek00:
My software gets registered with Crowdsec as a machine (with a password), so I thought that by this way I would be able to retrieve the list of scenarios that the instance has installed, the same thing ``cscli scenarios list`` does. I'm only running one instance, but maybe as you said this don't work on multiple instances. I will do it on a different way to get the same data that ``cscli scenarios list`` returns, that's what I wanted to achieve on the beginning.

LaurenceJJones:
I guess you mean [this](https://github.com/TheDuffman85/crowdsec-web-ui)? you can look into it, it was more of initial thoughts on it, doesnt mean we will accept it as an initial look through I see very little value in exposing this data across the LAPI but open to the idea if there a way to make data actionable (compare hub latest to current scenario we may need to do a hub upgrade etc etc).

As if it is the project above is the software, it seems your just trying to replicate the data that we show in our console (not a bad thing but not many users actual go to the engine information to see this data as users primarily care about alerts/decisions)

JGeek00:
Yeah I'm using that app (I'm not the developer). I think it's a good app to monitor Crowdsec instances that are set up on home servers (my case) or small infrastructures. For enterprises and bigger infrastructures using the console makes more sense.

blotus:
Hello,

I'm going to expand a bit on what @LaurenceJJones mentioned.

We really do not want to add "random" endpoints in LAPI. Here, you add an endpoint for scenarios, why not one for parsers ? Or any other hub item ? If we were to add something like, the amount of endpoints should either be limited (ie, one endpoint for everything), or at the very least, scoped under a specific path.

About the data returned: different machines can have different scenarios, so returning just a single list is not that useful.

Also, I would consider this endpoint to be more of an "administrative" one (as a LP could get informations about other machines than itself), and as such it should not be accessed using LP credentials.
In a perfect world, we would introduce another type of permission to allow access to this type of endpoints, but it's not one of our priority.

With all of that in mind, I don't think it's something we want for now: we had internal discussions about centralized configuration management (which would have required something like this most likely), but we decided to move it in the future due to the complexity and required architectural changes.

JGeek00:
No problem @blotus, you are the people in charge of this project and you decide what can be added. I just want to tell you that if in the future you decide to expand the capabilities of LAPI it would be helpful for some users of Crowdsec. Thank you for checking this PR.

PR #4268, opened 2026-02-02
Waf challenge mode

blotus:


buixor:
Questions:
 - Currently, having appsec-configs that set `SetLegitimateBot` and others that set `SendChallenge` will only be processed based on order. Should we have `SetLegitimateBot` take priority ?

Remarks:
 - `IsLegitimateBot` seems to be using the global logger instead of a dedicated logger, making it harder to debug. (Or maybe it should be at the `.json` file as an indicator or smth)
 

Issues (not directly related to this PR):

 - `crowdsecurity/appsec-challenge-too-many-requests` : will trigger as a false-positive when the same IP validates the challenge too often and/or request a few without submitting. We need to have a cancel_on approach or similar.
- Currently, when trying to load appsec-config that create a startup error, it doesn't show which one is faulty:

```
time="2026-06-22T14:05:07Z" level=fatal msg="crowdsec init: while loading acquisition config: /etc/crowdsec/acquis.d/appsec.yaml: datasource of type appsec: unable to build appsec_config: inband: unable to build post_eval hook : unable to compile apply SetLegitimateBot() : unknown name SetLegitimateBot (1:1)\n | SetLegitimateBot()\n | ^"
```

buixor (approved):
Let's fucking go

Issue #2810, opened 2024-02-02
[crowdsec] TLS agent limitation on cert expiry


LaurenceJJones:
### What would you like to be added?

When using TLS authentication for "agents" there is a limitation that has been found with the way that we load the certificates, since we load the certs at startup time if the cert expires whilst the "agent" is running it will hit an infinite authentication failure "wall"

We need the "agent" to be smarter to self heal itself for example within k8's cert manager will automatically renew the certificate over the existing cert, however, the "agent" will try infinitely to re authenticate without reloading the certificates

Current workaround is killing the existing pods so that startup happens again, however, this is not an ideal solution when using short life certs

/kind enhancement

### Why is this needed?

"agents" can be smarter about how to deal with a 401 response from LAPI when using certificate authentication

especially-relative:
Just ran into this

As a workaround we have given the certs longer lifespans than the expected time between node updates, not ideal but better than alternatives

PR #4269, opened 2026-02-02
enhance: Reload cert if 401 response from LAPI

LaurenceJJones:
fixes: #2810

Quick fix for now till a better solution could be achieved, if the response code from LAPI is 401 and we have certs configured simply re-read them even if they have expired and havent been updated on disk, the retry method will continue to backoff and fail if ultimately if the setup is errored.

### DIFF
--- a/pkg/apiclient/client.go
+++ b/pkg/apiclient/client.go
@@ -24,7 +24,37 @@ var (
 	Cert               *tls.Certificate
 	CaCertPool         *x509.CertPool
 	lapiClient         *ApiClient
+
+	// CertPath and KeyPath store the paths to the client certificate and key files.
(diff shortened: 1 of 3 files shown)


pkg/apiclient/client.go
```diff
@@ -24,8 +24,38 @@ var (
 	Cert               *tls.Certificate
 	CaCertPool         *x509.CertPool
 	lapiClient         *ApiClient
+
+	// CertPath and KeyPath store the paths to the client certificate and key files.
```
LaurenceJJones: @mmetc will kill me for more global vars, but atm they are held globally and scope of PR is a quick fix rather than a re implementation.

PR #4294, opened 2026-02-09
pkg/csplugin: prevent race condition, deadlock

mmetc:


mmetc:
this should reduce (if not remove) the flakiness in https://github.com/crowdsecurity/crowdsec/actions/runs/21713230734/job/62644176583

PR #4295, opened 2026-02-09
feat(notification-http): add dynamic URL query params support

TO19:
Hello,

In the http notification plugin I needed query params to be dynamic for an API that only uses query parameters.

So I implemented it this way :
If the rendered format output is a JSON object containing a url_query_params field, those key-value pairs are URL-encoded and appended to the configured URL. An optional request_body field controls the HTTP body; if omitted, the body is sent empty.

If the format output is not JSON or lacks url_query_params, existing behavior is unchanged

A cleaner implementation could use a dedicated config key like query_params, but since templating happens in the broker, that would require deeper changes.

How to test :

- Run bats tests
- Build the notification-http binary and replace it on a running crowdsec setup

Example of a format in the configuration notifications/http.yaml
```yaml
format: |
  {"url_query_params": {"message": "{{range $i, $v := .}}{{if $i}},{{end}}{{$v.Source.Value}} banned for {{$v.Scenario}}{{end}}"}}
  ```
  It will produce `https://www.example.com/?message=19.19.19.19+banned+for+scenario`
  
  Tested on `version: v1.7.6-debian-pragmatic-amd64-eacc8192`

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -57,7 +57,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@9e907b5e64f6b83e7804b09294d44122997950d6 # v4.32.3
+      uses: github/codeql-action/init@89a39a4e59826350b863aa6b6252a07ad50cf83e # v4.32.4
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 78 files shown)


blotus:
Hello,

Sorry for not looking at this sooner.

While the code looks fine, I really don't like this approach: we are mixing data and configuration.

What's your actual use case ? A webhook that only support fetching data from the query string ?

Also, note that notification plugins are standalone: you can easily build your own that includes whatever functionality you want. You can find the documentation [here](https://docs.crowdsec.net/docs/next/local_api/notification_plugins/writing_your_own_plugin).

TO19:
Hello,

Totally agree it's mixing stuff but adding a clean key would require too much change.

Yes exactly I need to talk to an api that only accept query params.

I end up with an automated script to build my own.

PR #4355, opened 2026-03-05
WAF: enforce body size limitation

blotus:
Allow to enforce body size limitation in the appsec datasource itself:
 - Action can be drop (drops the request) / partial (only process up to X bytes from the body) / allow (do not process the body at all)
 
 This configuration is different from the one in coraza. Because we read the body ourselves before passing it to coraza, we need to be able to properly limit the amount of body buffered in memory.
 
For now, this only allows setting a global value used for all requests as handling this properly in `pre_eval` would require a lot of refactoring/changes.

A `DisableBodyInspection()` helper is also added to dynamically disable reading the body from a `pre_eval` hook.


buixor:
update PR desc and content

sabban (commented):
I don't see anything wrong here. I just don't like the forwardedheaders global. But I don't feel knowledgeable enough to be trustworthy on this PR.

buixor (approved):

Issue #3373, opened 2024-12-19
LAPI stram mode doesn't return active decision when filtering origin "crowdsec"


AR2000AR:
### What happened?

LAPI stram mode doesn't return active decision when filtering origin "crowdsec" but will when using querry mode

### What did you expect to happen?

The active decision should have been returned.

### How can we reproduce it (as minimally and precisely as possible)?

- Have one active crowdsec ban decision
- `curl --header "X-Api-Key: <REDACTED>" "http://127.0.0.1:8080/v1/decisions/stream?origins=crowdsec&scenarios_containing=ssh&scopes=Ip&startup=true" -s` **Fail**
<details>

```json
{"deleted":null,"new":null}
```
</details>

- `curl --header "X-Api-Key: <REDACTED>" "http://127.0.0.1:8080/v1/decisions?origins=crowdsec&scenarios_containing=ssh&scopes=Ip&startup=true" -s` **Normal**

<details>

```json
[
    {"duration":"4m1s","id":17209360,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"12m56s","id":17209361,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"22m43s","id":17209362,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"31m38s","id":17209363,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"41m25s","id":17230684,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"50m20s","id":17230685,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
    {"duration":"59m14s","id":17230686,"origin":"crowdsec","scenario":"crowdsecurity/ssh-slow-bf","scope":"Ip","type":"ban","value":"103.216.116.126"},
[...]

PR #4368, opened 2026-03-14
Add dedup Param to apiClient

Teifun2:
With #3373 it was discussed that the dedup option can be used to prevent deduplication.
This option was added to the server with #687

Exposing this would help other client use this query param properly.

### DIFF
--- a/pkg/apiclient/decisions_service.go
+++ b/pkg/apiclient/decisions_service.go
@@ -33,6 +33,7 @@ type DecisionsStreamOpts struct {
 	Startup                bool   `url:"startup,omitempty"`
 	CommunityPull          bool   `url:"community_pull"`
 	AdditionalPull         bool   `url:"additional_pull"`
+	Dedup                  bool   `url:"dedup"`
 	Scopes                 string `url:"scopes,omitempty"`
 	ScenariosContaining    string `url:"scenarios_containing,omitempty"`
 	ScenariosNotContaining string `url:"scenarios_not_containing,omitempty"`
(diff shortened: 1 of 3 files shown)


blotus:
Hello,

I'm not sure I understand the goal of the PR ?
There's already a `dedup` parameter present in LAPI.

Here, you have just added it to the CAPI swagger and to the queries made to CAPI, but this won't have any impact: CAPI does not support deduplicating decisions, and never will: CAPI only serves blocklists content, which will always have unique IPs.

lluisd:
> Hello,
> 
> I'm not sure I understand the goal of the PR ? There's already a `dedup` parameter present in LAPI.
> 
> Here, you have just added it to the CAPI swagger and to the queries made to CAPI, but this won't have any impact: CAPI does not support deduplicating decisions, and never will: CAPI only serves blocklists content, which will always have unique IPs.

To give context this comes from my unifi bouncer issue (https://github.com/Teifun2/cs-unifi-bouncer/issues/77) where I explain it. 

The problem is that I use 2 crowdsec bouncers. The first one is the firewall of my Unifi router for crowdsec/cli origins only because if I use all then my unifi router crashes. Then I use the CAPI/lists origins on the traefik bouncer as the second layer. 

But I noticed I was losing some crowdsec decisions on my Unifi bouncer and I realized this was because they were already in CAPI or lists. Then I realize that using the dedup=false query param like that ```/v1/decisions/stream?additional_pull=false&community_pull=false&origins=crowdsec%2Ccscli&dedup=false``` I can get them, but this can only be used by calling the api manually, not by using the github.com/crowdsecurity/crowdsec package like its done in https://github.com/Teifun2/cs-unifi-bouncer repo.

blotus:
Ok, got it, thanks for the clarification.

If you don't mind, I'll try to make the changes myself this week, as I'd like to clean up some things as well (for example, there's no good reason for calls to LAPI to have the `additional_pull` and `community_pull` parameters set, as they are only relevant for CAPI)

lluisd:
@blotus is there any update?

PR #4435, opened 2026-04-24
db: add some missing indexes

blotus:
Somes indexes were missing.
While some of them were on tables with very little rows (machines, bouncers, ...), other can have a measurable impact (alert scenario or allowlist expiration).

Indexes added:
 - `Bouncer(api_key, auth_type)`: runs on every bouncer request. Small table but cleaner
 - `Bouncer(api_key, ip_address)`
 - `Bouncer(last_pull, created_at)`: used when flushing inactive bouncers from the DB
 - `Machine(last_heartbeat, created_at)`: same for agents
 - `Alert(uuid)`: used by PAPI when inserting a new alert from the console
 - `Alert(scenario)`: used on the `/alerts` endpoint when filtering per scenario (`cscli alerts list --scenario XX`)
 - `AllowListItem(expires_at)`: Used to flush expired allowlist items or when updating blocklists from CAPI
 - `AllowList(allowlist_id)`

### DIFF
--- a/pkg/database/ent/migrate/schema.go
+++ b/pkg/database/ent/migrate/schema.go
@@ -57,6 +57,16 @@ var (
 				Unique:  false,
 				Columns: []*schema.Column{AlertsColumns[0]},
 			},
+			{
+				Name:    "alert_uuid",
+				Unique:  false,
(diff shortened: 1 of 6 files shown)


sabban (approved):
lgtm

PR #4458, opened 2026-05-07
waf-challenge-mode : infrastructure improvements

buixor:
Keys management: epoch-bound key derivation from master
 - Key rotation, derived from "master" secret, used for signing/encrypting secrets, rotated at configuration interval.
 - Cookie sealed by secret derived from master key (TTL independent of key rotation interval)

Code obfuscation: Split obfuscation of crypto and fpscanner code
- Allow user to configure pool size (number of distinct obfuscated copies) and interval of code re-obfuscation

Configuration: Exposed via appsec-configs
- master secret
- key rotation settings (interval, max-live epochs)
- cookie ttl
- obfuscation settings (interval and pool size for both lib and crypto code)

Expr helpers: 
 - expose new phases: `on_challenge_suibmit` (invoked when user submits challenge) and `on_challenge` (invoked whenever there is a valid cookie)
 - added helper to reject (`RejectSubmission`) or give challenge bypass (`GrantChallengeCookie`)


pkg/appsec/challenge/crypto.go
```diff
@@ -19,12 +20,24 @@ var (
...
 )
 
 const hkdfInfo = "crowdsec-challenge-cookie"
 
-func deriveKey(secret string) ([]byte, error) {
-	h := hkdf.New(sha256.New, []byte(secret), nil, []byte(hkdfInfo))
+// Cookie wire format:
+//
+//   v1 (current): cookieVersionV1 || epoch_be8 || nonce || ciphertext
+//   v0 (legacy):  nonce || ciphertext
```
buixor: kill v0 compat it hasn't been released

blotus (approved):

Issue #4464, opened 2026-05-12
High CPU usage on v1.7.8


hanjo:
Hi,

I'm running crowdsec in docker on two Linux (Ubuntu) servers. After updating to v1.7.8, the CPU usage on both servers roughly doubled to tripled.

There is nothing of relevance in the logs.

After moving back to v1.7.7, CPU usage dropped to expected values.

Based on the release notes, I cannot pinpoint the issue. Any suggestions?

Thanks!

blotus:
Hello,

Could you give more informations about the servers, CPU type, ...

What was the baseline CPU usage ? Did you went from 10% to 30% or from 2% to 6% ?

How many bouncers do you have ? We made a change in 1.7.8 to improve the response time and greatly reduce the memory consumption of the decisions stream, but this comes at the cost of slightly higher CPU usage on LAPI.

Could you as well provide a CPU profile of crowdsec ?
Assuming you run the command on the machine where crowdsec is running:
 `curl http://localhost:6060/debug/pprof/profile -o profile.gz`.

The command will take 30s to run.

hanjo:
Thanks for the quick response.
Server A (Intel Core i5-9500T CPU @ 2.20GHz, 6 cores) went from 18-19% to 38-42%, memory dropped from 91-94MB to 55-57MB.
Server B (some VPS, using QEMU Virtual CPU version 2.5+, 4 cores) went from 10-11% to 40-47%, memory dropped from 130-450MB to 105-390MB.

On Server A I have 12 bouncers, but only 5 are active (there are duplicates due to multiple source IP addresses).
On Server B I have 4 bouncers (all active).

I have attached the CPU profiles, but I created them on v1.7.7 now, since I rolled back. Do you need them from v1.7.8, or doesn't it matter?

[profile-serverA.gz](https://github.com/user-attachments/files/27625752/profile-serverA.gz)
[profile-serverB.gz](https://github.com/user-attachments/files/27625751/profile-serverB.gz)

tibouh:
Same issue on my side.
Moreover [crowdsec_web_ui](https://github.com/TheDuffman85/crowdsec-web-ui) cannot connect saying CrowdSec LAPI is offline.
Reverted to v1.7.7 and it's working again.

blotus:
[...]

PR #4468, opened 2026-05-16
Add composite indexes for alert decision filters

jimstrang:
## Overview

This PR addresses the `/v1/alerts` slowdowns and `database is locked` reports in [#4464](https://github.com/crowdsecurity/crowdsec/issues/4464) by adding a few composite indexes on the `decisions` table. The reports started showing up after the 1.7.8 decision-stream chunking changes, but I do not think the chunked stream is the slow query itself. My read is that it made the SQLite/LAPI workload more likely to overlap with already-expensive alert-list queries, which is where the lock symptoms surface.

The slow path I was able to reproduce directly is an alert-list query that combines decision filters such as `decision_type`, `origin`, and `has_active_decision`. Today those filters are translated into separate `EXISTS` checks against `decisions`. With only the existing `alert_decisions` index, SQLite can find the decisions attached to an alert, but it still has to walk all of them to check `type`, `origin`, or `until`.

That gets painful when an alert has a large number of linked decisions. In the database I used for validation, one alert was linked to roughly 15,000 decisions (CAPI community blocklist), and the original query shape could take tens of seconds.

## What Changed

This adds composite indexes for the columns used by the alert decision filters:

- `(alert_decisions, type)`
- `(alert_decisions, origin)`
- `(alert_decisions, until)`

The schema change is in `pkg/database/ent/schema/decision.go`, with the generated migration schema updated in `pkg/database/ent/migrate/schema.go`. Existing installations should get the indexes through the normal Ent auto-migration path on startup.

## Why This Approach

There is another possible fix that keeps everything in code: combine the decision-related alert filters into one `HasDecisionsWith(...)` call so Ent emits a single `EXISTS` subquery instead of several. That also performs well.

I did not take that route here because it can change the meaning of `/v1/alerts` filters.

The current query matches alerts that have at least one linked decision for each filter, even if those filters match different decision rows. The combined-predicate version would require the same decision row to satisfy all filters at once. That may be the intended behavior, but it is a semantic decision rather than just a performance fix, and I did not see existing tests or documentation that clearly pin it either way.

If maintainers can clarify the intended filter semantics here, I think that code-only predicate change would still be worth exploring. In other words, should multiple decision filters on `/v1/alerts` mean "find alerts where one linked decision matches all of these conditions" or "find alerts where the linked decisions collectively satisfy these conditions"?

The index-only change keeps the current behavior and just gives the database a better access path.

## Validation

I tested this against a SQLite database from an affected setup with:

- 107 alerts
- 30,101 decisions
- one alert linked to about 15,000 decisions

For the original `/v1/alerts` decision-filter query shape:

| Case | Runtime |
| --- | ---: |
| Before these indexes | ~39s |
| After these indexes | <1ms |

The row count stayed the same before and after the indexes, so this is a performance-only change.

After the change, SQLite uses covering indexes for the correlated decision lookups, for example:

```text
SEARCH decisions EXISTS USING COVERING INDEX decision_alert_decisions_until  (alert_decisions=? AND until>?)
SEARCH decisions EXISTS USING COVERING INDEX decision_alert_decisions_type   (alert_decisions=? AND type=?)
SEARCH decisions EXISTS USING COVERING INDEX decision_alert_decisions_origin (alert_decisions=? AND origin=?)
```

On the same snapshot, creating the three indexes took about 125ms total for ~30k decisions. Larger databases will take longer, but this should be a one-time migration cost.

Refs [#4464](https://github.com/crowdsecurity/crowdsec/issues/4464)

### DIFF
--- a/pkg/database/ent/migrate/schema.go
+++ b/pkg/database/ent/migrate/schema.go
@@ -249,6 +249,21 @@ var (
 				Unique:  false,
 				Columns: []*schema.Column{DecisionsColumns[16]},
 			},
+			{
+				Name:    "decision_alert_decisions_type",
+				Unique:  false,
+				Columns: []*schema.Column{DecisionsColumns[16], DecisionsColumns[5]},
+			},
+			{
+				Name:    "decision_alert_decisions_origin",
+				Unique:  false,
+				Columns: []*schema.Column{DecisionsColumns[16], DecisionsColumns[13]},
+			},
+			{
+				Name:    "decision_alert_decisions_until",
+				Unique:  false,
+				Columns: []*schema.Column{DecisionsColumns[16], DecisionsColumns[3]},
+			},
 		},
 	}
 	// EventsColumns holds the columns for the "events" table.
--- a/pkg/database/ent/schema/decision.go
+++ b/pkg/database/ent/schema/decision.go
@@ -57,5 +57,8 @@ func (Decision) Indexes() []ent.Index {
 		index.Fields("value"),
 		index.Fields("until"),
 		index.Fields("alert_decisions"),
+		index.Fields("alert_decisions", "type"),
+		index.Fields("alert_decisions", "origin"),
+		index.Fields("alert_decisions", "until"),
 	}
 }


jimstrang:
Small update: I’m moving this to draft for now.

The indexes still look useful, and they do improve the access path for `/v1/alerts` decision filters in isolation. But after more testing against a copy of an affected SQLite DB, I can still reproduce the Homepage-style bans query timing out under concurrent LAPI activity on current `master`, including with WAL enabled.

So I don’t want to present this as the confirmed fix for #4464 yet. It may still be part of the fix, or a worthwhile performance improvement on its own, but I’d like to verify the concurrent case before asking for a full review.

PR #4473, opened 2026-05-18
downgrade github.com/mattn/go-sqlite3 to v1.14.32

blotus:
Should fix the underlying regression for #4470 and #4464 (in addition to #4471 which does seem to help).

After investigation, the regression came from the update of github.com/mattn/go-sqlite3 from `v1.14.23` to `1.14.44` (in this PR #4453 for the final update).

After testing the various releases of the go library, it seems the regression was introduced in `v1.14.33`.

I'm not sure *what* the regression is exactly (claude points to [this commit](https://sqlite.org/src/info/e33da6d5dc), but I didn't looked at it in details).

So for now, let's just downgrade sqlite to get back to the performance we had before the release, and we will also improve the queries themselves (with at least #4471, and I've also identified potential gains in the way we perform the pagination for `/v1/alerts`, but that will come later)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	github.com/jedib0t/go-pretty/v6 v6.7.10
 	github.com/jszwec/csvutil v1.10.0
 	github.com/mattn/go-isatty v0.0.22
-	github.com/mattn/go-sqlite3 v1.14.44
+	github.com/mattn/go-sqlite3 v1.14.32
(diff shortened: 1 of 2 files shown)


jimstrang:
Cool, this matches what I found in https://github.com/crowdsecurity/crowdsec/pull/4471#issuecomment-4479300299 - will run some benchmarks on my database and report back in the #4464 issue thread.

buixor (approved):

PR #4475, opened 2026-05-19
allowlist: add disable_allowlist_ingestion feature flag

### LINKED ISSUE #4115: Allowlist feature cannot withstand ddos load
### What happened?

I have one central API and hundreds of agents that are connected to one API.
The API is deployed on bare metal.

In v1.7.3, I noticed that the API stopped working when there was a huge wave of DDOS attacks.
On the API side, I saw in the log:
```
{"level":"error","msg":"error while checking allowlist: unable to check if value is allowlisted: context canceled","time":"2025-11-21T18:24:20+03:00"}
{"level":"error","msg":"error while checking allowlist: unable to check if value is allowlisted: context canceled","time":"2025-11-21T18:24:20+03:00"}
```

The API has stopped working.
I'm not using any allowlist functions, but crowdsec is still trying to connect to the database and request the allowlist.

After downgrading to version 1.6.5, everything is working and it keeps the load on

If you need any additional information, let me know.

Image
Image
Image

Allowlists should not be checked if they are not in use.

You can reproduce it using >= 1.6.6

_No response_

### Crowdsec version

problem starting from [v1.6.6](https://github.com/crowdsecurity/crowdsec/releases/tag/v1.6.6)

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
# paste output here
```

</details>

<details>

```console
$ cscli hub list -o raw
# paste output here
```

</details>

### Acquisition config

<details>
```console
# On Linux:
# paste output here

# On Windows:
# paste output here
</details>


### Config show

<details>

```console
$ cscli config show
# paste output here
```

</details>

### Prometheus metrics

<details>

```console
$ cscli metrics
# paste output here
```

</details>

<details>

</details>

aa1ex:
Adds a `disable_allowlist_ingestion` feature flag that skips the allowlist lookup during alert ingestion. 

Requested in #4115: on high-load setups that don't rely on the allowlist, the per-alert allowlist DB lookup is pure overhead. The  flag lets such operators opt out of it. 

Closes #4115

### DIFF
--- a/pkg/apiserver/alerts_test.go
+++ b/pkg/apiserver/alerts_test.go
@@ -13,12 +13,13 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/go-openapi/strfmt"
+	logtest "github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	logtest "github.com/sirupsen/logrus/hooks/test"
 
 	"github.com/crowdsecurity/crowdsec/pkg/csconfig"
 	"github.com/crowdsecurity/crowdsec/pkg/database"
+	"github.com/crowdsecurity/crowdsec/pkg/fflag"
 	"github.com/crowdsecurity/crowdsec/pkg/models"
 )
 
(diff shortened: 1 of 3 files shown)


blotus:
Hello,

Thanks for the PR, but I don't think it's the right approach.

As I mentioned in the issue you linked, I think the allowlists are a red herring: the function call will take no time at all ( microseconds according to my quick local testing, and it's definitely negligible compared to the time needed to actually write the alerts to the DB).

If you can provide a profile showing that the call is indeed too expensive, I'd  be happy to revisit this, but I don't want to add code/features purely based on hunches.

aa1ex:
Hello @blotus,

Your microbench result made me want to test this end-to-end. Built a single-machine docker stand (real MySQL, paired A/B with just the feature flag flipped between arms) and ran N=20 paired-bootstrap measurements on two hosts:

https://github.com/aa1ex/crowdsec-issue-4115

A = allowlist on, B = patched off. Numbers (paired delta A-B, 95% CI):

- **+1 SELECT per alert at the DB level**, the same in every pair on both hosts (`Com_select` from `SHOW GLOBAL STATUS`, not our instrumentation). One extra round-trip per alert, doesn't depend on hardware.
- **Per-call at p50: ~234us on Apple M3 Pro, ~1.15ms on Intel i7-8700.** Per request (batch of 10): +2.34 ms [+2.24, +2.45] Mac, +11.53 ms [+11.43, +11.63] Linux. CIs exclude zero on every measured percentile (p50/p95/p99) on both hosts.
- That's ~18% (Mac) / ~27% (Linux) of the variant-A median per-request latency, so not negligible vs the rest of the per-request work.

You're right that the function-level CPU cost is microseconds. What the patch removes is the SQL round-trip (ent -> driver -> MySQL -> response), which on an empty `allow_list_items` table is still hundreds of microseconds to over a millisecond depending on hardware/transport.

Caveats: this isn't a reproduction of the unresponsive-LAPI cascade in #4115 (only measures per-call overhead); pool-wait isn't the growth mechanism on this stand (pool isn't saturated at the tested rate); absolute ms don't generalize (5x spread between the two hosts).

Patch in the PR is exactly what's measured (`patches/disable-allowlist-ingestion.patch`, ~100 lines, 3 files, applies cleanly to v1.7.8). One command runs everything (`./scripts/run.sh`), ~25 min for smoke, ~3 h for N=20. HTML reports per host with profile diffs are linked from RESULTS.md. Let me know if you want a different scenario covered.

PR #4481, opened 2026-05-20
buckets: make a copy of all event maps before pouring the event to a bucket

### LINKED ISSUE #4459: Data race in pkg/leakybucket.EventsFromQueue causes 'concurrent map iteration/read and map write' panic (v1.7.7)
### What happened?

CrowdSec v1.7.7 (docker) crashes repeatedly with **Go runtime fatal data race** in `pkg/leakybucket.EventsFromQueue`, called from `NewAlert` during bucket overflow. Two distinct race variants observed within ~35 min on the same instance:

**Crash 1 — `concurrent map iteration and map write`:**
```
fatal error: concurrent map iteration and map write

goroutine 420788 [running]:
internal/runtime/maps.(*Iter).Next(...)
      internal/runtime/maps/table.go:792 +0x86
github.com/crowdsecurity/crowdsec/pkg/leakybucket.EventsFromQueue(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/overflows.go:215 +0x129
github.com/crowdsecurity/crowdsec/pkg/leakybucket.NewAlert(0xc003ff91e0, 0xc0053258e0)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/overflows.go:350 +0x95c
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*Leaky).overflow(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/bucket.go:294 +0x68
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*Leaky).LeakRoutine(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/bucket.go:209 +0xcdd
github.com/crowdsecurity/crowdsec/pkg/leakybucket.LoadOrStoreBucketFromHolder.func1()
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/manager_run.go:188 +0x97
created by github.com/crowdsecurity/crowdsec/pkg/leakybucket.LoadOrStoreBucketFromHolder in goroutine 989
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/manager_run.go:184 +0x35d
```

**Crash 2 — `concurrent map read and map write` (35 min later, same instance):**
```
fatal error: concurrent map read and map write

goroutine 695697 [running]:
github.com/crowdsecurity/crowdsec/pkg/leakybucket.EventsFromQueue(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/overflows.go:222 +0x725
github.com/crowdsecurity/crowdsec/pkg/leakybucket.NewAlert(0xc0011651e0, 0xc005cc22e0)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/overflows.go:350 +0x95c
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*Leaky).overflow(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/bucket.go:294 +0x68
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*Leaky).LeakRoutine(...)
      github.com/crowdsecurity/crowdsec/pkg/leakybucket/bucket.go:209 +0xcdd
```

Both races point at `EventsFromQueue` iterating/reading `qEvents[idx].Meta` (an `Event` `Meta` map) without synchronization. Looking at master `pkg/leakybucket/overflows.go`:

```go
for idx := range qEvents {
    if qEvents[idx].Meta == nil { continue }
    meta := models.Meta{}
    skeys := make([]string, 0, len(qEvents[idx].Meta))
    for k := range qEvents[idx].Meta {        // <-- iteration race (line 215 in v1.7.7)
        skeys = append(skeys, k)
    }
    sort.Strings(skeys)
    for _, k := range skeys {
        v := qEvents[idx].Meta[k]              // <-- read race (line 222 in v1.7.7)
        ...
    }
}
```

The `Event` queued in the bucket appears to still be referenced/mutated by other goroutines (e.g. enrichment/parse paths) while the alert builder iterates its `Meta`.

No fatal panic; alert generation should be safe against concurrent access of `Event.Meta`, either by deep-copying the map before iteration or guarding writes with a mutex.

### How can we reproduce it?

- **Image:** `crowdsecurity/crowdsec:latest` (digest `sha256:6ca53ad26196ca59ddd4fa692a586b73d8fcde085046163b9ca2f04887dca563`)
- **Version:** `v1.7.7-981e6166`, Go 1.25.5, docker build 2026-04-01
- **Host:** Ubuntu 24.04, Linux 6.17, x86_64, Docker Engine
- **Stack:** Traefik (reverse proxy with `crowdsec-bouncer-traefik-plugin`) → CrowdSec LAPI + AppSec
- **Collections:** `crowdsecurity/traefik`, `crowdsecurity/appsec-virtual-patching`, `crowdsecurity/appsec-generic-rules`
- **Acquisition:** Traefik access log file + AppSec endpoint
- **Traffic profile during incident:**
  - ~6 bouncer `GET /v1/decisions?ip=` queries per second (32 685 queries over 90 min)
  - 160 730 lines read from Traefik log, 94 990 poured to buckets in ~90 min
  - 402 AppSec events
  - Active scanner traffic generating multiple overflow scenarios (`crowdsecurity/http-admin-interface-probing`, `crowdsecurity/http-sensitive-files`) every few seconds

Steps:
1. Run `crowdsecurity/crowdsec:v1.7.7` with the Traefik + AppSec collections.
2. Drive sustained mixed traffic — legitimate + scanner — sufficient to overflow several scenarios per minute.
3. Container crashes within ~30–45 minutes with one of the two stack traces above. Without a restart policy, CrowdSec stays down. With `restart: always`, the crash is masked but `RestartCount` accumulates.

### Why this differs from #4216

Issue #4216 reports **nil-pointer** panics in the same area and was tentatively attributed to "high memory pressure". This report is different and rules that explanation out:

- Crash is **not** nil-pointer — Go runtime explicitly says `concurrent map iteration and map write` / `concurrent map read and map write`. The runtime only emits these messages when it detects parallel access on the same map.
- Memory headroom is large: container RSS **127 MiB / 15.42 GiB host RAM (0.8 %)**, no cgroup limit, no swap pressure, no OOM-kill (`OOMKilled: false`, `ExitCode: 0`).
- The race is reproducible across two independent goroutines on the same instance within 35 minutes.

- Crash is silently masked by `restart: always` — many users likely hit this without noticing because the bouncer caches decisions across the short restart gap.
- A grep of master `pkg/leakybucket/overflows.go` shows the same unsynchronized iteration; recent commits to this file (#4250, #4284, #4285, #4290) are refactors and do not address synchronization. No open PR appears to fix this race.
- Happy to provide full goroutine dumps, full container logs, or run a debug build if useful.

### Crowdsec version
```
version: v1.7.7-981e6166
Codename: alphaga
BuildDate: 2026-04-01_13:14:30
GoVersion: 1.25.5
Platform: docker
libre2: C++
```

### OS version
Ubuntu Server 24.04, Linux 6.17.0-1012-aws, x86_64
Docker Engine (rootful)

- `crowdsecurity/traefik`
- `crowdsecurity/appsec-virtual-patching`
- `crowdsecurity/appsec-generic-rules`
- `crowdsecurity/whitelists`
- a few local custom postoverflows / parsers / appsec rules (whitelists only)

### Acquisition config
- Traefik JSON access log via `file` source
- AppSec endpoint

blotus:
Fixes #4459.

Each bucket now gets its own copy of the two maps a scenario can write to. `Enriched` and `Unmarshaled` are not cloned: `Unmarshaled` is intended to be used only in parsers, and the various enrichment method cannot be called from a scenario.

Cost, per (event, bucket) pour — realistic Caddy JSON event, M4 Pro:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before (maps shared) | 11 | 0 | 0 |
| after | 185 | 1000 | 6 |

End to end, a pour into 48 loaded scenarios with 3 matching goes from 418 to 439
allocs/op; the wall-clock difference is inside the noise of that benchmark.

As a side-effect, this also fixes a potential race conditions in the meta:
 - Scenario A sets a new meta with `evt.SetMeta`
 - Scenario B overflows
 - The meta set by scenario A would appear in the meta of the alert generated by scenario B because the maps were shared between all buckets.

### DIFF
--- a/pkg/leakybucket/bench_race_test.go
+++ b/pkg/leakybucket/bench_race_test.go
@@ -1,0 +1,152 @@
+package leakybucket
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/crowdsecurity/crowdsec/pkg/pipeline"
+)
+
+// benchEvent returns a fresh, realistic parsed HTTP event. Meta is a single map
+// that PourItemToHolders will share (by reference) across every matching bucket.
+func benchEvent(now time.Time) pipeline.Event {
+	return pipeline.Event{
+		Time: now,
+		Meta: map[string]string{
+			"source_ip":       "1.2.3.4",
+			"service":         "http",
+			"log_type":        "http_access-log",
+			"http_status":     "404",
+			"http_path":       "/wp-admin/",
+			"http_verb":       "GET",
+			"http_user_agent": "curl/8.0",
+			"http_host":       "example.com",
+			"machine":         "test",
+			"datasource_path": "/var/log/traefik/access.log",
+			"datasource_type": "file",
+			"k1":              "v1",
+			"k2":              "v2",
+			"k3":              "v3",
+		},
+	}
+}
+
+// runPourBenchmark drives PourItemToHolders against `readers` trigger scenarios
+// (each overflows immediately and reads evt.Meta while building its alert). If
+// withWriter is true, a final scenario whose filter mutates the shared event via
+// evt.SetMeta(...) is appended — reproducing the crowdsecurity/http-technology-probing
+// side effect and the crowdsecurity/crowdsec#4459 data race.
+func runPourBenchmark(b *testing.B, readers int, withWriter bool) {
+	b.Helper()
+
+	// keep the overflow alert validation noise out of the benchmark output
+	oldLevel := log.GetLevel()
+	log.SetLevel(log.PanicLevel)
+	defer log.SetLevel(oldLevel)
+
+	// response collects overflows emitted by every bucket (AllOut). It must be
+	// drained continuously, otherwise the overflowing trigger goroutines block
+	// on the send *after* having read the shared map.
+	response := make(chan pipeline.Event, 4096)
+
+	stop := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
+			case <-response:
+			}
+		}
+	}()
+	defer close(stop)
+
+	holders := make([]BucketFactory, 0, readers+1)
+
+	// Readers: each is a trigger that overflows immediately and reads evt.Meta
+	// while building the alert.
+	for i := range readers {
+		holders = append(holders, BucketFactory{
+			Spec: BucketSpec{
+				Name:        fmt.Sprintf("reader_%d", i),
+				Description: "reads shared Meta on overflow",
+				Type:        "trigger",
+				Filter:      "true",
+			},
+		})
+	}
+
+	if withWriter {
+		// The writer last: its filter has a side effect on the shared event, just
+		// like crowdsecurity/http-technology-probing's `evt.SetMeta(...)`.
+		holders = append(holders, BucketFactory{
+			Spec: BucketSpec{
+				Name:        "writer",
+				Description: "mutates shared Meta from its filter",
+				Type:        "trigger",
+				Filter:      `evt.SetMeta("injected", "1")`,
+			},
+		})
+	}
+
+	for idx := range holders {
+		holders[idx].ret = response
+		if err := holders[idx].LoadBucket(); err != nil {
+			b.Fatalf("while loading holder %d: %s", idx, err)
+		}
+
+		if err := holders[idx].Validate(); err != nil {
+			b.Fatalf("while validating holder %d: %s", idx, err)
+		}
+	}
+
+	ctx := b.Context()
+	bucketStore := NewBucketStore()
+	now := time.Now().UTC()
+
+	for b.Loop() {
+		evt := benchEvent(now)
+		if _, err := PourItemToHolders(ctx, evt, holders, bucketStore, nil); err != nil {
+			b.Fatalf("while pouring item: %s", err)
+		}
+	}
+}
+
+// BenchmarkPourSharedMetaRace reproduces the data race reported in
+// crowdsecurity/crowdsec#4459.
+//
+// Root cause: PourItemToHolders pours a single parsed event into every matching
+// scenario, but Event.Meta is a map (reference type) and the pour path only
+// makes a shallow struct copy (pipeline.Queue.Add). Every bucket queue across
+// all matching scenarios therefore ends up sharing the *same* Meta map.
+//
+// It wires up several "reader" trigger scenarios followed by a "writer" whose
+// filter mutates the shared map via evt.SetMeta(...). When runPour evaluates the
+// writer's filter (a map write) while an already-poured reader bucket iterates
+// the same map during overflow (a map read), the Go runtime aborts with:
+//
+//	fatal error: concurrent map iteration and map write
+//	fatal error: concurrent map read and map write
+//
+// Run as a race reproducer (reports a data race before the fix, clean after):
+//
+//	go test -tags 'netgo,osusergo,expr_debug,nomsgpack' -race \
+//	    -run='^$' -bench=BenchmarkPourSharedMetaRace ./pkg/leakybucket/
+func BenchmarkPourSharedMetaRace(b *testing.B) {
+	runPourBenchmark(b, 8, true)
+}
+
+// BenchmarkPourNoSideEffect is the same pour workload without the side-effecting
+// filter, so no map is ever mutated after being poured. It is race-free both
+// before and after the #4459 fix, which makes it the apples-to-apples way to
+// measure the fix's only cost: the per-pour CopyForBucket clone. Compare ns/op
+// and allocs/op with and without the fix applied:
+//
+//	go test -tags 'netgo,osusergo,expr_debug,nomsgpack' \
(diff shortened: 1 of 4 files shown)


YakGravity:
Ran into this on v1.8.1 too, so it's still in the current release. Same spot as the report from 20 Aug: the panic is on pour, inside the uniq processor, not on overflow.

```
fatal error: concurrent map read and map write

goroutine 34359 [running]:
reflect.mapaccess_faststr(...)
	runtime/map.go:236 +0x18
reflect.Value.MapIndex(...)
	reflect/map.go:154 +0xbb
github.com/expr-lang/expr/vm/runtime.Fetch(...)
	github.com/expr-lang/expr@v1.17.8/vm/runtime/runtime.go:63 +0x245
github.com/expr-lang/expr/vm.(*VM).Run(...)
	github.com/expr-lang/expr@v1.17.8/vm/vm.go:142 +0xd47
github.com/crowdsecurity/crowdsec/pkg/leakybucket.getElement(...)
	github.com/crowdsecurity/crowdsec/pkg/leakybucket/uniq.go:82 +0xfb
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*UniqProcessor).OnBucketPour(...)
	github.com/crowdsecurity/crowdsec/pkg/leakybucket/uniq.go:30 +0x11b
github.com/crowdsecurity/crowdsec/pkg/leakybucket.(*Leaky).LeakRoutine(...)
	github.com/crowdsecurity/crowdsec/pkg/leakybucket/bucket.go:166 +0x85d
```

Version was v1.8.1-909b5157 from the official docker image, build date 3 Sep. It happened during a scanner burst, a single source filling two http scenarios within a couple of seconds, and the crash came while the overflow was being processed. Nothing exotic in the setup.

The process did not come back on its own. The container runs with `restart: unless-stopped`, but docker made no attempt: RestartCount stayed at 0 and there is no restart line in the daemon log. So the LAPI stayed down, and with the traefik bouncer plugin in live mode that fails closed, meaning 403 on every route behind the proxy rather than only the one the scanner was hitting. That lasted a bit over 13 hours until it was noticed. Moving the plugin to stream mode keeps a crash from taking everything down, but new decisions still stop until crowdsec is back.

Adding the scope here in case it is useful for prioritising.

nicheosala:
Tested this on v1.8.1 (aarch64, Arch Linux ARM) — the race is still present there, and this patch fixes it.

I hit the same crash in production: agent died after ~11h30 with `fatal error: concurrent map read and map write`, this time with the read coming from `alertcontext.EvalAlertContextRules` via `NewAlert` → `EventToContext` (`overflows.go:371`) rather than `EventsFromQueue`.

**Reproducer:** 87 Caddy JSON access-log lines from a single WordPress scanner (`POST /wp-json/batch/v1` and encoding variants, all 403/404/405/308, ~8s), replayed one-shot with a `-race` build. Stock hub config (`crowdsecurity/caddy`, `base-http-scenarios`, `http-cve`, `http_base` context), default routine counts, no AppSec, no postoverflows. Each run takes ~4s. Log attached.

**A/B, alternating both binaries in the same loop, 20 iterations each:**

| build | runs with a race report | overflows per run |
|---|---|---|
| v1.8.1 | 12/20 | 3 |
| v1.8.1 + 761e851 | 0/20 | 3 |

Only variable is the cherry-pick of 761e851 onto the v1.8.1 tag (I did not include 60fe0aa, which only adds benchmarks). The constant `overflow=3` on both sides confirms the copy happens after the filter evaluation, so `evt.SetMeta` in a filter still lands in the bucket that set it — `crowdsecurity/http-cve-probing` keeps triggering normally.

**On the writer side**, in case it helps close the analysis in #4459: it is always a scenario filter calling `evt.SetMeta()` during filter evaluation in `PourItemToHolders` (`manager_run.go:225`). Two stock hub scenarios do this — `crowdsecurity/http-cve-probing` and `crowdsecurity/http-technology-probing`. Across 20 pre-fix runs the detector reported 20 races with the same writer and three different readers:

| reader | occurrences |
|---|---|
| `UniqProcessor.OnBucketPour` → `getElement` (`uniq.go:82`) | 11 |
| `NewAlert` → `EventsFromQueue` (`overflows.go:225`) | 8 |
| `NewAlert` → `alertFormatSource` → `eventSources` (`overflows.go:108`) | 1 |

Plus `alertcontext` from the production crash, so four distinct readers in total. Copying at pour time covers all of them, which a defensive copy inside `EventsFromQueue` would not have.

One minor note: `maps.Clone` is shallow, so for `Unmarshaled` (a `map[string]any`) the nested values stay shared between buckets — with the Caddy datasource, that's the whole decoded JSON object. Nothing writes into those nested structures today, so it isn't a live bug, just something that could bite later.

Happy to re-run the A/B on a rebased branch if that would help.
[burst.log](https://github.com/user-attachments/files/32009546/burst.log)

buixor (approved):

Issue #2605, opened 2023-11-20
Inotify queue overflow in file acquisition


blotus:
We are using [fsnotify/fsnotify](https://github.com/fsnotify/fsnotify) to watch for new files if the user specifies a glob pattern.

Unfortunately, `fsnotify` hardcodes the events for which we are notified (see [here](https://github.com/fsnotify/fsnotify/blob/cfc9c4f277ea6ec18de92444b31983b183deb4fb/backend_inotify.go#L379))

Because crowdsec always adds an inotify watch on the parent directory if the user provides a glob pattern, if this directory contains a lot of files that are written constantly to, we are at risk of overflowing the kernel event queue, even though we are not interested in the events.

`fsnotify` has a longstanding [issue](https://github.com/fsnotify/fsnotify/issues/7) about event filtering, but there does not seem to be a consensus on how it should be handled.

I see a few options available to us:
 - Allow increasing the number of goroutines to handle inotify events. This feels like a hack but shouldn´t take much time to implement and likely won´t add a lot of overhead as most events will be discarded immediately (but we will still consume CPU for "nothing")
 - Drop `fsnotify` and do our own minimalist implementation (the main complexity will probably be the cross-platform support)
 - Contribute this to `fsnotify` (the better option long term , but it will probably take some back-and-forth and some time)

buixor:
A viable option seems to be:
 - switch to https://github.com/kubernetes/utils/blob/master/inotify/inotify.go on linux, as it will allow us to filter the event types.
 - keep fsnotify for other platforms, and hope we have no windows users facing the same issue :smiley:

buixor:
Issues have been solved on fsnotify, so we need to take care of it.

PR #4484, opened 2026-05-23
fix: avoid unused fsnotify watchers in file cat mode

immanuwell:
## Summary

`mode: cat` was still creating an `fsnotify` watcher, even though oneshot reads dont use it.
On boxes that are already close to `fs.inotify.max_user_instances`, that can fail early with:

`could not create fsnotify watcher: couldn't initialize inotify: too many open files`

This keeps watcher creation in tail mode only, and closes it if tail config bails early. small fix, real footgun.

## Repro

1. Run on a host that's close to the inotify instance ceiling. Linux exposes it in `/proc/sys/fs/inotify/max_user_instances`.
2. Call the file datasource configure path with `mode: cat`.
3. Before this patch it can fail before reading anything, just because it grabs an inotify instance for no reason.
4. After this patch, cat mode does not touch inotify, so the same path keeps working.

## Checks

- `go test ./pkg/acquisition/modules/file -run 'TestConfigure|TestOneShot' -count=1`

related-ish: #2605 #4074

/kind fix

### DIFF
--- a/pkg/acquisition/modules/file/config.go
+++ b/pkg/acquisition/modules/file/config.go
@@ -85,9 +85,23 @@ func (s *Source) Configure(_ context.Context, yamlConfig []byte, logger *log.Ent
 	s.tailMapMutex = &sync.RWMutex{}
 	s.tails = make(map[string]bool)
 
-	s.watcher, err = fsnotify.NewWatcher()
-	if err != nil {
-		return fmt.Errorf("could not create fsnotify watcher: %w", err)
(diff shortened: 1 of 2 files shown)


blotus:
Hello,

Do you have an actual use case for setting `mode: cat` in the acquisition config ?
It's mostly used internally when running in replay mode (with the `--dsn` flag), but `Configure` is not called in this case, `ConfigureByDSN` is called instead and does not set an inotify watch.

PR #4488, opened 2026-05-24
fix(plugin-broker): drain pending alerts and wait for in-flight deliveries on shutdown

andreabravetti:
When cscli notifications test kills the plugin tomb immediately after enqueuing an alert, the broker's select loop can pick the Dying() channel before reading PluginChannel, silently dropping the notification. The inner loop after watcher.Kill() also races Dead() vs PluginEvents, losing alerts queued by addProfileAlert.

Fix with a 3-step shutdown sequence:
1. Drain PluginChannel
2. Flush alertsByPluginName map
3. Wait for in-flight delivery goroutines via sync.WaitGroup before Kill()

Also wrap PluginEvents delivery in a goroutine with WaitGroup tracking.

Fixes issue #4487

### DIFF
--- a/pkg/csplugin/broker.go
+++ b/pkg/csplugin/broker.go
@@ -51,6 +51,7 @@ type PluginBroker struct {
 	pluginProcConfig                *csconfig.PluginCfg
 	pluginsTypesToDispatch          map[string]struct{}
 	newBackoff                      backoffFactory
+	deliveryWg                      sync.WaitGroup
 }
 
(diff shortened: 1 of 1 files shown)


andreabravetti:
I'm having trouble choosing the right area, perhaps a mix of agent and cscli?
A `plugin` or `notification` area might be more suitable.

andreabravetti:
I just fixed this PR, there was a mix of two works in the same branch, sorry about that. Now it should be clean.

PR #4546, opened 2026-07-06
CI hardening

blotus:
- Add `permissions` block
 - Disable creds persistence on checkout
 - Pass manual job parameters through env
 - Check javy/protoc checksums
 - Use `gh cache` instead of a third-party extension
 - Inline the coverage badge update instead of a third-party action
 - Fetch the hub index from a git clone in the bats jobs to avoid cdn-hub rate limiting (403), logging the runner's public IP on failure


sabban (approved):
lgtm

PR #4553, opened 2026-07-08
fix(syslog): don't close the shared acquisition output channel

alxrxs:
## What

The syslog datasource is the only acquisition module that closes the shared pipeline output channel. Removing that `close(out)` fixes a `send on closed channel` panic that crashes the log processor on shutdown/reload when syslog runs alongside another datasource.

## The bug

`StartAcquisition()` passes the **same** `out chan pipeline.Event` to every configured datasource, and the acquisition orchestrator owns it — shutdown is driven by context cancellation (then `drainChan`), and the orchestrator never closes the channel. Every datasource (`file`, `journalctl`, `docker`, …) treats `out` as send-only and leaves it open.

`syslog` is the exception — in `pkg/acquisition/modules/syslog/run.go`:

```go
g.Go(func() error {
    defer close(out)   // <-- closes a channel it does not own
    ...
})
```

When syslog and another datasource are configured together, syslog's `close(out)` on shutdown races with the other datasource still sending on `out`, and Go panics.

This reliably reproduces with the **AppSec** datasource, which sends events on the shared channel from its asynchronous request handling (`appsec_runner.go`, both the in-band and out-of-band paths, e.g. `r.outChan <- evt`). On a `systemctl restart`/reload while a request is still in flight, syslog closes the channel first and AppSec panics.

### Production stack trace (v1.7.8, a `syslog` + `appsec` config)

```
error: send on closed channel
...
panic({0x297c5e0?, 0x2e953c0?})
	runtime/panic.go:860 +0x13a
github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/appsec.(*AppsecRunner).handleOutBandInterrupt(...)
	pkg/acquisition/modules/appsec/appsec_runner.go:426   # v1.7.8 line numbers
github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/appsec.(*AppsecRunner).handleRequest(...)
	pkg/acquisition/modules/appsec/appsec_runner.go:497
github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/appsec.(*AppsecRunner).Run(...)
	pkg/acquisition/modules/appsec/appsec_runner.go:516
github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/appsec.(*Source).StreamingAcquisition.func1.1()
	pkg/acquisition/modules/appsec/run.go:141
```

The process is `Restart=on-failure`'d back up, but the AppSec listener can come back degraded (high CPU, climbing memory, WAF requests timing out) until a subsequent clean restart, so an operator-facing symptom is "AppSec intermittently stops responding after a restart."

## The fix

Remove the `close(out)` from syslog. No consumer relies on it:

- `runParse` reads via `select { case <-ctx.Done(): return; case ev := <-input: }` — it terminates on context cancellation, not on channel close.
- `drainChan` is a non-blocking best-effort drain that handles both open and closed channels. (Its own comment already notes *"We should close the chan on the writer side rather than this"* — there is no single writer to do so, which is exactly why no datasource should.)

This aligns syslog with every other datasource. I also documented the "never close `out`" contract on both the `RestartableStreamer` and `Tailer` interfaces so it doesn't regress.

### Alternatives considered

- **`recover()` in AppSec's send path** — papers over the symptom at one call site while leaving the channel-ownership violation (and the same latent race for any future multi-source combo) in place.
- **`sync.Once` / orchestrator-side close** — the orchestrator has no single writer to coordinate a safe close, and it already shuts the pipeline down cleanly via context cancellation + `drainChan`, so a close is simply unnecessary.

Closing a channel you don't own (with other senders live) is a textbook Go anti-pattern; "don't close what you don't own" is the minimal, root-cause fix.

## Testing

- New regression test `TestStreamDoesNotCloseOutput` (`pkg/acquisition/modules/syslog/syslog_test.go`): starts `Stream()`, cancels the context to tear it down, and asserts the shared output channel is **never** observed closed. It passes on this change and fails on the pre-fix code (I verified both).
- `go build ./...` and `go vet ./pkg/acquisition/...` pass; the new code is `gofmt`-clean.


pkg/acquisition/modules/syslog/run.go
```diff
@@ -43,8 +43,11 @@ func (s *Source) Stream(ctx context.Context, out chan pipeline.Event) error {
 	})
 
 	g.Go(func() error {
-		defer close(out)
-
+		// out is the shared pipeline output channel, fed by every configured
```
blotus: Don't add comments for deleted code, especially for something that is very unlikely to occur again.

pkg/acquisition/modules/syslog/syslog_test.go
```diff
@@ -167,3 +167,45 @@ disable_rfc_parser: true`,
 		})
 	}
 }
+
+// TestStreamDoesNotCloseOutput is a regression test: Stream() must never close
```
blotus: Don't add a test so specific.

If you do want to have a test for this, it should be independent of the datasource and tested for everything datasource automatically.

blotus (changes requested):

alxrxs:
Thanks for the review — both addressed:

- Removed the comment on the deleted line; it's now a plain 2-line removal.
- Replaced the syslog-specific test with `TestDatasourcesDoNotCloseSharedOutputChannel` in `pkg/acquisition/modules/`: it statically parses the AST of every `pkg/acquisition/modules/<name>` directory and fails if any `Stream`/`StreamingAcquisition` implementation calls `close(out)`. No per-datasource config or runtime needed, and it covers every present *and future* module automatically (verified it fails naming the exact file/line when I reintroduce the bug into syslog, and passes clean otherwise).

Force-pushed the updated commit.

alxrxs:
CI failures were a single root cause: `golangci-lint` (staticcheck SA1019) flagged `parser.ParseDir` as deprecated in the new generic test. Replaced it with a manual directory walk + `parser.ParseFile` per file — same static check, no behavior change. Verified locally with the exact CI `golangci-lint` v2.10.1 binary: 0 issues.

blotus (approved):

PR #4555, opened 2026-07-09
fix(docker): avoid panic on colliding nested crowdsec labels

arpitjain099:
`parseKeyToMap` walks into nested label maps with an unchecked type assertion:

```go
for i := 1; i < len(parts)-1; i++ {
    if _, ok := m[parts[i]]; !ok {
        m[parts[i]] = make(map[string]interface{})
    }
    m = m[parts[i]].(map[string]interface{})
}
```

If a container carries both a leaf label and a branch label under the same key, the intermediate key already holds a string, so the assertion panics:

```
crowdsec.enable=true
crowdsec.enable.foo=bar
-> interface conversion: interface {} is string, not map[string]interface {}
```

Container labels are set by whoever launches the container, so on a shared host this can crash the acquisition goroutine (an unrecovered goroutine panic takes down the process). The trigger is also order-dependent because `parseLabels` ranges over a Go map, so today it is a flaky panic vs. a silent clobber depending on iteration order.

Fix: use the comma-ok form and replace a non-map (or absent) value with a fresh nested map, so a deeper key takes precedence instead of panicking.

Added `TestParseLabelsNestedCollisionDoesNotPanic` (asserts no panic and that unrelated labels survive; the exact winner is map-order-dependent, which is pre-existing). Ran `go test ./pkg/acquisition/modules/docker/` locally (passes; the test panics without the fix).


pkg/acquisition/modules/docker/utils.go
```diff
@@ -28,10 +28,15 @@ func parseKeyToMap(m map[string]interface{}, key string, value string) {
...
 
 	for i := 1; i < len(parts)-1; i++ {
-		if _, ok := m[parts[i]]; !ok {
-			m[parts[i]] = make(map[string]interface{})
+		next, ok := m[parts[i]].(map[string]interface{})
+		if !ok {
+			// The key is absent, or a leaf value (e.g. a sibling label like
+			// crowdsec.enable=true) is already stored here. Replace it with a
+			// nested map instead of panicking on the type assertion.
+			next = make(map[string]interface{})
```
blotus, line 50: I think there's a chance of skipping some values depending on the order of the elements in the map (which is random) ?

We probably want iterate over a sorted slice of all labels to avoid potential overwrite.

blotus (changes requested):

pkg/acquisition/modules/docker/utils.go
arpitjain099: Good catch, you're right that the random map order made this non-deterministic. I pushed a change to parseLabels that sorts the keys before building the tree, so the shorter key (crowdsec.enable) is always applied before the more specific one (crowdsec.enable.foo).

With genuinely colliding labels only one can survive, since enable can't be both a string and a nested map at once, but now the branch wins on every run instead of it being a coin flip on iteration order. I strengthened the test to assert the branch wins and to run parseLabels repeatedly so any remaining order dependence would fail it.

blotus (approved):

PR #4564, opened 2026-07-16
revamp the cscli hub list output to make it human friendly

buixor:
image


buixor:
closed in favor of #4567

Issue #2784, opened 2024-01-25
Add the label "machine" to the prometheus metric cs_active_decisions


LuminatiHD:
### What would you like to be added?

I think it would be helpful if the prometheus metric `cs_active_decisions` would show which machine triggered those decisions, similarly as one can with the cscli using  the `-m` flag.  For example:
```
cs_active_decisions{action="ban", environment="prod", instance="lapi.com:6060", job="crowdsec",  machine="machine01.example.com", reason="crowdsecurity/ssh-bf, "}                         10
cs_active_decisions{action="ban", environment="prod", instance="lapi.com:6060", job="crowdsec",  machine="machine02.example.com", reason="crowdsecurity/ssh-bf, "}                         100
```

/kind enhancement

### Why is this needed?

It allows one to know at a glance which machine is the most endangered.
Also, if there is a sudden spike in active decisions, it would be helpful to see which machine triggered that spike / if it happens to be a coordinated attack on multiple machines or one singular machine.

LaurenceJJones:
Hi thank you for opening a enhancement request.

Currently with the release of `v1.6.0` the team has their hands full with other projects.

I have added the "good first issue" tag to indicate pull requests from the everyone are welcome to resolve this.

PR #4568, opened 2026-07-21
Add machine label to cs_active_decisions metric

kushiemoon-dev:
## Summary
- Adds a `machine` label to the `cs_active_decisions` Prometheus GaugeVec, so each time series shows which enrolled machine owns the alert behind those active decisions.
- Resolves the owning machine via `Decision -> Alert -> Machine` using the native `alert_decisions` FK column (no raw SQL join), folding rows that share `(reason, origin, action, machine)`. Falls back to `"N/A"` for decisions whose alert has no owning machine.
- Existing consumers (`cscli metrics`) are unaffected: the new label only splits existing rows, the summed totals per `(reason, origin, action)` stay identical.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/database/...` (existing suite + 4 new regression tests: machine-owned alert, no-owning-machine fallback, folding across alerts owned by the same machine, splitting across different machines)
- [x] `go vet ./pkg/metrics/... ./cmd/crowdsec/...`

Related: #2784

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,4 @@ __pycache__
 # local agent/tooling dirs (not part of the repo)
 .agents/
 .claude/
+.worktrees/
--- a/cmd/crowdsec/metrics.go
+++ b/cmd/crowdsec/metrics.go
@@ -46,7 +46,7 @@ func computeDynamicMetrics(next http.Handler, dbClient *database.Client) http.Ha
 		metrics.GlobalActiveDecisions.Reset()
 
 		for _, d := range decisions {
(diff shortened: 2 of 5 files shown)


blotus:
Hello,

I'm going to close this PR: the implementation won't scale on a lot of installations.

Currently, we fetch one row per (basically) scenario with active decisions: a few dozens for most setups.
With this change, we would fetch one row per decision: this can easily end up returning hundreds of thousands of rows (as most alerts will have only a single decision).
Because this query runs on every prometheus scrape (so likely multiple times per minute), the DB load would be way too high.

You also pass all alert ids at once to the query to fetch the owner, which will blow up if you have too many (32k for sqlite, 65k for PG, ...).

The performance tradeoff is definitely not worth it.

If you find a better way to implement this, feel free to open another PR.

PR #4569, opened 2026-07-21
Add cs_machines_heartbeat_seconds metric

### LINKED ISSUE #2758: Add machines heartbeat as a prometheus metric

/kind feature

### Why is this needed?

it would help monitoring machines

kushiemoon-dev:
## Summary
- Adds a new Prometheus gauge vector `cs_machines_heartbeat_seconds`, labeled `machine`, exposing seconds elapsed since each registered machine's last heartbeat.
- Registered at `MetricsLevelFull` only (same tier as `cs_active_decisions`/`cs_alerts`), recomputed on every `/metrics` scrape via `.Reset()` + repopulate from `ListMachines`, so machines that no longer exist don't leak stale series.
- Uses label `machine` rather than the `instance` example in the issue, since `instance` is a Prometheus-reserved meta-label for the scrape target itself and would collide; `machine` matches the existing `cs_lapi_machine_requests_total` convention.
- Machines that have never heartbeated (`LastHeartbeat == nil`) emit no series.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/metrics/... ./cmd/crowdsec/...` (3 new tests: active machine value/label, nil-heartbeat skip, stale-series eviction across scrapes)
- [x] `gofmt -l` / `go vet ./pkg/metrics/... ./cmd/crowdsec/...`

Closes #2758


blotus:
Hello,

Thanks for the PR, but I think this is the wrong approach.

You compute the duration since the last heartbeat every time a scrape occurs, which is fine when you have a single/low number of machines, but this will take more time (and DB load) for a bigger setup.

I think a better approach would be to expose instead a `cs_machines_last_heartbeat_timestamp` which would contain the timestamp from the last heartbeat: this would be set in the heartbeat handler, which has access to the machine name, so this would turn this into a free operation even for bigger setup.

Exposing the timestamp allows you to recompute the delta directly in prometheus if needed + gives a lot more flexibility in terms of alerting/display.

blotus (approved):
Thanks !

PR #4574, opened 2026-07-24
Populate missing config files when only part of /etc/crowdsec is mounted

saruprim:
## Problem

The container only populates `/etc/crowdsec` from `/staging/etc/crowdsec` when **both** `config.yaml` and `local_api_credentials.yaml` are missing:

```bash
if [ ! -e "/etc/crowdsec/local_api_credentials.yaml" ] && [ ! -e "/etc/crowdsec/config.yaml" ]; then
    rsync -av --ignore-existing /staging/etc/crowdsec/* /etc/crowdsec
fi
```

This means that if a user mounts only one of these files, the staging configuration is never copied and the missing file is not created.

## Example

A valid use case is running an agent-only container with:

* `DISABLE_LOCAL_API=true`
* a bind-mounted `local_api_credentials.yaml` pointing to an external LAPI

For example:

```yaml
url: http://lapi.example.com:8080
login: my-agent
password: secret
```

In this scenario, only `local_api_credentials.yaml` is provided by the user.

Because the file already exists, the condition above evaluates to false, `config.yaml` is never copied from `/staging`, and the container later fails because `/etc/crowdsec/config.yaml` is missing.

## Expected behavior

If either `config.yaml` or `local_api_credentials.yaml` is missing, the container should populate any missing default configuration files from `/staging` without overwriting existing ones.

Since `rsync` is already called with `--ignore-existing`, it is safe to run even when some configuration files are already present.

One possible fix would be to remove the condition entirely and always execute:

```bash
mkdir -p /etc/crowdsec
rsync -av --ignore-existing /staging/etc/crowdsec/ /etc/crowdsec/
```


saruprim:
Just following up on this PR. It has been open for a few weeks without a review.
The fix addresses a real issue with partially mounted `/etc/crowdsec` configuration and should be backwards-compatible thanks to r`sync --ignore-existing`.
Would someone from the team be able to take a look when they have a chance?

PR #4587, opened 2026-07-30
WAF: supress erroneous warnings/errors when loading config / no rule has matched

blotus:
- Update coraza to suppress warnings for unknown keys in collections
- Do not attempt to load modsec datafiles as crowdsec datafiles


sabban (approved):
lgtm

buixor (approved):

PR #4619, opened 2026-08-20
db: add composite index for decision dedup

blotus:
### Indexes changes

**1. `decision_value` → `decision_value_type_scope_until`**

The decision dedup query is a self-join on `(value, type, scope)` and a range on `until`.
Indexing only `value` meant an extra row fetch for every candidate, just to check the other columns.

**2. `decision_value` dropped**

This one needed to be dropped because of the previous one. If present, the DB would tend to use it, negating the improvement of `decision_value_type_scope_until`.

Because we do not configure ent to drop all unknown indexes on startup (as this would also delete indexes manually created by users), a hook was added to only drop this index (using versioned migrations would probably be a cleaner solution, but it's a lot more involved as we don't use them at the moment)

### `decisionDeleteBulkSize` 256 → 2000

The addition of the `decision_value_type_scope_until` index slows down the write on the table (since changing it now means updating 2 indexes).
To compensate for this, increasing the decision deletion bulk size keeps the expiration at around the same level before the new index.

### Impact

On a DB with 1.5M decisions (1.2M active, 300k expired, ~500 expired inside the delta window, 100k values shared across 3 blocklist origins), LAPI response time:

| Query | SQLite | MySQL 8.4 | PG 16 |
|---|---|---|---|
| `new` / startup | 4.03s → **3.56s** | 2.87s → **2.14s** | 1.90s → 1.88s |
| `new` / delta | 103ms → 89ms | 334ms → **1.8ms** | 31ms → **1.8ms** |
| `deleted` / startup | 788ms → **712ms** | 942ms → **762ms** | 486ms → 504ms |
| `deleted` / delta | 2.1ms → 1.8ms | 1.7ms → 1.4ms | 1.3ms → 1.5ms |


pkg/database/decisions.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/crowdsecurity/crowdsec/pkg/database/ent/decision"
 )
 
-const decisionDeleteBulkSize = 256 // scientifically proven to be the best value for bulk delete
+const decisionDeleteBulkSize = 2000 // bulk expire/delete batch; matches csconfig.maxDecisionBulkSize used for inserts
```
sabban, line 19: why not use csconfig.maxDecisionBulkSize then ?

sabban (approved):
minor comment

PR #4620, opened 2026-08-21
lapi: use decision id as cursor for delta pull

### LINKED ISSUE #508: Live LAPI decisions never reach the nftables set, silently — is this expected?
_The cause may well be on the LAPI side rather than in the bouncer — please move this issue if it belongs elsewhere._

## The short version

The LAPI held 2 590 live blocklist decisions. The corresponding nftables set contained **zero**
elements. The bouncer was `(healthy)`, its sets and rules were in place, and its own metrics agreed
with the empty state. Restarting the bouncer put **3 257 IPs** into the firewall that had been in the
LAPI but in no set.

Nothing in the tooling reports this. I'd like to know whether it's expected, and if so how one is
supposed to detect it.

It has since happened a second time, watched live — see the update at the end.

## Evidence

`otx-webscanners` was re-imported on **2026-08-15 at 11:50 UTC**. Blocklist decisions carry a 24 h
duration, so those decisions were live until 08-16 11:50 UTC.

Earlier that day the bouncer had completed a clean full resync —
`2026-08-15T05:49:51Z "64926 decisions added"`, including
`Created set and rule for origin lists-otx-webscanners` on both chains — so the set existed and was
populated at that point.

| nft set | 08-15 22:05 | 08-16 02:39 | after restart, 02:53 | 08-16 03:56 |
|---|---|---|---|---|
| `…-lists-otx-webscanners` | **0** | **0** | **2 454** | 2 454 |
| `…-lists-crowdsec_cve_2025_55182` | 23 631 | 23 200 | 24 002 | 24 002 |
| `…-lists-firehol_cruzit_web_attacks` | 13 157 | 13 157 | 13 157 | 13 157 |
| `…-CAPI` | 25 999 | 26 204 | 26 204 | 26 231 |
| `…-crowdsec` (local decisions) | 26 | 25 | 26 | 25 |
| **total IPv4** | 62 813 | **62 586** | **65 843** | 65 869 |

The resync reconciles exactly with the bouncer's log: **65 843 (v4) + 429 + 72 (v6) = 66 344**, matching
`"66344 decisions added"`. The recovered 3 257 IPs are 2 454 `otx` + 802 `cve` + 1 local decision.

Measurement window: my first reading is at 22:05, so the set is confirmed empty for the 4 h 45 up to
the restart. I have no reading between the 11:50 import and 22:05.

The `cve` re-import had run 63 minutes before the restart and `update_frequency` is 10s, so this is not
a batch still being applied.

## The bouncer reported success throughout

```
2026-08-15T11:50:09Z level=info msg="17458 decisions added"
2026-08-16T01:50:10Z level=info msg="51428 decisions added"
```

No error or warning lines after the successful start at 05:49:51. In every log window I inspected —
including the minutes around each import — `GET /v1/decisions/stream` returns HTTP 200 on its 10 s
cadence. (The engine logs to stdout only, so I could not grep the entire period.)

## What I ruled out

- **Authentication.** A 403 would leave every set empty; four of five were populated, and the stream
  calls return 200.
- **LAPI de-duplication across origins**, i.e. the IPs being served under `CAPI` instead, as described in
  [#506](https://github.com/crowdsecurity/cs-firewall-bouncer/issues/506). That would keep the total
  constant; instead it rose by 3 257 with `CAPI` unchanged.
- **Decision expiry.** The `otx` decisions were live until 08-16 11:50 UTC while the set was empty from
  22:05.
- **Set definition.** All sets are `type ipv4_addr; flags timeout`, no `size` limit, with a `drop` rule
  on both the `input` and `forward` chains.

TTLs read right after the resync, identifying which import each generation came from:
`CAPI` 6 d 21 h · `firehol` 22 h 56 m · `cve` 22 h 56 m (23 200) and 20 h 56 m (802) ·
`otx` **8 h 56 m**, i.e. exactly 24 h from the 11:50 import.

One figure I haven't accounted for: the import created 2 590 decisions (`ban:2585` on the alert) and the
set holds 2 454 after resync. Allowlisting and expiry are the likely explanation; I haven't confirmed it.

## A possible lead on the mechanism

Offered as a hypothesis, not a claim — it isn't needed for the report above, and you'll judge it faster
than I can.

Every blocklist re-import logs as many deletions as additions:

```
lists:otx-webscanners            : added  2590 entries, deleted  2590 entries (alert:104)
lists:firehol_cruzit_web_attacks : added 13231 entries, deleted 13231 entries (alert:134)
lists:crowdsec_cve_2025_55182    : added 24342 entries, deleted 23969 entries (alert:133)
crowdsecurity/community-blocklist: added 15000 entries, deleted 14973 entries (alert:136)
```

If the `new` and `deleted` arrays of one `/v1/decisions/stream` response can carry the same IPs — retired
decision IDs alongside their replacements — then a set keyed by IP would see the deletions cancel the
additions from that same batch.

That would fit the fact that the community blocklist is unaffected: it's a rolling window, so its added
and deleted IPs should be largely different, whereas a stable third-party list replaces a set of IPs with
the same IPs. I have not verified that disjointness, and I can't see the bouncer's internal ordering.

At the 03:50 UTC pull, the community blocklist rotated (`added 15000, deleted 14973`), no third-party
list was re-imported, and no set was damaged — consistent with that reading.

### Update: it happened again, this time watched live

`otx-webscanners` was re-imported on **2026-08-16 at 09:50 UTC** — the engine force-pulled it as its
decisions approached expiry.

| time (UTC) | event |
|---|---|
| 09:50:01 | engine: `lists:otx-webscanners : added 2590 entries, deleted 2590 entries (alert:144)` |
| 09:50:09 | LAPI serves the stream, `HTTP 200` in 371 ms |
| 09:50:10 | bouncer: `17456 decisions added` |
| 14:44 | nft set `…-lists-otx-webscanners` = **0** |

Those 2 590 decisions are live in the LAPI until 2026-08-17 09:50 UTC, so this is a current 2 590-IP
enforcement gap, five hours after an import the bouncer reported as applied. No restart happened in
between — the bouncer has been up since 02:53 UTC.

This also clears the engine side: `ShouldForcePullBlocklist()` worked exactly as designed, forcing the
re-import despite an unchanged `Last-Modified`.

Possibly the same symptom, reported without an answer:
[discourse — "OpenWRT: entries got added and deleted"](https://discourse.crowdsec.net/t/openwrt-entries-got-added-and-deleted/2710)
(2025-10-31). Same log pattern on a different platform. Their pasted set shows
`timeout 23h59m52s expires 10h42m19s` — inserted ~13 h earlier despite a re-import having just run.

Related background: [crowdsec#3373](https://github.com/crowdsecurity/crowdsec/issues/3373) on duplicate
decisions across origins, and [crowdsec#3464](https://github.com/crowdsecurity/crowdsec/issues/3464) on
the global 24 h duration for custom blocklists, which is what makes this surface within a day.

## Environment

- `crowdsec-firewall-bouncer` **v0.0.36** (latest release), Docker, host network + `NET_ADMIN`
- `crowdsec` **v1.7.8** (latest stable), Docker, SQLite backend
- Blocklists: `crowdsec_cve_2025_55182`, `firehol_cruzit_web_attacks`, `otx-webscanners`

```yaml
api_url: http://127.0.0.1:8080/
api_key: <redacted>
update_frequency: 10s
log_media: stdout
log_level: info
mode: nftables
deny_action: DROP
deny_log: true
deny_log_prefix: "crowdsec-drop: "
nftables_hooks:
  - input
  - forward
nftables:
  ipv4:
    enabled: true
    set-only: false
    table: crowdsec
    chain: crowdsec-chain
    priority: -10
  ipv6:
    enabled: true
    set-only: false
    table: crowdsec6
    chain: crowdsec6-chain
    priority: -10
prometheus:
  enabled: false
```

Happy to provide anything else that would help.

### Update 2026-08-21: four days without a restart — 2 % of blocklist decisions actually applied

My previous comment (08-19) reported that the mechanism below did not reproduce, and I withdrew
it. That was **one** non-reproduction: the bug looks **intermittent**, not absent. Since then I
left the bouncer running untouched for 4 days — until now a nightly container recreation had been
silently resyncing it, which masked the steady state.

**Reading at 2026-08-20 22:35 UTC, bouncer up 4 days, no restart in between:**

| origin | live in LAPI | in nftables (IPv4) | missing |
|---|---|---|---|
| `lists:crowdsec_cve_2025_55182` | 21 699 | 736 | 20 963 |
| `lists:firehol_cruzit_web_attacks` | 13 229 | **0** | 13 229 |
| `lists:otx-webscanners` | 2 590 | 7 | 2 583 |
| **all `lists:*`** | **37 518** | **743** | **36 775 (98 %)** |
| `CAPI` (control, same reading) | 23 670 | 22 661 | coherent |

The `CAPI` row is the point: same bouncer, same stream, same instant. The delta stream is
working — only the `lists:*` origins fail to land.

**That day's imports match `survivors ≈ added − deleted`:**

```
01:28:17  lists:crowdsec_cve_2025_55182    : added 21718 entries, deleted 21142 entries   -> delta 576
01:28:20  lists:firehol_cruzit_web_attacks : added 13231 entries, deleted 13231 entries   -> delta 0
03:28:17  lists:otx-webscanners            : added  2590 entries, deleted  2590 entries   -> delta 0
01:28:17  crowdsecurity/community-blocklist: added 15000 entries, deleted 14979 entries   (control)
```

Both lists where `added == deleted` ended at ~0. The only list with a non-zero delta is also the
only one with a non-empty set — 736 against a delta of 576; I cannot account for the remaining
160. The community blocklist, whose added and deleted IPs are disjoint, is unaffected, as before.

**A full resync still applies everything correctly.** Restarting the bouncer logged
`59831 decisions added`, and the sets reconcile to the unit: 59 317 (v4) + 514 (v6) = **59 831**.
`lists:*` went from 743 back to 36 086. So the apply path is sound — it is specifically the
**streaming delta path for `lists:*` origins** that drops them.

**Observability note:** `cscli alerts list --origin lists` renders these same imports as
`update : +21718/-0 IPs` — the deletions are not visible there. I first misread that as "no
deletions happened" and drew the wrong conclusion; only the engine log shows them.

blotus:
We currently use the timestamp of the last pull from a bouncer to compute the delta that needs to be sent on a refresh.

A blocklist import sets `created_at` inside a transaction, so a pull while that transaction is still open uses a pre-commit snapshot, sends nothing, and still advances `last_pull` past rows that have not committed yet, meaning the bouncer will never receive them.

Adds `bouncers.stream_cursor`, the highest decision id visible when the request starts reading, and makes the delta `id > cursor`. Anything not yet visible belongs to an uncommitted transaction and will get a higher id, so the cursor cannot step over it .

The delta and startup queries collapse into one — a startup resync is the same query with a zero cursor. `QueryNewDecisionsSinceWithFilters` and the `created_at` predicate are removed.

`stream_cursor` defaults to 0, so every existing bouncer receives one full set of active decisions on its first pull after upgrade, then runs on the cursor.

We keep the timestamp method for deletion, as there are not order guarantee (insertion order is not always the same as expiration order, and a user can delete any decision at any time).

Fixes crowdsecurity/cs-firewall-bouncer#508.

### DIFF
--- a/pkg/apiserver/controllers/v1/decisions.go
+++ b/pkg/apiserver/controllers/v1/decisions.go
@@ -155,10 +155,13 @@ func (c *Controller) DeleteDecisions(gctx *gin.Context) {
 	gctx.JSON(http.StatusOK, deleteDecisionResp)
 }
 
-func writeStartupDecisions(gctx *gin.Context, now time.Time, filters map[string][]string, dbFunc func(context.Context, time.Time, map[string][]string) ([]*ent.Decision, error)) error {
+// writeDecisions streams the decisions returned by query into the array the caller has already
+// opened, paginating by id from startID. A startup resync passes 0, a stream delta passes the
+// bouncer cursor: both are the same query, they only differ in where they start.
+func writeDecisions(gctx *gin.Context, now time.Time, filters map[string][]string, startID int, query func(context.Context, time.Time, map[string][]string) ([]*ent.Decision, error)) error {
 	limit := 30000 // FIXME : make it configurable
 	needComma := false
-	lastId := 0
+	lastId := startID
 
 	ctx := gctx.Request.Context()
 
@@ -177,7 +180,7 @@ func writeStartupDecisions(gctx *gin.Context, now time.Time, filters map[string]
 			filters["id_gt"] = []string{lastIdStr}
 		}
 
-		data, err := dbFunc(ctx, now, filters)
+		data, err := query(ctx, now, filters)
 		if err != nil {
 			return err
 		}
@@ -211,7 +214,7 @@ func writeStartupDecisions(gctx *gin.Context, now time.Time, filters map[string]
 			lastId = data[len(data)-1].ID
 		}
 
-		log.Debugf("startup: %d decisions returned (limit: %d, lastid: %d)", len(data), limit, lastId)
+		log.Debugf("stream: %d decisions returned (limit: %d, lastid: %d)", len(data), limit, lastId)
 
 		if len(data) < limit {
 			gctx.Writer.Flush()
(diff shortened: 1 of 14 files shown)


sabban (approved):
lgtm

PR #4633, opened 2026-08-26
alerts: create metas only if alert was not discarded

### LINKED ISSUE #4631: Orphan meta rows are never deleted: FlushOrphans skips them and dropped alerts leak their metadata
### Description

`meta` rows are written before the alert that owns them, and when the alert is
then dropped, nothing ever deletes them. `FlushOrphans` reaps orphan `events`
and orphan `decisions` but never orphan `meta`, so the rows accumulate for the
life of the database.

`buildMetaCreates` (pkg/database/alerts.go#L513-L541) inserts every meta row
with `alert_metas = NULL`. The link is made later, when the alert row is built
with `AddMetas(metas...)`.

In `createAlertBatch` the metas are inserted, and then the alert can still be
dropped:

```go
metas, err := buildMetaCreates(ctx, c.Log, txEnt, alertItem)   // rows inserted here
...
// if all decisions were discarded, discard the alert too
if discardCount > 0 && len(decisions) == 0 {
    c.Log.Warningf("dropping alert %s: all decisions invalid", alertItem.UUID)
    continue                                                    // metas stay behind
}
```

The batch transaction added in #4438 does not help here, because `continue`
is not an error path. The loop moves to the next alert, the remaining alerts
succeed, and `tx.Commit()` at L715 commits the orphaned meta rows along with
them.

Nothing reaps them afterwards. `FlushOrphans` handles events and decisions and
stops:

```go
eventsCount, err := c.Ent.Event.Delete().Where(event.Not(event.HasOwner())).Exec(ctx)
...
eventsCount, err = c.Ent.Decision.Delete().Where(
    decision.Not(decision.HasOwner())).Where(decision.UntilLTE(time.Now().UTC())).Exec(ctx)
```

Only `DeleteAlertGraph` / `DeleteAlertGraphBatch` delete metas explicitly, and
the flush scheduler does not call them. The `alerts` -> `meta` FK is
`ON DELETE CASCADE`, so metas that do get linked are cleaned up correctly with
their alert. Only the never-linked ones survive.

### Impact

Measured on a Postgres LAPI handling roughly 290 alerts/sec, 79 days after the
stats were last reset:

| table | total size | live rows | inserted | deleted | % deleted |
| -- | -- | -- | -- | -- | -- |
| meta | 102 GB | 326,426,322 | 12,479,912,757 | 11,956,425,370 | 95.8% |
| events | 827 MB | 34,307 | 3,495,822,517 | 4,370,086,982 | ~100% |
| alerts | 275 MB | 19,316 | 1,966,684,586 | 1,917,117,915 | 97.5% |
| decisions | 39 MB | 4,447 | 52,149,449 | 52,559,869 | ~100% |

Alerts, events and decisions all return to roughly 100% deleted, so the flush
job is working. `meta` stops at 95.8%. That missing 4.2% is about 6.6M rows a
day, or 2.1 GB a day of disk growth that never comes back.

19,316 live alerts carry about 6.3 metas each, so roughly 120k of those 326M
rows are real. The other 99.96% are orphans. Autovacuum was keeping up
throughout (351k dead tuples), so this is live-row accumulation and not bloat.
The primary key index alone had grown to 58 GB.

The 4.2% figure will vary with how often an install hits a drop path. #4506
shows the same two-phase insert failing loudly on MySQL, and every one of those
failed batches leaks its metas too.

### Suggested fix

Reaping them in `FlushOrphans` matches what the function already does for
events, and covers any drop path that gets added later:

```go
metaCount, err := c.Ent.Meta.Delete().Where(meta.Not(meta.HasOwner())).Exec(ctx)
```

That needs an age guard, since a meta row is legitimately unlinked between its
own insert and the insert of its alert.

Building the metas only after the discard check would close the specific path
above, but it leaves the general shape of the problem, where any early return
between the meta insert and the alert insert leaks rows that nothing collects.

### Environment

- Observed on 1.7.7, Postgres 15.
- Code paths above are from master at d66b45e0, so 1.7.8 and 1.8.0-rc2 are
  affected as well.

blotus:
Fixes #4631 

The fix is two-fold:
 - Fix the root cause by only creating metas if we keep the alert
 - Add a batched flush of orphan metas in FlushOrphans to fix existing database. The flush is batched because a long running database might have accumulated a tons of orphan metas (linked issue shows 300M rows), and it could take a very long time to run on the 1st flush.

### DIFF
--- a/pkg/database/alerts.go
+++ b/pkg/database/alerts.go
@@ -653,11 +653,6 @@ func (c *Client) createAlertBatch(ctx context.Context, machineID string, owner *
 			return nil, rollbackOnError(tx, err, fmt.Sprintf("building events for alert %s", alertItem.UUID))
 		}
 
-		metas, err := buildMetaCreates(ctx, c.Log, txEnt, alertItem)
-		if err != nil {
-			c.Log.Warningf("error creating alert meta: %s", err)
-		}
-
 		decisions, discardCount, err := c.buildDecisions(ctx, c.Log, txEnt, alertItem, stopAtTime)
 		if err != nil {
 			return nil, rollbackOnError(tx, err, fmt.Sprintf("building decisions for alert %s", alertItem.UUID))
@@ -667,6 +662,13 @@ func (c *Client) createAlertBatch(ctx context.Context, machineID string, owner *
 		if discardCount > 0 && len(decisions) == 0 {
 			c.Log.Warningf("dropping alert %s: all decisions invalid", alertItem.UUID)
 			continue
+		}
+
+		// after the discard check: metas are linked to the alert only when it is built,
+		// so inserting them earlier leaves orphan rows behind when the alert is dropped
+		metas, err := buildMetaCreates(ctx, c.Log, txEnt, alertItem)
+		if err != nil {
+			c.Log.Warningf("error creating alert meta: %s", err)
 		}
 
 		builder := txEnt.Alert.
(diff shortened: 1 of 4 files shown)


sabban (approved):
lgtm

PR #4634, opened 2026-08-26
CI: fix CAPI pull test on windows

blotus:
TestAPICPull/test_pull was flaky on Windows runners: the test started api.Pull() in a
goroutine, slept 500ms, then asserted the decision was in the DB, meaning that on a slow runner (windows) the write could happen after the assertion.

Test now waits for the condition (require.EventuallyWithT) instead of a fixed sleep, and
shuts down + joins the pull routine in a cleanup, so it no longer leaks a goroutine that
kept pulling against a closed DB for the rest of the package run.

apic.go: Pull()'s "waiting for scenarios" loop used a bare time.Sleep and ignored the
tomb, so the routine could not be stopped until a machine registered a scenario. It now
selects on pullTomb.Dying(). This is what made the leaked goroutine unstoppable in the
test, and it also blocks shutdown in production.

### DIFF
--- a/pkg/apiserver/apic.go
+++ b/pkg/apiserver/apic.go
@@ -1061,7 +1061,14 @@ func (a *apic) Pull(ctx context.Context) error {
 			toldOnce = true
 		}
 
-		time.Sleep(1 * time.Second)
+		select {
+		case <-time.After(1 * time.Second):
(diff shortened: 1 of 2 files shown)


sabban:
lgtm

sabban (approved):
lgtm

PR #4656, opened 2026-09-03
[bot-detection] drop the URI from the cookie

buixor:
drop uri from the cookie : it can be big, user-controlled and waste cookie space. Keep it available in on_submit tho


buixor:
need to test how it behaves during upgrade : what happens to existing cookies that do have the URL inside them.

blotus (approved):

PR #4666, opened 2026-09-08
add cscli bot subcommand

blotus:
Add a new cscli `bot` command that contains:
 - `match`: takes an IP/UA/path as a parameter to test whether it matches any bot data files with `MatchKnownBot`
 - `fsid`: takes an FSID as a parameter and parses it to reconstruct part of the original fingerprint for easier understanding


pkg/appsec/challenge/fsid/fsid.go
```diff
@@ -0,0 +1,461 @@
+// Package fsid decodes an FSID (the JA4-inspired fingerprint id produced by
```
buixor, line 1: I'm afraid this will become out of sync with fpscanner on upcoming release. Can we have a way to make it more tightly attached to fpscanner version or so ?

PR #4667, opened 2026-09-08
waf: add SkipProcessing() helper

blotus:
Add a new `SkipProcessing(reason)` helper to return early before evaluating any rules, without dropping the request (counterpart of `DropRequest(reason)`).

If `SkipProcessing()` and `DropRequest()` are called in the same reques, the 1st one will win.


pkg/appsec/appsec.go
```diff
@@ -603,46 +632,118 @@ func (w *AppsecRuntimeConfig) ClearResponse(state *AppsecRequestState) {
...
+	}
+}
+
+// setOutcome records a terminal hook decision for the request's band and
+// reports whether it was kept. The first decision wins: a hook that drops and
+// then skips (or the reverse) gets the one it took first, rather than one
+// picked by the order of the runner's checks.
+func (w *AppsecRuntimeConfig) setOutcome(state *AppsecRequestState, request *ParsedRequest, outcome *HookOutcome) (bool, error) {
+	if existing := state.Outcome(request); existing != nil {
+		w.Logger.Debugf("ignoring %s outcome (%s): request already had a %s outcome (%s)",
```
buixor: is worth more than debug ? not really supposed to happen (but can be pretty verbose too ...)

pkg/appsec/appsec.go
```diff
@@ -1634,6 +1735,14 @@ func (w *AppsecRuntimeConfig) SendChallenge(ctx context.Context, state *AppsecRe
 		return errors.New("SendChallenge can only be called from an in-band hook (on_challenge or post_eval)")
 	}
 
+	// A hook already took a terminal decision for this band (DropRequest,
+	// SkipProcessing). Serving a challenge on top of it produced an
+	// incoherent response: a "ban" remediation carrying a challenge page.
+	if outcome := state.Outcome(request); outcome != nil {
+		w.Logger.Debugf("SendChallenge no-op: request already %s (%s)", outcome.Action, outcome.Reason)
```
buixor: same as above

buixor (approved):

PR #4672, opened 2026-09-09
postoverflow: run them in a goroutine to not block writers, enforce max in-flight alerts, and warn if queue is becoming too full

### LINKED ISSUE #4600: AppSec: synchronous reverse_dns postoverflow enrichment stalls in-band responses via unbuffered pipeline backpressure (bouncer fail-open bursts)
**Summary**

With the `crowdsecurity/rdns` postoverflow installed, a slow PTR lookup for an alert's source IP freezes AppSec in-band inspection for the whole engine. Remediation-component requests then exceed `APPSEC_PROCESS_TIMEOUT` and **fail open** (`APPSEC_FAILURE_ACTION=passthrough`), i.e. requests are served uninspected  - on our busiest host 76-1,413 timeouts/hour (~0.5% of inspections, bursts to 480/min).

**Environment:** crowdsec v1.7.8 (rpm), AlmaLinux 8.10, central LAPI, appsec acquisition on `127.0.0.1:7422` with `crowdsec-openresty-bouncer` v1.2.1 in-band, default `routines: 1`, ~600 alerts/h (WAF + log scenarios), OS resolver = ISP shared resolver, no local cache, default resolv.conf timeouts (5sx2).

**Mechanism** (all refs v1.7.8):

1. `crowdsecurity/rdns` runs in the postoverflow parse inside `runOutput` -> `parser.reverse_dns` (`pkg/parser/enrich_dns.go:19`) -> `net.LookupAddr`  - synchronous, unbounded by any context/timeout of its own. Hostile alert IPs frequently sit in lame/slow PTR zones -> 0.2-10 s lookups.
2. While it blocks, the event train jams: overflow chan <- buckets <- parsers <- acquisition output (all unbuffered).
3. The AppSec runner shares that output channel (`runner.outChan = out`, `pkg/acquisition/modules/appsec/run.go:137`) and blocks in `handleRequest` on `r.outChan <- evt` (`appsec_runner.go:336/341/420/426`). For in-band matches this send happens **before** the response is written to the remediation component (`appsec_runner.go:472`).
4. `InChan` is also unbuffered (`appsec/config.go:169`), so every handler blocks; all in-flight inspections exceed the client's read budget -> fail-open burst. Burst size = request rate x PTR stall length.

**Evidence**

- Goroutine dump captured *while* a probe to `:7422` was hanging (437 ms)  - `runOutput` blocked in the lookup:

```
goroutine 157 [IO wait]:
internal/poll.runtime_pollWait(...)
...
net.dnsPacketRoundTrip(...)
net.(*Resolver).exchange(...)
net.(*Resolver).tryOneName(...)
net.(*Resolver).goLookupPTR(...)
net.LookupAddr(...)
github.com/crowdsecurity/crowdsec/pkg/parser.reverse_dns(...)
	github.com/crowdsecurity/crowdsec/pkg/parser/enrich_dns.go:19
github.com/crowdsecurity/crowdsec/pkg/parser.(*RuntimeStatic).Apply(...)
	github.com/crowdsecurity/crowdsec/pkg/parser/runtime.go:167
[... postoverflow parse inside main.runOutput ...]
```

- The inband parsing histogram is blind to it: 1.04 M inspections, mean 0.47 ms, only 9 ever >250 ms  - while the same period logged tens of thousands of >1000 ms client read-timeouts. The stall sits between accept and the timed span.
- Every clustered burst-second across a 90-min window (22/22) had an alert landing at LAPI +2...+12 s later  - the stalled-lookup-then-flush signature.
- Burst-adjacent alert source IPs re-resolved hours later (warm cache = lower bound): 10/26 still >200 ms, incl. 1,056 ms and a SERVFAIL at 487 ms.
- **Removal experiment:** `cscli postoverflows remove crowdsecurity/rdns --force` + restart dropped the host from 76-1,413/h to **57 in the first full hour** (lowest on record) and 15-74/h over the next nine hours, while an untouched sibling host (same stack, same LAPI) ran high (127 that hour). Probe watcher: 0 stalls >=350 ms in 6,038 probes post-change.
- `routines: 4` does **not** mitigate (measured null): all runners share the same jammed output channel.

**Related:** `cmd/crowdsec/output.go` already acknowledges this failure class  - *"Under high load, LAPI may take between 1 and 2 seconds to process ~100 alerts, which slows down everything including the WAF"*  - and made `PushAlerts` async. Parser-stage enrichment has the same coupling but no such fix. Symptom also matches crowdsecurity/cs-openresty-bouncer#75 (closed unresolved: high-traffic timeouts with 0.2 ms engine latency).

**Suggested fixes** (any one breaks the coupling): bound the enrichment (`LookupAddr` with a short-timeout context), make enrichment async/cached, buffer the overflow/acquisition channels, or give the AppSec runner a dedicated buffered emission queue so in-band responses never share fate with the alert pipeline.

**Side note:** in this deployment the `reverse_dns` value did not appear in any of 1,834 alert payloads at LAPI  - if that's expected for appsec/scenario alerts, the enrichment cost bought nothing here.

**Workaround for affected users:** remove `crowdsecurity/rdns` (requires `--force`; owned by `crowdsecurity/linux` / `whitelist-good-actors`). `seo-bots-whitelist` degrades gracefully to its IP-list expression.

blotus:
Postoverflow parsing ran inline in the `runOutput` select loop. Postoverflow nodes do
blocking I/O (such as `crowdsecurity/rdns` that does a PTR lookup) and the
pipeline channels are unbuffered, so one slow lookup backpressured everything upstream:
buckets, parsers, acquisition, and the appsec runner, which emits into that same chain
*before* writing the in-band response. WAF requests then exceeded the remediation
component's `APPSEC_PROCESS_TIMEOUT` and failed open.

Postoverflow now runs in a worker behind a bounded queue, so the pipeline never waits on
it.
On a full queue the overflow is dropped and counted, rather than blocking: postoverflow
is where the whitelists run, so pushing an alert unparsed would turn a stall into false
positive bans. Dump mode (`cscli explain`, `-dsn`) still parses inline.

New config, with the flat `*_routines` keys kept as the fallback:

```yaml
crowdsec_service:
  pipeline:
    parser:  {routines: 4}
    buckets: {routines: 2}
    output:  {routines: 1, queue_size: 256}
```

New metrics cs_postoverflow_queue_depth and cs_postoverflow_dropped_total, plus a
warning at 75% full so the queue is visible before anything is lost.

Fixes #4600


cmd/crowdsec/output.go
```diff
@@ -80,24 +90,159 @@ func PushAlerts(ctx context.Context, alerts []pipeline.RuntimeAlert, client *api
...
+
+		if err := process(ctx, event); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Warns on crossing 75% full, then on falling back under 25%: once we drop, it's
```
buixor, line 174: + different warning at 100% with an explicit message about overflows being dropped

cmd/crowdsec/output.go
```diff
@@ -80,24 +90,159 @@ func PushAlerts(ctx context.Context, alerts []pipeline.RuntimeAlert, client *api
...
+// too late to react. The gap between the two keeps a queue hovering at one
+// threshold from logging on every tick.
+func warnQueuePressure(depth int, size int, warned bool) bool {
+	switch {
+	case depth*4 >= size*3 && !warned:
+		log.Warnf("postoverflow queue is %d/%d full, a postoverflow parser is slow (dns?): overflows will be dropped if it fills up", depth, size)
+		return true
+	case depth*4 < size && warned:
+		log.Info("postoverflow queue is draining")
```
buixor: specify below 25%

cmd/crowdsec/output.go
```diff
@@ -80,24 +90,159 @@ func PushAlerts(ctx context.Context, alerts []pipeline.RuntimeAlert, client *api
...
+
 	for {
 		select {
 		case <-ticker.C:
+			depth := len(povfw)
+			queueDepth.Set(float64(depth))
+			warnedPressure = warnQueuePressure(depth, queueSize, warnedPressure)
+
+			if droppedSinceTick > 0 {
+				log.Warnf("postoverflow queue full, dropped %d overflow(s)", droppedSinceTick)
```
buixor: maybe we should throttle this message : every 10s is a tad aggressive. Once per minute top ?

if we do this correctly, we don't really need the 100% warning separately.

cmd/crowdsec/output.go
```diff
@@ -115,13 +260,34 @@ func runOutput(
 				}
 				return nil
 			})
-		case <-outputsTomb.Dying():
+		case err := <-workerErr:
+			// a postoverflow failure killed the output routine before, keep that
```
buixor: was it the case already ? looks very aggressive

pkg/csconfig/crowdsec_service.go
```diff
@@ -31,6 +32,59 @@ type CrowdsecServiceCfg struct {
 
 	SimulationFilePath string              `yaml:"-"`
 	ContextToSend      map[string][]string `yaml:"-"`
+
+	PostOverflowQueueSize int `yaml:"-"` // resolved from Pipeline
+}
+
+type PipelineStageCfg struct {
```
buixor, line 39: let's be crazy and try to be consistent on 's' or no 's' in routine parsers/buckets/etc.

pkg/metrics/global.go
```diff
@@ -114,3 +114,23 @@ var GlobalPourHistogram = prometheus.NewHistogramVec(
 	},
 	[]string{"type", "source"},
 )
+
+const GlobalPostOverflowQueueDepthMetricName = "cs_postoverflow_queue_depth"
```
buixor, line 118: let's try to ship those to capi metrics

PR #4676, opened 2026-09-10
notification-http: document and validate mTLS configuration

### LINKED ISSUE #4337: Support mTLS in http plugin

Hi,

Would it be possible to add Mutual TLS (mTLS) authentication to the http plugin?
Similar to `headers`, it would allow to authenticate to endpoints who require mTLS, by loading a key/certificate in the notifications/http.yaml configuration.

/kind enhancement

### Why is this needed?

We would like to forward alerts to our on-premise Opensearch, which is configured to only allow mTLS

wangyusheng1985:
Support and document mTLS authentication in the HTTP notification plugin.

Added documentation in `http.yaml` for `cert_path`, `key_path`, and `ca_cert_path`, added validation in `cmd/notification-http/main.go` to require both `cert_path` and `key_path` when either is specified, and added test coverage for mTLS in `cmd/notification-http/main_test.go`.

Tested: `go test ./cmd/notification-http/...` and `go vet ./cmd/notification-http/...`.

Validation observed for this change:
- `go test ./cmd/notification-http/...`
- `go vet ./cmd/notification-http/...`
- `go test -run TestConfigureAndNotifyMTLS -v ./cmd/notification-http`

Fixes #4337


blotus:
I'd rather avoid doing comment-only changes in shipped configuration file, as this may warn users on upgrade about a mostly meaningless diff in a config file.

It would be better to open a PR in the documentation to cover the missing config options.

PR #4678, opened 2026-09-11
acquisition/file: don't send half-written lines, or skip the lines after them

n1koo:
## What and why

The file datasource silently drops lines when a log is written in bursts. If the tailer hits EOF mid-line, nxadm/tail sends the partial line, then seeks to EOF. That skips everything written in between ([tail.go#L336-L342](https://github.com/nxadm/tail/blob/v1.4.11/tail.go#L336-L342)). On a production Traefik log (~65k lines/day, v1.8.1), this showed up as 37 `UnmarshalJSON : unexpected end of JSON input` errors in 7 days. The skipped lines log nothing.

This sets `CompleteLines: true`, nxadm/tail's own option for this. It is off by default because the nxadm fork merged hpcloud/tail#126, and was restored as opt-in in nxadm/tail#26. Trade-offs:
- a line is buffered until its newline arrives, so a final line without one is not sent;
- a partial line pending at rotation or truncation is dropped instead of sent broken;
- nxadm then reports read errors as EOF ([tail.go#L254-L267](https://github.com/nxadm/tail/blob/v1.4.11/tail.go#L254-L267)), so a tail that hits one waits silently instead of dying with `file reader died`.

Not covered: at startup the tail seeks to EOF, which can still land mid-line.

Human review: I read the diff line by line and verified the testing below.

## How it was tested

```
$ go test -race -count=10 -run TestLiveAcquisitionPartialLine ./pkg/acquisition/modules/file/
ok  	github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/file	7.314s
$ go test -race ./pkg/acquisition/modules/file/
ok  	github.com/crowdsecurity/crowdsec/pkg/acquisition/modules/file	15.978s
```

On master the new test fails: the partial line arrives as its own line (`["{\"a\":", "1}", "third"]`). It is file-only, because this is the only datasource using nxadm/tail.

I replayed 40,000 production lines in request bursts, one `write()` per line. v1.8.1 parsed 39,961 of them in one run and 39,998 in another; with this patch it parsed all 40,000 both times.

golangci-lint v2.10: 0 issues on linux, windows and freebsd. Tests were run on Linux only.

## Checklist

- [x] One concern only.

AI assistance used: mostly


k4d1sm0:
Another production data point supporting this fix, on v1.7.8.

Setup: a Traefik v3 JSON access log written to a host file (Docker bind mount) and tailed by the CrowdSec agent through a `hostPath` mount in Kubernetes. Default file datasource settings. Average ~1.4 lines/s with bursts, mostly HLS segment/playlist polling.

Measured over ~21h, by comparing the lines written to the file in the agent's uptime window with the `reads` counter in `cscli metrics show acquisition`:

- 111,373 lines written, 110,984 read, so **389 lines never read (0.35%)**
- 168 `UnmarshalJSON : unexpected end of JSON input` errors, all on fragments of lines that are valid JSON in the file
- the remaining ~220 lost lines produced no log at all, which matches the seek-to-EOF behavior described above

I checked `v1.8.1` and `master`: neither sets `CompleteLines`. Detection impact is small for repetitive traffic like scans, but the silent part is what makes this hard to notice.

blotus:
Hello,

It seems the tests are flaky on windows (https://github.com/crowdsecurity/crowdsec/actions/runs/34595197049/job/104708935711?pr=4678#step:6:14353), which is often the case when relying on small sleep for synchronization (the windows runners are extremely slow)

n1koo:
You're right, thanks. The test relied on a 100ms sleep for the tailer to open the file and seek to its end, which isn't enough on the windows runners: the lines were written before the tail started, so it timed out.

It now waits on the tailer instead of sleeping: it writes a marker line until one comes back, then uses a whole line as the signal that the tailer has read the write ending in a partial line. I also kill the tomb from `t.Cleanup`, so a failing subtest no longer leaves a tailer running against a deleted temp dir.

Locally: 10 runs with `-race`, and 10 more with the container capped at 0.25 CPU as a slow-runner stand-in, all pass. On master it still fails immediately, now with `["{\"a\":", "1}", "third"]` instead of a 10s timeout.

PR #4679, opened 2026-09-11
fix(database): keep simulated decisions from suppressing real bans

4erdenko:
The regression coverage and index migration belong to the same fix: separating them would leave either the simulation behavior or the stream query regression untested.

## What and why

A longer simulated decision can hide a real ban from the decision stream, including its deletion when it expires. The deduplication query now excludes simulated competitors by default. With `simulated=true`, both kinds participate and the longest decision wins.

Uses `NOT EXISTS` with `(value, type, scope, until, simulated)` to avoid the MySQL slowdown of the corrected left join. Startup replaces the old index through the existing `dropLegacyIndex` hook, preserving decision rows and operator-created indexes. The simulation filter is retained when the stream reuses its filter map.

Overlapping real bans still prevent premature deletion. No decision data or payload format changes. The separate origin/scenario filtering issue in [#3373](https://github.com/crowdsecurity/crowdsec/issues/3373#issuecomment-2554198963) is outside this fix.

## How it was tested

- `TEST_LOCAL_ONLY=1 GOMAXPROCS=4 GOFLAGS='-p=4 -count=1' make test BUILD_VERSION=v1.8.1`: 2034 tests, 24 skipped.
- `make lint BUILD_VERSION=v1.8.1`, golangci-lint 2.10.0: Linux, Windows and FreeBSD each report zero issues.
- Linux builds, including static builds, and native Windows binaries and notification plugins.
- 20 API cases, 30 database cases and two index-upgrade cases. Database cases also pass on MySQL, PostgreSQL and MariaDB.
- Selected decision/stream/simulation/SQL BATS tests on all four backends: 95 passed, 49 existing skips. MariaDB needed a local GTID-reset adaptation in the test harness.
- 144 result comparisons against an independent reference calculation and eight upgrades of 100,000-row databases; rows and an operator-created index are preserved, including on repeat startup.

Tested with Go 1.26.8 on Linux and Go 1.26.4 on Windows. Also applied to `master` at `bc7d9cc7` and tested the database/API packages with Go 1.27.1. Performance measurements and their limitations are in the follow-up comment.

## Checklist

- [x] One concern only.


pkg/apiserver/decisions_simulation_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/crowdsecurity/crowdsec/pkg/database/ent"
+)
+
+func TestStreamDecisionSimulationOverlap(t *testing.T) {
+	for _, mode := range []string{"startup", "delta_delete", "delta_new"} {
```
blotus: The 3 nested for are not ideal for a test. You can replace them a stable of tests, like we do in most other test files: longer, but much easier to reason about when reading a test

pkg/database/decisions.go
```diff
@@ -171,6 +171,10 @@ func longestDecisionForScopeTypeValue(s *sql.Selector) {
 			t.C(decision.FieldScope),
 			s.C(decision.FieldScope),
 		),
+		sql.ColumnsEQ(
```
blotus: Have you tested the performance impact of the change ? I'd expect it to be minimal, but we had (bad) surprises in the past when making changes to the stream query

pkg/database/decisions_simulation_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/crowdsecurity/crowdsec/pkg/database/ent"
+)
+
+func TestDecisionSimulationOverlap(t *testing.T) {
+	for _, scope := range []string{"Ip", "Range"} {
```
blotus: Should be part of the table below.

blotus (commented):
Thanks !

The change seems ok in principle, but I'd like to have a different behavior when `simulated=true` is passed to the query:  with your change, if you have 2 decisions (one simulated, one not), LAPI will return both of them.
If simulated is passed, I'd like for the simulated decisions to be part of the dedup (ie, return whatever is the longest decision, regardless of the flag)

A few things on top the comments I made:
 - Have you tested the change with sqlite and mysql ? They tend to have vastly different behavior from PG performance-wise (especially mysql)
 - You did not check the 2 boxes that you reviewed + tested the code yourself

4erdenko:
Updated the tests to use flat tables. With `simulated=true`, real and simulated decisions now participate in the same deduplication. The filter is also preserved across stream queries and pages.

The corrected left join was slower on MySQL. Switched to `NOT EXISTS` and appended `simulated` to the composite index, replacing the old index through the existing migration hook.

Tested SQLite 3.50.4, MySQL 8.4.11, PostgreSQL 17.10 and MariaDB 12.3.3. These measurements use isolated Linux containers on one Ryzen 7 5700X workstation, not separate physical servers. Timings cover the database method, not HTTP. Each variant has its own synthetic 100,000-row database and one connection. "Original" is the query before this PR; "Corrected LEFT JOIN" has the requested simulation semantics; "Final" is the updated query and index.

Median full active-decision query with ten overlapping real decisions per key, in milliseconds. Ten samples per variant, alternating A/B/C and C/B/A batches after warm-up:

| Database | Original | Corrected LEFT JOIN | Final |
|---|---:|---:|---:|
| SQLite | 107.53 | 137.49 | 85.65 |
| MySQL | 395.90 | 876.79 | 327.92 |
| PostgreSQL* | 57.09 | 60.10 | 57.94 |
| MariaDB | 282.10 | 696.12 | 321.57 |

*PostgreSQL uses a follow-up with persistent clients, five warm-ups and 30 alternating calls per variant because the first run varied widely (98.92 / 95.67 / 111.69 ms).

Also measured the stream's 30,000-row page limit on unique rows, with persistent clients, five warm-ups and 30 calls per variant in balanced order:

| Database | Original | Corrected LEFT JOIN | Final |
|---|---:|---:|---:|
| SQLite | 116.96 | 117.52 | 123.46 |
| MySQL | 101.53 | 122.58 | 109.76 |
| PostgreSQL | 194.78 | 273.12 | 194.81 |
| MariaDB | 60.93 | 77.07 | 77.70 |

MariaDB unique pages remain about 17 ms slower than the original (+27.5%), and close to the corrected left join (+0.8%). This does not improve every workload. Startup/delta additions and deletions, mixed simulation data, pagination and `simulated=true` were also checked; corrected and final results have identical ordered-ID hashes. The original returns too few real decisions in the mixed workload, so its timing there is not an equal-result comparison.

Local validation: `make test` with `TEST_LOCAL_ONLY=1` completed with 2034 tests and 24 skips; `make lint` passed for Linux, Windows and FreeBSD; Linux and native Windows builds passed. Selected BATS tests passed on all four backends (95 passes, 49 existing skips; MariaDB required a local GTID-reset harness adaptation). An earlier full run hung in the unchanged `TestPluginWatcherInterval`; the subsequent full run completed.

Eight upgrade checks, covering both old indexes on all four backends, preserved all 100,000 rows and an operator-created index and passed a second startup. Index creation adds one-time startup work. The regression tests and 144 additional reference comparisons cover expiry boundaries, equal/NULL expiries, both scopes, decision types, filtering and pagination.

PR #4688, opened 2026-09-15
Appsec mTLS Auth-v2

### LINKED ISSUE #3172: [APPSEC] Allow MTLS auth

zs-gabor:
## 📝 Description
This is a complete rewrite and the official successor to the closed **#4571** PR, implementing mutual TLS (mTLS) support for the AppSec module (#3172). 

After reviewing the architectural feedback regarding LAPI limitations, tokens, and socket constraints, this implementation introduces strict, localized security controls mirroring CrowdSec's standard TLS middleware layer.

## ✨ Key Improvements over #4571:

1. **No More Blanket Auth Bypass (Fixed):** Instead of skipping authentication entirely via a loose `if-else` block, the `w.tlsAuth.ValidateCert` function now acts as a strict security gate. It enforces **Allowed OUs validation**, **Certificate Expiration checks**, and full **CRL/OCSP revocation path validation** (with caching) on every incoming mTLS request.
2. **Unix Socket Isolation (Fixed):** Restored the explicit `canTLS` flag state inside `listenAndServe`. The Unix socket listener strictly passes `false` (`startServer(listener, false)`), ensuring local bouncers communicating over socket paths remain plaintext and do not suffer protocol or connection upgrade breakages.
3. **Cleaned Validations:** Removed duplicate validation logic between configuration unmarshaling and server initialization as requested.

## ⚠️ Architectural Note on LAPI mTLS Constraints:
Please note that in CrowdSec's current architecture, LAPI does not support mixed authentication: it is either mTLS or JWT, but never both simultaneously. 

If an operator enables mTLS on LAPI, no JWT tokens are generated, meaning local bouncers (like the NGINX bouncer) must rely on mTLS. However, running mTLS over a Unix socket is highly non-standard and blocks the bouncer side upstream configuration. Therefore, this PR provides the clean separation via the `canTLS` flag to ensure that while TCP handles mTLS safely via local PKI validation (OU/CRL checks), the Unix socket path isn't forcefully broken by TLS handlers.

Closes #3172


blotus:
Hello,

I'm going to close this PR: changes impacting the way bouncers can authenticate must be discussed beforehand.

In this case, supporting mTLS in the appsec is not just about adding support for mTLS in the HTTP server itself.

Because the appsec does not have access to the LAPI database, the current authentication works by forwarding the API key to LAPI, which can validate it for the appsec (which is not possible when using mTLS).
With your change, the appsec would become responsible for authenticating the client by itself.

While this seems to be the only practical solution, it's something that needs to be talked about, and not just blindly opening a PR.

I'm also a bit confused by this statement in the description:
```
Please note that in CrowdSec's current architecture, LAPI does not support mixed authentication: it is either mTLS or JWT, but never both simultaneously.
```

While a single bouncer can only authenticate with an API key or a cert, you can mix multiple type of authentication between multiple bouncers (which does not matter here at a first glance, because LAPI is not involved in the certificate validation ?).

